PR #26, opened 2026-07-02
ci: version gate for vendored tarballs + twin image publish gates (FDRS-593, FDRS-586)

GaganSD:
Two CI-hardening changes for launch readiness. No application code touched.

## 1. Close the CLI version-bump vendor hole (FDRS-593)
The version-bump gate fired only on `cli/src/**` and ignored `cli/vendor/**`. The CLI bundles vendored tarballs (`@pome-sh/correlator`, `@pome-sh/shared-types`, and — after the twin consolidation — the twin packages) via `bundleDependencies`; swapping one ships a behavior change to users while touching no `cli/src/**` file, so it bypassed the gate.

- Add `cli/vendor/**` to the workflow trigger paths and to the touched-file check.
- Include deletions (`--diff-filter` gains `D`): dropping a vendored `bundleDependencies` entry is a shipping change too.
- Capture the diff to a variable instead of piping into `grep -q`, so a `pipefail` SIGPIPE can't be misread as "no changes."
- Job `name:` kept stable (GitHub matches required checks by name — renaming would silently unbind branch protection).

Verified locally: no-cli-change → skip(0); vendor modify/delete without bump → fail(1); with bump → pass(0).

## 2. Gate twin image publish behind tests + Trivy (FDRS-586)
`twin-{github,slack,stripe}-image.yml` built and pushed rolling GHCR tags on every main push with no test gate and no vulnerability scan.

- Add a `test` job (typecheck + test for the twin package); `publish` now `needs:` it.
- Build+load → **Trivy scan (HIGH,CRITICAL; ignore-unfixed)** before push; push only on non-PR events (fork-safety unchanged). PRs now also run the scan.
- Add a documented `.trivyignore` exception policy.

## Tickets
FDRS-593, FDRS-586. Part of the pome-twins launch-readiness wave-2 execution (PR 1 of 5).

### DIFF
--- a/.github/workflows/cli-version-gate.yml
+++ b/.github/workflows/cli-version-gate.yml
@@ -3,6 +3,7 @@ on:
   pull_request:
     paths:
       - 'cli/src/**'
+      - 'cli/vendor/**'
       - 'cli/package.json'
       - 'cli/.changeset/**'
       - 'scripts/check-cli-version-bump.sh'
@@ -14,7 +15,7 @@ concurrency:
 
 jobs:
   check:
-    name: cli/src/** must bump version or add a changeset
+    name: cli/src/** or cli/vendor/** must bump version or add a changeset
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.github/workflows/cli-version-gate.yml
+++ b/.github/workflows/cli-version-gate.yml
@@ -15,7 +15,11 @@ concurrency:
 
 jobs:
   check:
-    name: cli/src/** or cli/vendor/** must bump version or add a changeset
+    # NB: keep this job name stable — GitHub matches required status checks by
+    # check-run name, so renaming it silently unbinds the gate from any branch
+    # protection that lists the old name. The vendor coverage lives in the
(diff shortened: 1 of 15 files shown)


GaganSD:
## Code review (automated paranoid pass) — all findings addressed in-branch

Two review passes run (version-gate + image-gate). Findings, all fixed:
- **High** — renaming the job `name:` would silently unbind a required status check (GitHub matches by check-run name). Reverted to the stable name. (`main` is currently unprotected, so advisory today.)
- **Medium** — `--diff-filter` excluded deletions, so *removing* a vendored tarball bypassed the gate. Added `D`.
- **Low** — `grep -q` SIGPIPE under `pipefail` could be misread as "no changes"; capture to a variable first.
- **Medium (image gate)** — Trivy scanned build #1 but a second build was pushed (scan/rebuild TOCTOU). Now builds+loads once, scans, then `docker push`es that exact image.
- Verified sound: regex `^cli/(src|vendor)/` doesn't overmatch; `cli/vendor/**` glob matches top-level tarballs; PR push still gated to non-PR; Trivy failure blocks the push.

PR #27, opened 2026-07-02
feat(evaluator): unified outcome model + stripe matchers + golden gate (FDRS-591/611/597/646)

GaganSD:
## What & why

One coherent evaluator/scoring PR covering three linked tasks.

### T5 — Unified per-criterion outcome model (FDRS-591 + FDRS-611)
- Each criterion now carries an explicit four-state `outcome`: `passed | failed | skipped | errored`.
  - `skipped` — harness could not evaluate deterministically (no twin-plugin predicate matched, or no LLM judge configured; also judge auth/context-window limits).
  - `errored` — judge/infra failure (LLM 429/5xx, provider error).
- Reporting: `satisfaction = passed / (passed + failed)`. `skipped` and `errored` are excluded from the denominator but surfaced as explicit counts.
- All-skipped / all-errored runs render as **un-evaluated**, never as a misleading 0% (which previously read as a hard fail).
- **A5 inflation guard:** a scenario PASS now additionally requires every required criterion to have been evaluated (`can_pass`). "1 passed + 9 skipped = 100%" can no longer read as a pass — it is "un-evaluated (cannot pass)".
- Wire compatibility: `outcome` is additive + optional; `skipped: boolean` stays true for both skipped and errored so existing cloud consumers are unaffected. Added the optional field to `packages/shared-types` too.
- Threaded the guard through the self-host exit code (`runScenarioCore`), the matrix (`cellPassed`), and `pome inspect` / run rendering (new `scoreStatus` / `scenarioPassed` helpers, `✓/✗/-/!` markers).

### T6 — Stripe evaluator matchers (FDRS-597, M2 headline bug)
- The Stripe evaluator only recognised refund operations, so a correct agent scored 0% on scenarios 10–13.
- Added matchers for PaymentIntent create/amount/status, Stripe-shaped error surfacing (`invalid_request_error`), emitted events (`payment_intent.succeeded`, …), charge + balance-transaction creation, and the x402 402→200 challenge flow.
- Existing refund behavior (scenarios 14/19) is preserved verbatim. A correct final state now yields satisfaction 100 on 10–13.

### T7 — Golden-scenario CI gate (FDRS-646)
- New `[D]`-only deterministic test (`cli/test/golden/`) + workflow (`.github/workflows/golden-gate.yml`) that needs no live LLM (judge keys explicitly unset).
- Asserts a correct-agent fixture scores ≥ threshold (and passes the A5 guard) and a wrong-agent fixture scores below it, with the expected per-criterion breakdown — so a matcher regression that zeroes out a correct agent fails CI.

## A1 hosted-engine caveat (documented, not changed)
pome-cloud re-resolves scores with its own hosted judge (`apps/control-plane/src/routes/result.ts` ignores CLI scores; `services/judge.ts` computes passed/total). The new outcome semantics (skipped/errored excluded + the A5 cannot-pass guard) must be adopted cloud-side under **FDRS-618 (Part 2)**. Until then, local and hosted scores can differ for the same trace. This is called out in code comments near the formula (`score.ts`), in `shared-types/run.ts`, and in the hosted Score synthesis (`runScenarioHosted.ts`), which deliberately keeps the cloud score authoritative.

## Verification
- `bun run test test/unit/evaluator test/golden test/matrix/aggregate.test.ts test/integration/probabilisticJudge.test.ts` → 239 passed.
- New golden gate + Stripe + score tests: 57 passed.
- `bun run typecheck`: no new type errors from these changes (the only diagnostics in the sandbox are pre-existing missing-dependency resolutions — `commander`/`ai` — that also fail on `main` in this environment; the 7 failing e2e tests are the same pre-existing environment failures, unrelated to this diff).
- Added a changeset (`cli/.changeset/evaluator-scoring.md`) to satisfy the CLI version-bump gate.


GaganSD:
## Code review (automated paranoid pass) — findings

**Verdict:** No Critical. Core satisfaction math + A5 guard verified sound; 57 new + 312 consumer tests green; golden gate confirmed deterministic. Recommended fixes before merge:

**High**
- **H1 — hosted `pome run` prints PASS while `pome inspect` of the same run prints "un-evaluated (cannot pass)".** The self-host path routes through `scoreStatus`/`scenarioPassed`, but the hosted render path (`main.ts:~623/657`) still uses raw `satisfaction >= threshold`, while `runScenarioHosted.ts:~423` writes `can_pass=false` when the cloud judge marks any criterion skipped. Fix: route hosted rendering through `scoreStatus` too, or don't derive `can_pass` for hosted scores (cloud is authoritative). Exit-code bypass itself is acceptable (FDRS-618).
- **H2 — a transient judge 429/5xx (`errored`) now hard-fails the cell/run with no retry.** `can_pass` requires `errored===0`; one rate-limit on a required `[P]` criterion flips an otherwise-passing matrix cell to fail. Confirm intended, or add retry-on-errored before failing.

**Medium/Low**
- M1: matrix heatmap can show high satisfaction but 0 pass-rate for un-evaluated cells (arguably correct, confusing).
- L1: `report-html.ts:~123` buckets `errored` under `skipped` (display gap; new errored count invisible in HTML).
- L3: x402 matcher (`stripe.ts:~200`) has no ordering/`X-PAYMENT` check — `[200,402]` order passes; spoofed 200-only correctly fails. Add an ordering check or negative test.
- L4: golden gate catches false-negatives well but not all false-positives (an over-permissive matcher could stay green).
- N1: the `_error` matcher arm is broad (`/\b([a-z_]+_error)\b/`) and ordered early; safe for shipped 10–13 but order-fragile.

**Coordination**
- The `packages/shared-types/src/run.ts` edit is source-only and **redundant for the CLI** (the CLI enforces its own `cli/src/types/shared.ts`, which was updated). It **auto-merges cleanly** with #29 (disjoint regions, 0 conflict markers). **Merge #29 first**, then rebase this; since this PR doesn't bump the version, regenerate the vendored shared-types tgz with `--vendor` after #29 so it isn't stale.

PR #28, opened 2026-07-02
feat(cli): scenario catalog for all twins, exit-codes table, quickstart fix (FDRS-624/607/604)

GaganSD:
CLI front-door / quickstart fixes for launch readiness (PR 3 of 5). Three tickets.

## FDRS-624 — register all twins in the scenario catalog
`pome scenarios` only listed 5 github scenarios; the **stripe** (10–14, 19) and **slack** (20, 21) twins were invisible. Now all three twins and their bundled scenarios are registered.
- Added a **drift gate** test: every `.md` in the bundled `scenarios/` dir must be registered exactly once, so a new scenario file can't silently fall out of `pome scenarios` / `pome init`.

## FDRS-607 — exit-codes table
The README documented only `0–3` and mislabeled `2`. Replaced with the full `0–5` table matching `exitCodeFor` (`cli/src/hosted/errors.ts`) and the hosted docs, plus the `--local` nuance (exit 0 = trace captured, not scored).

## FDRS-604 — broken quickstart secret path (+ FDRS-647 smoke)
`docker compose` writes each twin's secret to a **per-twin subdir** (`.pome-data/github/secret`), but the bundled triage agent only probed the legacy flat `.pome-data/secret` and threw `Could not locate the twin auth secret`.
- Rewrote `readSecretFromDisk` as a testable `secretCandidatePaths()` that probes the per-twin path (`POME_TWIN`-aware) with the legacy layout as fallback.
- Guarded top-level `main()` with `import.meta.main` so the module is importable by the unit test without launching a run.
- Fixed the three stale `<repo-root>/.pome-data/secret` docstrings.
- **New `quickstart-smoke` CI workflow (FDRS-647):** the secret-path unit test + a `docker compose up` preflight that resolves the secret and reaches the twin `/healthz` (runs in `POME_PREFLIGHT=1`, no `ANTHROPIC_API_KEY`).

## Stale repo URLs
Pointed clone/report/bug URLs at `pome-sh/pome-twins` (root + cli READMEs, `cli/package.json`, the packaging-bug link in `skills.ts`, the scenario report link in `report-content.ts` + its two tests).

## Verification
- `cd cli && bun run typecheck` clean; affected unit + matrix tests pass (30).
- `cd examples/triage-agent && bun test` → 6 pass (secret-path candidates).

## Notes
- Scope: twin `package.json` / package README URLs are left to the twin-consolidation PR to avoid collisions. A pre-existing, unrelated `tsc` generic-variance warning in the example's tools array is out of scope (not run in CI).

Tickets: FDRS-624, FDRS-607, FDRS-604, FDRS-647.


GaganSD:
## Code review (automated paranoid pass) — findings addressed in-branch

Findings, all fixed in the follow-up commit:
- **Critical** — the `quickstart-smoke` job would fail every run: `preflight()` asserts `ANTHROPIC_API_KEY` is set (it only probes the twin, never calls Claude). Now passes a throwaway value so the token/twin probe actually runs.
- **High** — the twin GHCR image is private; added `packages: read` + a ghcr login before `docker compose up`. Also added a root `bun install` so the example's adapter dep resolves.
- **Medium** — exit code `3` is overloaded (hosted auth AND `--local` agent failure/timeout); README row now notes both.
- **Low** — dropped the `POME_TWIN` knob (agent is github-only; the override could mint a JWT with the wrong twin's secret). **Nit** — moved `withPome()` into the run guard so the module import is side-effect-free.
- Verified sound: catalog↔disk drift gate is real; exit-code rows 2–5 match `exitCodeFor`; secret-path resolves to `<repo-root>/.pome-data/github/secret` under `bun run start`.

PR #54, opened 2026-07-07
feat(sdk): twin-core engine — redaction, db wrapper, F-712 auth, sdk-boot contract proof (FDRS-681)

AFFFPupu:
## What
Grows `@pome-sh/sdk` into the one twin engine per FDRS-681: centralized redaction (recorder emits + `/_pome/state` are engine-redacted), a SQLite driver wrapper (`openTwinDatabase` + pragmas + same-shape `transaction()` — the F-703 node:sqlite swap becomes a one-file change), the full FDRS-712 auth mechanism (`verifyProviderToken`/`mintProviderToken` + `defineTwin()` shape options), JSON-RPC `POST /s/:sid/mcp`, contract-shaped `/healthz` (implementation + runtime block), per-twin unsupported-envelope hook, `state_delta` plumbing, and the non-loopback boot guard. Proof-of-harness: the slack twin assembled via `defineTwin()` (`contract/proof/slack-sdk-server.mjs`) passes the identical frozen FDRS-711 contract assertions (`contract/sdk-boot.test.mjs`; suite body extracted verbatim to `contract/suite.mjs`).

## Why
FDRS-681 (Better Digital Twins → M1 twin-core): every twin re-rolls ~14 infra modules and they've drifted — auth.ts ×4 with all hashes different, slack's recorder shipped without redaction. Auth spec of record is the FDRS-712 [DECISION]; rows 4/5 (raw bearer, sid-mismatch status) are pinned per twin so the upcoming ports (F-682/683/684) stay zero-wire-diff under the FDRS-711 suite.

## Notes for reviewer
- **FDRS-712 appendix bug fixed in the engine**: provider-token parsing is right-anchored (sig = trailing 22 b64url chars, optional exp segment, HMAC disambiguates) instead of the broken `([^_]+)` sid regex; regression tests cover an underscore-bearing sid and the sid-ends-in-`_<digits>` ambiguity. Minting is byte-identical to the pre-F-681 twin/cloud construction (compat test included).
- Mirror files are untouched (deleted in F-685/F-713); all mirror/copy-marker gates stay green. The engine's `redaction.ts` carries no mirror markers on purpose.
- `better-sqlite3` enters the sdk as an **optional peer dep**, loaded lazily via `createRequire` — the npm-pack clean-room smoke should be unaffected.
- Engine default auth envelopes render an expired JWT as `Bad credentials` (matching the frozen github wire — the old explicit `Token expired` branch was dead code, per FDRS-712 row 6); slack's `token_expired` comes from its `unauthorized(kind)` hook.
- The proof entry consumes built dists from `contract/proof/` — no twin package.json changes, so the CLI vendor pipeline is untouched.
- Engine `_pome/health` adds unasserted fields (`version`, `fidelity`) vs slack's frozen `{ok, twin}`; the suite is the freeze arbiter and stays green. Flagging for awareness.

## Review required
Yes — touches auth (engine now owns the FDRS-712 canonical mechanism) and the twin runtime contract surface (suite refactored into `contract/suite.mjs`, assertions unchanged).

### Founder briefing (3 min)
- **What changed** — the sdk is now the single twin engine (auth/redaction/db/MCP/admin), proven by booting slack through `defineTwin()` against the frozen FDRS-711 contract suite.
- **What it means for your repo / workflow** — pome-cloud sees zero wire change (contract suite green before and after, assertions untouched); the three twin ports can now start and should delete their infra files instead of forking them.
- **The one thing you must know** — provider-token sid parsing had a latent bug in all three twins (base64url sids containing `_` could never authenticate); the engine fixes it while staying byte-compatible with every token minted by today's cloud/CLI code.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -37,16 +37,20 @@
       },
       "devDependencies": {
         "@hono/node-server": "^1.19.6",
+        "@types/better-sqlite3": "^7.6.13",
         "@types/node": "^25.0.3",
+        "better-sqlite3": "^12.5.0",
         "tsx": "^4.21.0",
         "typescript": "^5.9.3",
         "vitest": "^4.0.15",
       },
       "peerDependencies": {
         "@hono/node-server": "^1.19.6",
+        "better-sqlite3": "^12.5.0",
       },
       "optionalPeers": [
         "@hono/node-server",
+        "better-sqlite3",
       ],
     },
     "packages/shared-types": {
--- a/contract/contract.test.mjs
+++ b/contract/contract.test.mjs
@@ -9,239 +9,24 @@
 // divergences that are themselves under review (see FDRS-712). Changing any
 // asserted status or shape is a contract change: update CONTRACT.md in the
 // same PR and coordinate the cross-repo pome-cloud PR per the contract doc.
+// The suite body lives in ./suite.mjs (FDRS-681) so the identical assertions
+// also run against the sdk-booted proof entry (./sdk-boot.test.mjs).
 //
 // Prerequisite: `bun run --filter '@pome-sh/shared-types' build:runtime` and
 // `bun run --filter '@pome-sh/twin-*' build` (the root `test:contract` script
 // chains all three).
 
-import { after, before, describe, it } from "node:test";
-import assert from "node:assert/strict";
-import { TWINS, mintSessionJwt, req, spawnTwin, spawnTwinRaw } from "./helpers.mjs";
-
-const SID = "contract-sid";
-const sPath = (p) => `/s/${SID}${p}`;
-
-// Per-twin frozen expectations, taken from live probes on 2026-07-07.
-const PER_TWIN = {
-  github: {
-    // /healthz carries fidelity + access_control on top of the shared core.
-    healthzFidelity: "semantic",
-    healthzExtras: ["access_control"],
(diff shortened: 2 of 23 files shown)


AFFFPupu:
**CodeQL check (the one red ✗): pre-existing pattern, routed to F-716 — not hot-patched here.**

The high-severity `js/polynomial-redos` alert points at the JWT + PEM scrub regexes in the new `packages/sdk/src/redaction.ts`. Those bytes are identical to the canonical redaction that already ships in 5 mirrored files on main (`cli/src/recorder/redaction.ts` + adapter + 3 twins) — the alert fires only because this PR adds a new (engine-canonical) copy, and CodeQL scans changed code.

I deliberately did not rewrite the regexes inside this PR: any linear-safe rewrite changes what the scrubber matches on boundary inputs (JWT glued mid-base64url-run, incomplete PEM blocks), and this PR's discipline is relocation-without-behavior-change. Exposure is bounded — redaction runs on local twin tape writes, not internet-facing input. The deliberate cross-copy fix (with a regression net proving unchanged redaction output) is https://linear.app/pome-sh/issue/F-716.

@GaganSD if you'd rather it fixed in-PR, say the word; otherwise dismiss/accept the alert and F-716 carries it.

PR #62, opened 2026-07-08
feat(twins): port twin-slack to thin twin-core plugin (F-683)

AFFFPupu:
## What

twin-slack is now a pure domain plugin on the `@pome-sh/sdk` engine: `twin.ts` declares a `defineTwin()` manifest, `server.ts` is a thin boot shim over engine `serve()`, and the per-twin harness modules (`auth.ts`, `recorder.ts`, `redaction.ts`, `mcp.ts`, `admin-gate.ts`, `app.ts`, `build-info.ts`) are deleted. Pilot-discovered engine gaps were fixed in `packages/sdk` tests-first (`pilot-gaps.test.ts`): ToolCallContext (session + reportDelta on all MCP dispatch surfaces), tool-list annotations + literal inputSchema override, `admin.forbidden` hook, recorder correlation/scenario-step stamping parity.

## Why

F-683 — the M1 pilot port: validates the engine API against a real twin before the github/stripe ports. The slack recorder once shipped without redaction (analysis §5.2); after this port redaction is structurally engine-owned and slack token/webhook shapes are asserted in `packages/sdk/test/redaction.test.ts`.

## Notes for reviewer

- **Zero wire diff proven both ways:** contract suite 44/44 at base `3cd86eb` (re-run in a throwaway worktree) and 44/44 on HEAD; sorted test-name diff identical. Full gauntlet green: twin-slack 212, sdk 131, workspace fan-out 1119, typecheck, all six lint gates.
- **F-712 pins declared, not forked:** `tokenResolvers` = Bearer + `?token=` + form-body; `allowRawBearer` stays engine-default false; `sidMismatch` → 401 `invalid_auth`; `token_expired`/`invalid_auth` classification via the `unauthorized(kind)` hook; minting through engine `mintProviderToken`.
- **One behavior ruling:** `/_pome/state` channel ordering flipped nondeterministically on `created_at` ms-ties — ruled a pre-existing export race (bug), fixed to insertion order (rowid). Contract suite pins zero wire diff.
- **Proof reconciliation:** `contract/proof/slack-sdk-server.mjs` (F-681) deleted as superseded; `contract/sdk-boot.test.mjs` now boots the real package entry (`node dist/src/server.js`).
- Domain keeps justified: `errors.ts` (Slack-shaped TwinError codes), `util.ts` (form-or-JSON parse, ts format, cursors), `db.ts` (DDL/reset only via `openTwinDatabase`, zero driver imports).

## Review required

[...]

PR #63, opened 2026-07-08
feat(twins): port twin-github to thin twin-core plugin (F-682)

AFFFPupu:
## What

twin-github becomes a pure domain plugin on the sdk engine, following the F-683 pilot pattern: harness modules deleted, `defineTwin()` manifest declares the F-712 pins, six engine additions land in `packages/sdk` (test-first, inert unless declared), and the deleted mirror copies are removed from the mirror/copy-marker gate lists.

## Why

F-682 — second M1 port. After it, twin-github carries domain only: entities, routes/serializers, seed schema, access-control catalog.

## Notes for reviewer

- **Stacked on #62 (F-683)** — review this PR's own commits only (`59f86c9`, `6781ae8`); merge bottom-up.
- Contract suite green before/after including all **62 MCP tools** via `GET /s/:sid/mcp/tools`; wire-faithful 4xx (FK-enforced fidelity) unchanged; determinism passes.
- **Pre-ruled dead code deleted:** github's explicit "Token expired" branch was unreachable (hono/jwt throws first; wire always said "Bad credentials") — deleting it is zero wire diff, pinned by the contract suite.
- **Residuals recorded, not fixed** (per F-712 "record residuals only"): `/_pome/state` fetches and JSON-RPC unknown-tool calls now appear on the recorder tape; sid-less signed JWT classifies as invalid ("Bad credentials") instead of "Forbidden" (both 401, unpinned corner).
- `serializers.ts` left at 790 lines — the ticket's split condition was the code-health tripwire, which does not fire.
- `access-control.ts` kept as domain (frozen `GET /s/:sid/_pome/access-control` route via the engine's `pomeRoutes` hook).

## Review required

Yes — auth pins (raw-bearer 401, sid-mismatch 401 pinned via options) and engine surface growth.

### Founder briefing (3 min)

- **What changed** — twin-github now declares its auth/envelope/admin behavior as `defineTwin()` options; six new engine capabilities exist but are inert unless a twin declares them.
- **What it means for your repo / workflow** — same contract, same wire bytes; cloud snapshot builds resolve the sdk dep exactly as for slack.
- **The one thing you must know** — two unpinned wire corners changed classification (recorded above); if cloud ever asserted on the sid-less-JWT "Forbidden" string, it now reads "Bad credentials".


AFFFPupu:
Fix round after the pre-merge review (48 verified findings, 15 blocking) + reconciliation with M6 which landed on main mid-flight. This branch was rebased onto current main and force-pushed; re-review from scratch is fair.

What changed on this lane:
- Rebase absorbs main/M6: twin package.json now the published-dep shape, Dockerfile already ships the sdk (main 3cdd34b), mirror scripts already deleted on main.
- CLI conformance: local Recorder interface replaces the twin type; standalone server signs with the env-pinned authSecret instead of the deleted resolveAuthSecret re-export. cli-ci tarball flow: typecheck clean, 541/541.
- Temporary file-size header on server.ts; the real split lands in #66.

PR #64, opened 2026-07-08
feat(twins): port twin-stripe to thin twin-core plugin (F-684)

AFFFPupu:
## What

twin-stripe (the largest twin, ~41 files) becomes a domain-only plugin: harness modules deleted, `domain/` + `routes/` + `x402`/`api-keys`/`session` kept as domain, `domain/index.ts` brought under the barrel policy. **failure-injection graduates into the engine** (`packages/sdk/src/failure-injection.ts` + tests) as a generic twin capability; rule payloads stay twin-domain.

## Why

F-684 — third M1 port. Also settles the graduation rulings the ticket demanded for idempotency and failure-injection.

## Notes for reviewer

- **Stacked on #62 (F-683)** — review this PR's own commits only (`faffbeb`, `fe93b23`); merge bottom-up.
- **Graduation rulings:** failure-injection → engine (generic "response lost on attempt N" mechanism for any provider double); idempotency → stays stripe domain (Stripe-Idempotency-Key semantics are wire behavior); x402/api-keys stay domain per the founder pre-ruling.
- **F-712 pins declared:** `allowRawBearer=true` (raw + case-insensitive bearer), `sidMismatch` → 403 with the frozen stripe forbidden envelope, `resolveCredential` via the `api_keys` table, `requirePathSid=false` + `mountSessionAtRoot` (F3 `/v1/*` SDK-compat), stripe `errors.ts` envelope.
- **Stale ticket reference:** the "scenario-10–13 golden tests" named in the Done-when were deleted repo-wide by FDRS-656/657 (capture-only CLI, PR #34) before this milestone — nothing to run; recorded as a [DECISION] on the Linear ticket.
- Stripe-specific engine gaps (mountSessionAtRoot, sessionHealthz=false → 501 catch-all, bearer-resolved session in the state hook, pomeHealth extras) landed test-first in the sdk.

## Review required

Yes — auth (liberal-bearer pin, DB-backed credential resolution) and failure-injection moving into the shared engine surface.

### Founder briefing (3 min)

- **What changed** — stripe declares its liberal auth + root-mount behavior as options; failure-injection is now an engine capability every twin can declare.
- **What it means for your repo / workflow** — cloud twins can get failure-injection without stripe-specific code; stripe wire behavior is byte-frozen by the contract suite.
- **The one thing you must know** — the scenario-10–13 golden gate the ticket references no longer exists in the repo (removed with the local evaluator in PR #34); its coverage now lives in the 144-test stripe suite + contract suite.


AFFFPupu:
Fix round after the pre-merge review (48 verified findings, 15 blocking) + reconciliation with M6 which landed on main mid-flight. This branch was rebased onto current main and force-pushed; re-review from scratch is fair.

What changed on this lane:
- Failure-injection + idempotency moved to the engine session-middleware slot so they cover MCP dispatch again — an Idempotency-Key MCP create replays (was minting duplicates) and injected rules on /mcp/* fire; pinned by test/mcp-middleware.test.ts.
- bodyReader restores both pre-port readers (admin/seed strict-JSON-with-null-default; MCP form-or-JSON); admin.recorded=false (stripe admin traffic never hit the tape); unknown api-key bearers answer the frozen 401 Invalid API Key provided. (+looksLikeApiKey perf gate before the DB); bounded 10k recorder with real dropped counters in _pome/health.
- CLI stripe harness rewritten for the engine factory; cli-ci tarball flow 541/541 incl. the scenario-14 injection path.

PR #65, opened 2026-07-08
feat(cli): vendor sdk into CLI; local twin boot stays green (F-713)

AFFFPupu:
## What

`@pome-sh/sdk` now ships inside the published CLI via the existing vendor-tarball pipeline: `scripts/pack-publishable.mjs` gains a `--vendor-cli` mode (exact-version rewrite of the sdk's shared-types dep, packs adapter after sdk), `verify:vendor` covers the sdk tarball, and `cli` + `packages/adapter-claude-sdk` consume redaction from the sdk — `cli/src/recorder/redaction.ts` (the old canonical) and the adapter mirror are deleted. A packed-CLI smoke job (pack → install into temp dir → local twin boot → `/healthz` 200) runs in `cli-ci.yml`.

## Why

F-713 — M1's "distribution solved" leg. The mirrors existed *because* twins were vendored as `file:` tarballs that couldn't depend on the sdk; this closes that loop until F-686 publishes `@pome-sh/*` for real.

## Notes for reviewer

- **Stacked on #62 (F-683)** — review this PR's own commit only (`5fc5c17`); merge bottom-up.
- Packed-artifact verification is real: `pome run --local` from the packed set boots engine-based twin-slack, `/healthz` 200. Note the frozen slack healthz has **no** top-level `fidelity`/`version` (version lives under `runtime.version`) — the smoke asserts their absence, since emitting them would be a wire diff.
- `packages/sdk` peer range for `@hono/node-server` widened to `^1.19.6 || ^2.0.0` so the vendored engine installs against the CLI's `^2.0.6` — manifest-only, contract suite untouched.
- CLI version gate satisfied via changeset (`cli/.changeset/f-713-vendor-sdk-engine-twin-slack.md`, patch).
- DX footgun documented: raw `bun test` in `cli/` (Bun runner) fails on better-sqlite3 dlopen; the canonical suite is `bun run test` (vitest, 520/520) — exactly what CI runs.

## Review required

Yes — self-host packaging (vendor pipeline + published artifact layout).

### Founder briefing (3 min)

- **What changed** — the sdk engine is now a vendored tarball inside the published CLI, resolved by the same exact-version hoist as shared-types.
- **What it means for your repo / workflow** — cloud snapshot builds and the GHCR workflows must resolve the twins' sdk dep the same way (that's F-714, your cross-repo check).
- **The one thing you must know** — the sdk's hono/node-server peer range was widened to span the CLI's v2; if cloud pins a different hono server major, the vendored install is where it would surface.


AFFFPupu:
Closing: superseded by M6 ("published contract", merged to main mid-flight — PR #61 + the 0.7.0 release chain). M6 removed cli/vendor entirely, publishes @pome-sh/* to npm (cli-ci resolves local tarballs pre-publish), and centralized redaction in @pome-sh/shared-types/redaction — which covers this ticket's vendor-tarball mechanism, verify:vendor coverage, and the cli/adapter redaction-mirror deletion wholesale. The one surviving concern (the CLI conforming to the engine-based twins) landed on the port lanes themselves: F-683 (slack recorder typing), F-682 (github Recorder type + resolveAuthSecret), F-684 (stripe harness rewrite) — each verified against the cli-ci tarball flow (541/541). [DECISION] recorded on F-713.

PR #163, opened 2026-07-17
fix(cli): exclude preflight-probe signals from the hosted usage ledger

AFFFPupu:
## What
On the hosted lane, `runScenarioHosted` runs the agent command twice against one shared signals file — a ≤10s preflight probe, then the real run — and uploads the file whole. The preflight's `LlmTurnEvent` (and step/tool_call) signals were landing in the uploaded `signals.jsonl`, double-counting per-turn LLM usage. This truncates the shared signals file after a successful preflight, before the real run, so the uploaded blob counts real-run telemetry only.

## Why
The uploaded signals blob feeds the usage ledger. A real run emitting 2 turns produced 4 rows in the blob (2 preflight + 2 real). `POME_PREFLIGHT=1` is only honored by scaffolded/fixture agents, so any real adapter-wrapped agent inflates its turn usage. Fixing it capture-side (runner) protects every agent regardless of what it emits; cloud-side dedup isn't viable because preflight turns carry distinct `event_id`s. Surfaced by the F-768 hosted-upload assertion, which previously *tolerated* the double-write by asserting distinct-turn identity — this PR tightens that assertion to an exact count.

## Notes for reviewer
- The truncate is gated on `preflight.exitCode === 0` and sits before the real `runAgentCommand`: the preflight-failed → single-run path is intentionally unchanged (it still uploads the preflight's signals).
- `runAgentCommand` awaits child-process exit, so the preflight's writes are flushed before truncation — no race.
- Test-first: the tightened e2e assertion (`toHaveLength(2)`, `turn_index [0,1]`) goes red on the unpatched runner (blob has 4) and green after.

## Review required
Yes — touches the hosted usage/billing ledger (per-turn LLM usage counting).

### DIFF
--- a/cli/.changeset/preflight-signals-ledger.md
+++ b/cli/.changeset/preflight-signals-ledger.md
@@ -1,0 +1,5 @@
+---
+"@pome-sh/cli": patch
+---
+
+Hosted runs no longer count the preflight probe's telemetry toward the uploaded usage ledger. The runner ran the agent command twice against one shared signals file (a ≤10s preflight probe, then the real run) and uploaded the file whole, so per-turn LLM usage (`LlmTurnEvent`) was double-counted. The shared signals file is now truncated after a successful preflight, before the real run, so the uploaded `signals.jsonl` reflects real-run telemetry only.
--- a/cli/src/runner/runScenarioHosted.ts
+++ b/cli/src/runner/runScenarioHosted.ts
@@ -331,6 +331,14 @@ export async function runScenarioHosted(
 
     let agentResult = preflight;
     if (preflight.exitCode === 0) {
+      // F-771 — the preflight probe appended its own signals (turns/steps/
+      // tool_calls) to the shared POME_ADAPTER_SIGNALS_PATH file. Those must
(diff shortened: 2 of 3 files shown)


AFFFPupu:
Addressed automated review feedback:

- **Greptile P2 (numeric sort):** fixed — `turn_index` assertion now uses `.sort((a, b) => a - b)` so it stays correct for indices ≥ 10 (`b276043`).
- **MergeProtect HIGH "destructive data operation" (×2):** this is the intentional `writeFile(signalsPath, "")` truncation, which *is* the fix. It targets a per-run temp file (`mkdtemp` signals path), is gated on `preflight.exitCode === 0`, and runs after `runAgentCommand` has awaited the preflight child's exit (writes flushed) — no persistent data or cross-run state is touched.
- **MergeProtect MEDIUM "bug-fix lacks test changes":** the PR does include the test change — `runScenarioHosted.test.ts` tightens the F-768 assertion from distinct-turn identity to an exact count (red→green).
- **MergeProtect MEDIUM "no preflight-failed-path test":** the truncate is gated on preflight success and deliberately leaves the preflight-failed → single-run path unchanged; that path is out of scope for this fix.

PR #197, opened 2026-07-22
Hero example: support-triage local examinee (F-849)

AFFFPupu:
Executive summary: The hero example gains a locally-runnable examinee — a minimal Claude-Agent-SDK agent the coach spawns after `run_task`, wired to the hosted twins via env-injected per-session MCP URLs. Ships the failing v1 triage rule as the default with the v2 one-line fix as the documented swap point, so the fail→fix→green story runs on a laptop with zero API keys (rides Claude Code credentials).

## What
`examples/support-triage/local/`: `src/index.ts` (SDK 0.2.141, twins attached as HTTP MCP servers with the session bearer; closed-book — WebSearch/WebFetch disallowed; `POME_PREFLIGHT` honored), env-contract unit tests (5), standalone package (degit-able; registry-only deps), README with the coach-spawn flow and the "33, not 0" fix story; parent README gains a Local examinee section.

## Why
M0's first-report walk runs the hero example against a LOCAL examinee (paste-prompt, hands-off); only a managed-agent YAML pair existed.

## Notes for reviewer
- Env contract (`POME_GITHUB_MCP_URL` / `POME_SLACK_MCP_URL` / `POME_AUTH_TOKEN` / optional `POME_TASK`) verified against both CLI runners' `buildAgentEnv` — the example runs unmodified under coach-spawn AND `pome run --agent`.
- `TRIAGE_RULE_V1/V2` are verbatim from the agent YAMLs (v2 taken from the YAML, not the ticket's paraphrase).
- Deliberately no `@pome-sh/adapter-claude-sdk` / `TWIN_AUTH_SECRET`: with external HTTP MCP servers the adapter's fetch hook would capture nothing, and registry-only deps keep the folder degit-able.
- Not live-run against a minted session (cannot mint here) — the first coach run is the real proof; flagged for the M0 walk.

## Tests / CI
local/: typecheck clean, 5/5 unit tests. Root: lint:legacy-markers, gate:no-eval, lint:dead-code all pass. Zero "scenario" in new prose (path refs to the existing scenarios/ file excepted; F-825 owns that rename).

Closes F-849


AFFFPupu:
Re the two Greptile P1s: (1) **allowedTools prefix** — fixed in 32440cb by removing the line entirely: under `permissionMode: "bypassPermissions"` an allowlist is redundant (everything is pre-approved), so the speculative prefix-vs-exact-name failure mode is eliminated rather than argued about; the closed-book `disallowedTools` deny-list stays. (2) **Preflight `/tools` probe** — refuted by precedent: `${MCP_URL}/tools` is the established probe in `examples/triage-agent/src/index.ts:277` ("The authenticated ${MCP_URL}/tools probe below already covers reachability + auth in every mode"), and triage-agent runs live under hosted `pome run` since the FDRS-667 fix using exactly this probe. Same convention, same route shape.

PR #234, opened 2026-07-24
Register agent sends manifest twins to POST /v1/agents

AFFFPupu:
## Summary

`pome register agent` and the run-path auto-create resolver both `POST /v1/agents` but only ever sent the `--twins` **flag**. A manifest declaring `twins: ["gmail"]` never reached the server, so the server's `github` default won — and the first `pome run` errored with `Requested twins are not enabled on agent '…': gmail`. This is an onboarding-funnel break for every curriculum example on a non-github twin (the cold-run walk).

Both POST sites now send the manifest's declared `twins`.

## What changed (`cli/`)

- **`project-config.ts`** — new `normalizeManifestTwins()` (trim + lowercase + de-dupe of `manifest.twins`); the shared normalizer for both POST paths. Manifest twins are normalized but not validated against `MOUNTED_TWINS` here — the schema keeps `twins` open and the server returns a friendly error for an unknown twin.
- **`register.ts`** — `runRegisterAgent` sends `mergeRegisterTwins(manifest.twins, --twins)` — the **union** of the manifest's twins and any flag additions. Union is correct because the server merges additively (there is no "reduce" to preserve), so re-register additively merges as required. The older-cloud "services not reported" warning now keys off the effective twins.
- **`agent-identity.ts`** — the register-on-first-run path (a fork / first `pome run` without explicit register auto-creates the agent) now also sends the manifest's twins. Safe/idempotent since the server merges additively.
- **`main.ts`** — `--twins` help text updated: unioned with the manifest; default is the manifest's twins, else the cloud default.

## Tests

`cli/test/unit/` — register carries manifest `twins` with no flag; flag ∪ manifest union; `mergeRegisterTwins` unit table; run-path re-resolve sends manifest twins. All three F-926 assertions were confirmed to go **red** when the wiring is reverted, then green after.

## Verification

- `npm run typecheck` — clean
- `npm run build` — clean
- Full unit suite — **647 passed**
- Patch changeset: `cli/.changeset/register-manifest-twins.md`

### DIFF
--- a/cli/.changeset/register-manifest-twins.md
+++ b/cli/.changeset/register-manifest-twins.md
@@ -1,0 +1,5 @@
+---
+"@pome-sh/cli": patch
+---
+
+`pome register agent` now sends the manifest's `twins` to `POST /v1/agents`, so the cloud agent's enabled services match the manifest instead of falling back to the server's `github` default. Previously a manifest like `twins: ["gmail"]` was ignored and the first `pome run` errored with `Requested twins are not enabled`. Any `--twins` flag is unioned with the manifest's twins (the server still merges additively).
--- a/cli/src/cli/agent-identity.ts
+++ b/cli/src/cli/agent-identity.ts
@@ -23,7 +23,7 @@ import {
   resolveCachedAgentId,
   writeLinkCache,
 } from "./link-cache.js";
-import { readManifest } from "./project-config.js";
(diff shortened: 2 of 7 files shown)


AFFFPupu:
Closes [F-926](https://linear.app/pome-sh/issue/F-926/pome-register-agent-ignores-manifest-twins-enabled-services-defaults)

PR #237, opened 2026-07-25
chore: unify cli into the root npm workspace and privatize internal packages

GaganSD:
## Summary

Collapses the two-lockfile / nine-package layout into one npm workspace and makes every package except `@pome-sh/cli` and `@pome-sh/adapter-claude-sdk` internal. This is Lane A of the packaging restructure; Lane C (twin registry) and Lane D (tsup + single release pipeline) stack on top.

**Publishing is frozen from this PR until Lane D's `release.yml` lands.** See the freeze section below — it did not go the way the plan assumed, and there is a new guard.

## What changed

**Workspace unification**
- root `workspaces: ["packages/*", "cli"]`; one regenerated `package-lock.json`; `cli/package-lock.json` and `cli/pnpm-workspace.yaml` deleted.
- every internal `@pome-sh/*` dependency → `"*"` (workspace-resolved). The exact pins had drifted: `sdk` and all five twins pinned `@pome-sh/shared-types@0.12.0` against a local 0.12.2, so npm installed a **second registry copy** of shared-types and the process carried two zod schema identities for the same wire types. The new lockfile has zero nested `@pome-sh` copies (`grep -c 'packages/.*/node_modules/@pome-sh' package-lock.json` → 0).
- root `build` chain ends with `-w @pome-sh/cli`.
- `cli/package.json` loses `prepare: npm run build` — it would fire on every root and production-only install now that cli is a workspace member.

**Privatization**
- `@pome-sh/{sdk,shared-types,twin-github,twin-slack,twin-stripe,twin-gmail,twin-linear}` → `private: true`, and lose `publishConfig`, `bin`, and the wildcard `"./*"` export. The CLI only ever imported named exports plus `@pome-sh/sdk/server`; `tsc` confirms.
- the twins' Docker `CMD node packages/twin-*/dist/src/server.js` never used the `bin` entry, and the contract suite spawns the dist directly — no boot surface moved.

**Deletions**
- `scripts/pack-publishable.mjs`, `scripts/use-local-pome-tarballs.mjs` (pack-time manifest rewriting: the committed manifests used to be uninstallable), `PACKAGE_RELEASE.md`, `.github/workflows/sdk-publish.yml`.

**Twin Dockerfiles (all five)**
- `COPY cli/package.json cli/package.json` — the lockfile now names that workspace, so `npm ci` needs the manifest present.
- the install is scoped: `npm ci --include-workspace-root -w @pome-sh/shared-types -w @pome-sh/sdk -w @pome-sh/twin-<t>`, so no twin image pulls the CLI's `@ai-sdk`/`ai`/`@anthropic-ai` graph into a build stage that never uses it.
- `twin-image.yml` gains a `cli/package.json` path filter and treats it as a shared input (full matrix).

**Workflows**
- `cli-release.yml`: push-to-main trigger removed (freeze), `workflow_dispatch` kept, reason + hotfix path documented in the file.
- `cli-ci.yml`: one root `npm ci`, no pack/tarball rewrite, no "publish the package batch first" gate.
- `quickstart-smoke.yml`: the pack + `use-local-pome-tarballs` + nested `npm ci` block collapses to `npm run build`. **The twin-start / health / gmail steps are byte-identical.**
- `agent-trace-overhead-gate.yml`: same collapse.
- `ci.yml`: `cli/` added to the heavy-suite path scope (root `build`/`typecheck`/`test` now cover the CLI).
- `renovate.json`: stale `sdk-publish.yml` reference → `cli-release.yml`.
- `scripts/check-first-party-twin-registration.mjs`: drops the `pack-publishable.mjs` and `sdk-publish.yml`/`cli-release.yml` npm-pin registration seams (those seams no longer exist). Still gates the other eight.

**CI ordering/robustness fixes forced by the layout**
- `cli-ci.yml` builds the workspace **before** typechecking the CLI. The CLI resolves its siblings through their built dists (`@pome-sh/sdk/server` → `dist/server.d.ts`); it used to get those declarations from packed tarballs that already shipped a dist, so a bare `tsc` worked. With a workspace link it does not — `typecheck` before `build` fails with TS2307 for every `@pome-sh` specifier. (`ci.yml` already had the right order.)
- the pack step captures `tar` listings to variables instead of piping into `grep -q`. `grep -q` closes the pipe on its first match, tar takes SIGPIPE and reports `stdout: write error`, and under `set -o pipefail` that fails the step even though the assertion passed — the same trap `scripts/check-cli-version-bump.sh` already documents.
- the three workflows that run the built CLI emit `@pome-sh/shared-types`' in-place runtime JS first (`build:runtime`). shared-types still exports raw `.ts` whose internal specifiers are `./x.js`, so the built CLI cannot import it until those files exist. The old pipeline got them for free inside the packed tarball; a workspace link does not. `contract/run.mjs` already did this before the black-box suite. (Lane D removes the need for the CLI path entirely — the bundle inlines shared-types.)

**Other incidental fixes forced by the layout**
- `tsx` is resolved through its own manifest (`cli/scripts/lib/resolve-tsx.ts`) instead of a hardcoded `cli/node_modules/.bin/tsx`, which npm no longer creates (hoisted to the repo root). Affected `test/e2e/twinStart.test.ts`, `scripts/overhead-gate.ts`, `scripts/cas-adapter-acceptance.ts`.
- `exportGitHubCloneState` gets an explicit `Promise<unknown>` return type — TS2742, exactly as the file's existing comment predicted for its sibling: the inferred type names the twin's own `dist/src/types.js` through the new workspace symlink.
- `TODOS.md` added with the two approved follow-ups (pome-cloud migration off the internal npm packages; framework-agnostic trace correlation).

## ⚠️ Deviation from the plan: the bundleDependencies fallback state does not exist

The plan asked to verify that `npm pack -w @pome-sh/cli` from the unified workspace still produces an installable tarball with `bundleDependencies` intact, "so A alone is a shippable state if C/D stall". **It does not, and the failure is silent.**

Measured:

```
$ npm pack -w @pome-sh/cli --ignore-scripts
pome-sh-cli-0.8.0.tgz          228 KB   (baseline published tarball: 15.2 MB unpacked)
$ tar -tzf pome-sh-cli-0.8.0.tgz | grep -c 'node_modules/@pome-sh'
0
$ cd /tmp/room && npm install .../pome-sh-cli-0.8.0.tgz
added 26 packages in 4s          # install SUCCEEDS
$ node node_modules/@pome-sh/cli/dist/src/cli/main.js --help
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@pome-sh/sdk' imported from
  /tmp/room/node_modules/@pome-sh/cli/dist/src/recorder/artifacts.js
```

Cause: `bundleDependencies` packs from the package's **own** `node_modules`. Once `cli/` joins the workspace, npm hoists the workspace links to the repo-root `node_modules` and `cli/node_modules/@pome-sh/` is empty, so npm packs a tarball that *declares* seven bundled deps and ships none. npm then trusts the declaration and does not fetch them at install time — so the install reports success and the failure only lands on the user's first command.

Rather than reintroduce pack-time machinery this lane exists to delete (and which Lane D removes anyway), this PR makes the failure impossible to ship:

- new `cli/scripts/assert-publishable.mjs`, wired as cli's `prepublishOnly`. It refuses any tarball with an `@pome-sh/*` runtime dependency that is not both declared in `bundleDependencies` and materialized in `cli/node_modules/`. It passes trivially in Lane D's end state (zero `@pome-sh/*` runtime deps), so it is a durable invariant, not scaffolding.
- `cli-ci.yml`'s pack step now asserts the guard **refuses** the tarball, so the freeze cannot be bypassed by an `npm publish` from a green build. The clean-room boot check returns in Lane D against the bundled tarball.

**Net effect on the plan:** Lane A alone is *not* a shippable fallback state. The only hotfix path while the freeze holds is outside-voice #10 — publish by hand from the pre-restructure `@pome-sh/cli@0.8.0` tag, whose tree still carries the exact-pin manifest and standalone `cli/` install. That is documented at the top of `cli-release.yml` and in the guard's own error output.

## Verification

| Check | Result |
|---|---|
| `npm install` (fresh, no lockfile) | 411 packages, 0 nested `@pome-sh` copies in the lock |
| `npm ls @pome-sh/twin-github` | `-> ./packages/twin-github` (workspace link) — same for all 9 |
| `npm run build` (packages + cli) | pass |
| `npm run typecheck` (all workspaces incl. cli) | pass |
| `npm run test:contract` | **109 tests, 104 pass, 0 fail, 5 skipped** |
| `npm run lint:dead-code` (knip) | pass |
| `npm run gate:no-native` | pass — 55 production packages clean |
| `npm run gate:no-eval` | pass |
| `check-first-party-twin-registration.mjs` | pass — github, slack, stripe, gmail, linear |
| copy-markers / code-health / no-cloud-imports | pass |
| `check-cli-version-bump.sh` (0.8.0 → 0.8.1) | pass |
| `check-cli-version-floor.sh` | pass — 0.8.1 ≥ npm latest 0.8.0 |
| **local Docker build, twin-github** | image built from the new Dockerfile, container boots, `GET /healthz` → `{"ok":true,"twin":"github","tools":65,...}` |

`npm test -w @pome-sh/cli`: 688/689 pass. The one failure is `test/e2e/twinStart.test.ts`, and it is **pre-existing local flake, not caused by this PR** — the same suite on `origin/main` in a separate checkout fails 3 e2e tests on this machine (`runTaskAdapterSignals`, `runTaskEgress`, `twinStart`) from parallel-server contention. `twinStart.test.ts` passes in isolation here (1/1). CI runs the suite on a dedicated runner.

## Post-merge

This merge rebuilds the rolling GHCR twin tags (`package-lock.json` + `packages/*/package.json` + Dockerfiles all changed), so it is prod-adjacent: **run the pome-cloud smoke against the fresh `ghcr.io/pome-sh/twins:<twin>` tags after merge** (outside-voice #6 / T4).


GaganSD:
Reopening to force CI re-trigger — pull_request workflows silently failed to fire on the last two pushes (only CodeQL + Greptile ran, confirmed via the Actions API).

PR #246, opened 2026-07-27
Span-lane output tokens come from message_delta, not the message_start snapshot

AFFFPupu:
## Summary

`gen_ai.usage.output_tokens` was the `message_start` snapshot — the count at the moment the stream envelope was cut, not the finished turn. The tell is `stop_reason: null` on every `assistant` message; a completed Anthropic message always has one.

The finished value exists only on the `message_delta` stream event, which the SDK emits only when `query()` was called with `includePartialMessages: true` — the **caller's** option. So `query()` now sets it whenever a telemetry lane is configured, and filters the consequences back out so an agent author's `for await` loop is unchanged.

| `02-injection-issue-body`, hosted | before | after |
| -- | -- | -- |
| `agent_tokens_out` | **31** | **1059** |
| `agent_tokens_in` | 18 → 155,330 (F-994) | 155,330 |
| span count | 9 → 4 (F-994) | 4 |

## The flag adds two message kinds, not one

Alongside `stream_event` comes a `system/status` message with `status: "requesting"`, one per API turn. Proven twice:

1. An A/B on a deterministic single-turn prompt — 0 occurrences without the flag, 1 per turn with it, two runs each way.
2. The `claude` binary itself, where one gate guards both emissions:

```js
case"stream_request_start": if(H) yield {type:"system",subtype:"status",status:"requesting"}
...                        if(H) yield {type:"stream_event", event:...}
```

The sibling `case"sdk_status"` path — `compacting`, the permission-mode `status:null`, `compact_result` / `compact_error` — is **not** behind that gate, so those reach an agent author with or without pome. Filtering every `system/status` would have broken pass-through fidelity in the other direction; the filter matches `requesting` only.

## Two things that would have been silent regressions

- **`message_delta` carries no message id.** It comes from the `message_start` that opened the message. Parallel Task subagents interleave their stream events with the main agent's and `parent_tool_use_id` is the only separator, hence one open-message slot per stream rather than a single global.
- **Injected messages must not move the turn boundary.** They land microseconds before the assistant message they describe, so letting them advance it would collapse every span window to ~0 ms and undo F-994's latency fix. The same guard goes into `withToolEvents`, where a subagent stream event carries enough `parent_tool_use_id` to mint a spurious `SubagentSpawnEvent`.

`finish_reasons` on the `LlmTurnEvent` lane is fixed by the same event — it had always been null in practice, because it read the assistant message's `stop_reason`.

## Verification

Live, through the **built** adapter on a 4-turn run (`.context/verify-f998.mjs`), all seven checks green — and repeated four times on independent runs, exact every time (557, 655, 629, 611):

```
span output_tokens == result rollup                 (557 vs 557)
LlmTurnEvent output_tokens == result rollup         (557 vs 557)
span input_tokens == result rollup (F-994 intact)   (126966 vs 126966)
no partial-message artifacts leaked                 (none)
one span per LlmTurnEvent row                       (4 vs 4)
every turn has a finish reason                      (tool_use x3, end_turn)
span windows not collapsed                          (428, 2948, 5393, 5340 ms)
```

Hosted: `run_cE3D58umjFbI9H3B`, 100/100 PASS, `agent_tokens_out` 1059, span count 4, `agent_error_count` 0.

Tests: 141 in this package, 1586+ across all 8 workspaces, contract suite 104/104. **Mutation-checked 15/15, 0 survived** — every guard in the diff fails at least one test when broken, in both directions for the filter. That sweep found a real hole in my own tests, written up in a comment below.

**Scope of the exactness claim.** It holds over **main-agent** turns. `SDKResultMessage.usage` counts only the main agent, and the SDK forwards no sub-agent stream events, so sub-agent turns keep the snapshot and sit outside the rollup on both sides. Live two-sub-agent run: main deltas `517 + 7 == 524 == result`. Details and the two residuals that fall out of it are in a comment below; neither is a regression here (F-994 shows the same asymmetry on the input side).

## Reviewer notes

- **`agent_model = unknown`** in the hosted rollup is pre-existing — the F-994 baseline run `run_TtKKRvLu3qzj5IXt` shows the same. Not touched here.
- **Stream-volume cost**, measured live: messages x3.3–3.6, bytes x1.3–1.4. Well short of the order of magnitude the ticket budgeted for — the CLI batches content deltas coarsely, so a 667-token text turn produced 29 stream events, not 667.
- **`POME_DISABLE_PARTIAL_MESSAGES=1`** reverts to the old behaviour without a release — exercised live, and it doubles as a clean A/B isolating the fix: same build, same script, `11` vs `629` output tokens, with input attribution and stream fidelity unchanged either way.
- Injection is gated on a telemetry lane being configured, matching how both lanes are already inert without one — so standalone dev pays nothing.
- No changeset: that gate is `cli/`-only. No version bump, matching F-994 — `packages/` ships in a separate `packages-v*` batch and this adds no public API surface.

### DIFF
--- a/packages/adapter-claude-sdk/README.md
+++ b/packages/adapter-claude-sdk/README.md
@@ -82,6 +82,15 @@ side-effect import is captured in the `[DECISION]` comment on
   them. Iteration is what hero tasks use today.
 - **`withPome()` is idempotent.** A second call is a no-op; the global fetch
   replacement is installed exactly once per process.
+- **`query()` turns on `includePartialMessages` when telemetry is on**, because
+  a turn's finished output-token count reaches the SDK only on a `message_delta`
+  stream event — the `assistant` message carries a `message_start` snapshot
+  roughly 5x too low. Every message that option adds is filtered back out before
+  it reaches your iterator, so your `for await` loop sees exactly what it saw
+  before; a test asserts the yielded sequence is identical with telemetry on and
+  off. Set the option to `true` yourself and nothing is filtered. Measured cost:
+  ~1.3–1.4x the stream bytes. `POME_DISABLE_PARTIAL_MESSAGES=1` opts out, at the
+  price of the low number.
 
 ## Signals file format
 
--- a/packages/adapter-claude-sdk/README.md
+++ b/packages/adapter-claude-sdk/README.md
@@ -91,6 +91,14 @@ side-effect import is captured in the `[DECISION]` comment on
   off. Set the option to `true` yourself and nothing is filtered. Measured cost:
   ~1.3–1.4x the stream bytes. `POME_DISABLE_PARTIAL_MESSAGES=1` opts out, at the
   price of the low number.
+- **Sub-agent turns still report the snapshot.** The SDK forwards no stream
+  events for Task sub-agents (measured: 34 stream events on a two-sub-agent run,
+  none of them tagged with a `parent_tool_use_id`), so there is no
+  `message_delta` to read and those turns keep the low count. They are a small
+  share of a run's output. Note also that `SDKResultMessage.usage` counts only
+  the main agent, while both lanes emit a row per turn either way — so on a run
+  that uses sub-agents the lanes total *more* than the SDK's own rollup rather
+  than matching it exactly.
 
 ## Signals file format
 
--- a/packages/adapter-claude-sdk/src/genai-spans.ts
+++ b/packages/adapter-claude-sdk/src/genai-spans.ts
@@ -22,12 +22,21 @@
 // every id ever seen: a resumed session can legitimately replay a message id,
 // and that is a real second turn, not a duplicate block.
 //
+// Token attribution splits across the two directions. Input is read off the
+// assistant message (`totalInputTokens` below). Output is NOT there — the
+// assistant message carries a `message_start` snapshot, ~5x low — so it comes
+// from the `message_delta` stream event via MessageDeltaTracker, with the
+// snapshot kept only as a fallback for a stream that ended before its delta
+// (F-998; see partial-messages.ts for how those events are obtained and hidden
+// again).
+//
 // On the terminal `result` message we `await flushPomeTelemetry()` BEFORE
 // yielding it, so all spans are exported before the agent's loop ends and it
 // calls process.exit() — pome's CLI reads the session trace blob immediately
 // after the agent returns (the finalize-before-flush contract).
 
 import { flushPomeTelemetry, recordGenAiSpan } from "./otel.js";
+import { MessageDeltaTracker, isPartialMessageArtifact } from "./partial-messages.js";
 
 type WithType = { type?: string };
 
@@ -78,17 +87,19 @@ export async function* withGenAiSpans<T extends WithType>(
   source: AsyncIterable<T>,
 ): AsyncGenerator<T, void, unknown> {
   // Boundary marking the start of the next turn. Initialized when iteration
-  // begins; advanced after every yielded message, so a turn's span starts at
-  // the message yielded before its first content block.
+  // begins; advanced after every yielded message an agent author would see, so
+  // a turn's span starts at the message yielded before its first content block.
   let turnStartMs = Date.now();
   let pending: PendingTurn | null = null;
+  const deltas = new MessageDeltaTracker();
 
   function closeTurn(): void {
     if (!pending) return;
+    const truth = deltas.take(pending.id);
     recordGenAiSpan({
       model: pending.model,
       inputTokens: pending.inputTokens,
-      outputTokens: pending.outputTokens,
+      outputTokens: truth?.outputTokens ?? pending.outputTokens,
       startTimeMs: pending.startTimeMs,
       endTimeMs: pending.endTimeMs,
       isError: pending.isError,
@@ -98,6 +109,8 @@ export async function* withGenAiSpans<T extends WithType>(
 
   try {
     for await (const msg of source) {
+      deltas.observe(msg);
+
       if (msg.type === "assistant") {
         const a = msg as AssistantLike & WithType;
         const usage = (a.message?.usage ?? {}) as Record<string, unknown>;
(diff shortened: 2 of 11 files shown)


AFFFPupu:
Closes F-998 — https://linear.app/pome-sh/issue/F-998/span-lane-output-tokens-are-a-message-start-snapshot-5x-low

AFFFPupu:
## Follow-up: drove the sub-agent path live, and it moved two of my assumptions

Worth reading before reviewing the tracker, because the first version of my verification script **failed** here and the failure turned out to be the script, not the code.

### 1. The SDK forwards no sub-agent stream events

Measured on a real two-sub-agent run against `claude` 2.1.220:

```
stream_event total: 34; with parent_tool_use_id: 0
message_delta total:  2; subagent ones:          0
assistant total:      7; subagent ones:          2
```

Sub-agent `assistant` messages arrive; their stream events do not. So a sub-agent turn has **no `message_delta`** and keeps the `message_start` snapshot.

That makes the tracker's `parent_tool_use_id` bucketing **insurance, not a live path**. I've corrected the comment to say so rather than implying it carries weight today. Keeping it: `SDKPartialAssistantMessage` declares the field, and the day the SDK starts forwarding those events, parallel Task sub-agents would otherwise silently cross-attribute each other's deltas — `message_delta` has no id of its own, so a single global "open message" slot is exactly the bug.

### 2. `SDKResultMessage.usage` counts only the main agent

The tape settles it arithmetically — main-agent deltas `517 + 7 == 524 == result.output_tokens`, with the two sub-agent turns' `4 + 4` outside that total. Second run: `390 + 7 == 397`.

So **the ticket's exactness invariant is really scoped to main-agent turns**, and on a run that uses sub-agents both lanes total *more* than the SDK's own rollup rather than matching it. That asymmetry is pre-existing, not from this change — F-994 shows the same on the input side (spans 100505 vs result 59610 on the run above), because neither lane has ever filtered by `parent_tool_use_id`.

### What I did about it

- Corrected the tracker docstring and `take()`'s contract to name the sub-agent case as the normal null.
- Added a test pinning the behaviour that actually matters: a sub-agent turn keeps its snapshot and **the main agent's delta does not leak onto it**.
- README caveat records both facts, so the next person reading `agent_tokens_out` on a sub-agent run is not surprised.
- Rewrote the live sub-agent check around the invariant that holds. Now green:

```
PASS  subagents actually ran                                  (2 spawn rows, 6 tagged msgs)
PASS  main-agent turns sum to the result rollup               (397 vs 397)
PASS  subagent turns are the only ones without a delta        (2 subagent / 2 main)
PASS  the main agent's delta did not leak onto a subagent turn ([4,4])
PASS  no artifacts leaked                                     (none)
PASS  one span per LlmTurnEvent row                           (4 vs 4)
PASS  span and signals lanes agree                            (out 405 vs 405)
```

### Not fixed here, deliberately

Sub-agent output tokens stay low, because **no data source exists** — `forwardSubagentText` forwards messages, not stream events, and those `assistant` messages already arrive. Nothing to read. Same reason their span windows come out at 1–2 ms: the stream reveals only the completed sub-agent message, so there is no honest start time to measure from. Both are smaller versions of the bug this PR fixes and want their own ticket rather than a guess bolted onto this one.

AFFFPupu:
## The kill switch, exercised live — and it isolates the fix

Ran the same verification twice on the same build, flipping only `POME_DISABLE_PARTIAL_MESSAGES`. This answers "how do I know the `message_delta` path is what fixed it, rather than something incidental":

| same build, same script | `POME_DISABLE_PARTIAL_MESSAGES=1` | default |
| -- | -- | -- |
| per-turn output | `2 + 6 + 2 + 1` = **11** | `226 + 81 + 81 + 241` = **629** |
| vs `result` rollup | 11 vs **613** ✗ | 629 vs **629** ✓ |
| `finish_reasons` | `[null, null, null, null]` | `tool_use ×3, end_turn` |
| input tokens | 124070 == 124070 ✓ | 123551 == 123551 ✓ |
| span windows | 675 / 12624 / 2220 / 5858 ms | 824 / 1964 / 2548 / 4472 ms |
| artifacts leaked | none | none |

So with the switch on you get exactly the old behaviour and nothing else changes — input attribution, span windows, and stream fidelity are all independent of it. Nothing crashes, no warnings.

Two things worth noting from this:

- **The gap is bigger than the ticket's ~5x when a run ends on a long text turn** — 11 vs 613 is 56x. The ~5x figure came from a tool-heavy tape where every turn was short. The dashboard error was worse than advertised on exactly the runs where output tokens matter most.
- `finish_reasons` going all-null under the switch confirms the `LlmTurnEvent` lane's finish reasons are sourced purely from `message_delta`, as intended — there is no hidden second path quietly filling them in.

AFFFPupu:
## Mutation coverage: 15/15, 0 survived

I found one real hole by mutation-testing rather than reading, so I did the rest systematically. For each guard: break it, run the suite, check that something fails.

The hole, for the record: replacing `take()`'s by-id lookup with a naive "return whatever delta we saw last" **passed the entire suite**. In the normal flow only one delta is ever outstanding — a turn closes on the *next* turn's first assistant message, which arrives before that turn's own delta — so positional and by-id were indistinguishable in every test I had written. They diverge the moment two are pending: a turn that never produced an `assistant` message leaves its delta behind, and if the SDK ever forwards sub-agent stream events, two streams are in flight by construction. Two tests now pin it, taken out of order on purpose.

Full sweep on the final tree:

```
caught   span lane: ignore the delta, use the snapshot (revert the fix)          4 tests
caught   signals lane: ignore the delta's output tokens                          2 tests
caught   signals lane: ignore the delta's stop reason                            1 test
caught   filter: stop claiming status:requesting (under-filter)                  5 tests
caught   filter: claim EVERY system/status (over-filter, drops compaction)       1 test
caught   tracker: never release the open slot on message_stop                    1 test
caught   tracker: sum cumulative deltas instead of latest-wins                   2 tests
caught   tracker: single global open slot, ignore parent_tool_use_id             1 test
caught   gate: forget the signals lane, only OTLP turns injection on             3 tests
caught   gate: ignore the kill switch                                            2 tests
caught   query: inject but do not filter (leaks the whole partial stream)        2 tests
caught   query: never inject (Option B)                                          2 tests
caught   span lane: let partial messages move the turn boundary                  1 test
caught   signals lane: let partial messages move the turn boundary               1 test
caught   tool lane: stop skipping partial messages (spurious SubagentSpawnEvent) 1 test

15/15 caught; 0 survived.
```

Both directions of the filter are covered, which is the part I most wanted proof of: under-filtering (leaking `status:requesting`) and over-filtering (swallowing a real `compacting` status) each fail a test, so the narrow `requesting`-only match is pinned from both sides rather than just asserted in a comment.

Script is at `.context/mutation-check.mjs` in my workspace — not committed, since it hard-codes source strings and would rot on the next refactor.

PR #253, opened 2026-07-29
Point the CLI at the published GitHub vocabulary (F-1075)

AFFFPupu:
## What

`cli/package.json` moves to **`@pome-sh/sdk@0.7.0` + `@pome-sh/twin-github@0.4.0`** (published in [packages-v20](https://github.com/pome-sh/digital-twins/releases/tag/packages-v20)), so `pome checks github` serves **eleven** declared checks instead of one.

```
$ pome checks github
github — 11 declared checks (@pome-sh/twin-github 0.4.0)

  github.issue-exists
    Issue #<issue> exists in `<repo>`
    Asserts an issue with this number is present in the repository's final state. It says
    NOTHING about the issue's content, state, labels or assignee …
    needs: the final state
    pome checks add <file> --check github.issue-exists --arg issue=1 --arg repo=acme/api
  …
```

## Why

**This is a compatibility fix, not a convenience.** `pome checks add` compares its vocabulary digest with the cloud's before writing. A CLI still bundling `twin-github@0.3.0` refuses *every* write against a cloud carrying the new vocabulary, naming the check that moved.

That refusal is the handshake working as designed — F-1074 built it precisely so a pin skew can never write a sentence the grader won't bind. But the repair is this pin, and **it should land before [pome-cloud#421](https://github.com/pome-sh/pome-cloud/pull/421) deploys**, or every author hits the refusal in between.

## Notes

Three sentence forms stop binding, and re-rendering is the repair. Anyone with a task file carrying them will see `pome checks add` / `save_task` say so by name:

- an issue/PR check must now **name its repository** — the old patterns took `` in `owner/repo` `` as *optional* and scanned repos first-match-wins without it
- `Issue #N has label X` is gone — there is one check, `github.issue-has-label`
- `A REQUEST_CHANGES review exists …` is gone — the API state is `CHANGES_REQUESTED`

The bundled `cli/tasks/` corpus was already re-rendered in #252, so nothing shipped by the CLI is affected.

Also bundles `@pome-sh/sdk@0.7.0`, whose `defineCheck` now rejects a param pattern that opens its own capture group — a declaration bug that would otherwise hand every later slot its neighbour's argument, silently.

## Review

- `cli/package.json` + lockfile: two pins.
- `cli/.changeset/github-vocabulary-eleven-checks.md`: **minor**, per `AGENTS.md` — the vocabulary a user can author against goes from 1 to 11, and the bundled sdk carries a signature change. Version and `CHANGELOG.md` are the bot's; not touched.

## CI/CD

Verified against the **published** packages, not a local tarball: `npm test` 746 passed · `npm run typecheck` clean · `pome checks github` renders all eleven from a real build.

Merging this makes the Changesets bot open a "Version Packages" PR; merging *that* publishes `@pome-sh/cli`. Per `project_twins_release_ops`, the bot's PR needs an empty commit from a human author to trigger the required CI.


AFFFPupu:
## ⚠️ Correction to this PR's body — I had the ordering backwards

The body says this "should land before [pome-cloud#421](https://github.com/pome-sh/pome-cloud/pull/421) deploys". **That is wrong, and the opposite is true for the publish step.**

The digest handshake refuses whenever the CLI's vocabulary and the cloud's *differ* — it does not care which side is ahead. So:

| CLI | cloud (prod) | `pome checks add` |
| -- | -- | -- |
| 1 check | 1 check | ✅ today |
| **11** | **1** | ❌ **refuses — the window publishing early would create** |
| 1 | 11 | ❌ refuses (users on an old CLI, unavoidable and by design) |
| 11 | 11 | ✅ |

Publishing `@pome-sh/cli` *now* would break authoring for anyone who upgrades before the next prod tag — the mirror image of the problem this PR exists to prevent.

**Correct sequence:**

1. **Merge this PR.** Safe — it publishes nothing; it makes the Changesets bot open a "Version Packages" PR that then *waits*.
2. **Ship pome-cloud** (#421 merge → `vX.Y.Z` tag → prod).
3. **Then merge the bot's Version PR**, which publishes the CLI, so the upgrade is available the moment users start needing it.

Users on an old CLI still get a refusal between steps 2 and 3 — that is unavoidable for any version-skewed client and is exactly what F-1074 built the handshake to do: refuse by name rather than write a sentence the grader will not bind.

Per `project_twins_release_ops` trap 2/7: the bot's PR needs an empty commit from a human author before its required checks run at all — **absent checks look like passing ones**, so count them against a human PR's list before merging.

PR #256, opened 2026-07-29
Commit .mcp.json for zero-setup openapi-specs MCP install (F-1118)

GaganSD:
## What
Commits `.mcp.json` + `scripts/mcp/openapi-specs.sh` so opening Claude Code in this repo gets the `openapi-specs` MCP server with no manual setup.

## Why
`openapi-spec-mcp` is private and lives in a separate repo, so a same-repo relative path (as used inside that repo's own `.mcp.json`) cannot resolve from here. The launcher script clones/refreshes `pome-sh/openapi-spec-mcp` into a local cache directory and runs its server from there. If the clone or install fails (e.g. an unauthenticated CI runner without access to the private repo), it exits non-zero with a clear error instead of silently starting a tool-less session.

Part of F-1118 (milestone C2, Verifiable Scores) — https://linear.app/pome-sh/issue/F-1118

## Notes for reviewer
Companion PR in `pome-cloud` carries the identical launcher. The primary implementation PR is `pome-sh/openapi-spec-mcp#6`.

## Test plan
- [ ] Fresh clone of this repo, open Claude Code, confirm `openapi-specs` MCP server loads with no manual config edits
- [ ] Simulate no network/auth access to the private repo and confirm the launcher fails loudly (non-zero exit, clear stderr) rather than silently


GaganSD (commented):
Full review of the diff (.mcp.json + scripts/mcp/openapi-specs.sh), which is byte-identical to the launcher already reviewed and merged in pome-cloud #427. (GitHub blocks self-approval, so recording the review as a comment.)

What I checked:
- Injection/traversal: every expansion of REPO_URL/CACHE_DIR and derived paths is double-quoted; no eval, no unquoted word-splitting sinks. The only inputs are the invoking user's own env vars (POME_OPENAPI_MCP_REPO_URL / POME_OPENAPI_MCP_CACHE_DIR), never repo content or network data, so there is no attacker-controlled surface.
- No-credentials path: set -euo pipefail plus explicit fail() on clone failure, missing cli.mjs, and npm install failure -- non-zero exit with a clear stderr message, never a silent tool-less session. The '[ before != after ] && needs_install=1' idiom is errexit-safe (LHS of an AND-list is exempt).
- .mcp.json: no secrets, just runs the launcher relative to repo root. CI never consumes .mcp.json (only Claude Code does, behind project-MCP approval), so no CI/local behavior divergence; an unauthenticated runner would just get the loud clone failure if it ever ran the script.
- Correctness: clone -> best-effort ff-only refresh -> verify committed server/dist/cli.mjs -> npm install only when HEAD changed or node_modules missing -> exec node with --dir specs. Matches the PR description.

Nits, non-blocking: fail() echoes REPO_URL, so an HTTPS override with an embedded token would print the user's own token to their own stderr (default is a plain SSH URL); and two simultaneous first launches could race on the cache dir (worst case one fails loudly and retries).

Verdict: approve.

PR #257, opened 2026-07-29
Re-pin the CLI at the twelve-check GitHub vocabulary (F-1132)

AFFFPupu:
`cli/package.json` moves to `@pome-sh/sdk@0.8.0` + `@pome-sh/twin-github@0.5.0`, so `pome checks github` serves **twelve** declared checks instead of eleven — gaining `github.no-unsupported-endpoint`, the tape-substrate check F-1076 declared. Both versions were already on npm, so this is a pin bump plus a CLI publish, not a package release.

## Why it's urgent

prod has served the twelve-check vocabulary since `v0.4.36`. `pome checks add` compares its vocabulary digest against the cloud's before writing — so `@pome-sh/cli@0.11.0` refuses **every** `github` criterion it is asked to write, not only the new one:

```
$ pome@0.11.0 checks add <file> --check github.file-exists --arg path=src/index.ts --arg repo=acme/api
Refusing to write: this CLI and the cloud disagree about github's vocabulary,
so a sentence written here might not be graded there.
      - github.no-unsupported-endpoint — the cloud has it, this CLI does not
```

The refusal is the handshake doing its job. Cloud-first is the ratified sequence, so this window was expected — this is its closing step.

## Verified against prod, not against a build

`cli-ci` runs `use-local-pome-tarballs.mjs`, which rewrites these pins to workspace tarballs. **CI is green on twelve checks either way and structurally cannot catch this skew** — so the verification below installed `cli/` deps from the real registry:

| | digest | checks |
|---|---|---|
| prod `api.pome.sh/v1/checks?twin=github` | `sha256:ae84a6…4694` | 12 |
| published `@pome-sh/cli@0.11.0` | `sha256:ff19d6…99e1` | 11 |
| **this branch** | `sha256:ae84a6…4694` | **12 — match** |

The same `checks add` command quoted above writes cleanly on this branch (exit 0), and `--check github.no-unsupported-endpoint` renders `- [code] No unsupported endpoint was called`.

## Not a breaking change for authored tasks

Diffed all eleven pre-existing checks between `twin-github@0.4.0` and `0.5.0`: zero drift in `id`, `template`, `params`, or `substrate`. Purely additive, so tasks written against `0.11.0` re-render unchanged. The digest moves only because a twelfth row joins the hash.

## Reviewer notes

- **The sdk pin is load-bearing, not tidiness.** The new declaration reads `CheckSubstrate.tape`, which does not exist before sdk `0.8.0` — `twin-github@0.5.0` resolving `sdk@0.7.0` would install a twin whose own check cannot compile.
- **One incidental line:** `cli/package-lock.json`'s own `version` field was left at `0.10.0` by the release bot (it bumps the manifest, not the lockfile) and catches up to `0.11.0` in the same `npm install`.
- `cli/package.json` `version` is untouched — the Changesets bot owns it.

## Checks run locally against the published pins

CI can't do this run, so it was done by hand:

- 746 tests / 86 files green; `typecheck` clean
- `gate:no-native`, `gate:no-eval` pass
- `npm publish --dry-run` + `npm pack`: the shipping tarball bundles `twin-github@0.5.0` and `sdk@0.8.0`
- `changeset status` → `@pome-sh/cli` at **minor** (matching the 0.11.0 precedent for the analogous 0.3→0.4 pin bump)

## Follow-up worth a ticket (not filed)

Nothing tracks the structural gap this exposed: **CLI CI tests the workspace, never the published pin.** That is why the skew reached users rather than a CI lane. F-940 (*Lane D: tsup bundling + single release pipeline*) is adjacent and may subsume it.


AFFFPupu:
Closes F-1132

PR #268, opened 2026-07-30
Linear declares its vocabulary, and a rendered sentence names its subject (F-1129)

AFFFPupu:
## What

The Linear twin declares an assertable check vocabulary — **eight checks**, a state model, fixture worlds, and a contract suite — and tasks 24, 25 and 26 are rewritten so every criterion names its own subject. `pome checks linear` answers with a vocabulary instead of "not migrated yet".

## Why

F-1129, milestone A3. Linear resolved **0 of 9** shipped `[code]` criteria — the joint-largest gap in the corpus. Its five cloud-side regexes keyed on `identifier` ("ENG-123") or issue number, and **not one** of the nine criteria uses either; every one names a title. The registries and the tasks were built for each other and had never once met.

Design of record: `pome-cloud docs/grading/linear-vocabulary.md`.

## Notes for reviewer

**Three findings changed the shape the ticket proposed. All measured, not argued.**

**1 · The anaphora count is six, not four — and the repair is nine rewrites.**
The ticket counted the four `that issue` pronouns; it missed `Issue has label "Needs triage"` (no subject at all) and `A reply comment exists with parentId equal to the seeded root comment` (points at an id in the seed). And all nine change, not six, because Linear inherits twin-github's scope rule rather than twin-slack's exemption: `packages/twin-linear/src/seed.ts:319-325` validates issue-title uniqueness via `issueTitlesByTeam` — **per team**. A title-keyed selector over a two-team world is exactly the ambiguity that rule closes, so slack's "there is nothing to disambiguate" argument is not available to borrow. Every state check names its team.

**2 · A selector miss FAILS, and task 26 is why.**
twin-github fails on a miss, twin-slack skips, and A2b never said which a third twin inherits. Slack's argument is about *negatives*; all seven Linear state checks are positive, where a skip is the unsafe direction. Measured on task 26 with slack's discipline:

```
[code:github] Issue #1 ... is in state open   negative, seed satisfies  -> PASSED
[code:linear] x3, issue never created          -> skipped, leaves denominator
                                                  1/1 = 100%, threshold 100 -> PASS
```

That reproduces the `KNOWN_NULL_AGENT_TASKS` entry this milestone exists to clear. With FAIL it is 25%.

**3 · The honest named skip is truncation, not renaming — and Linear is the only twin that can evidence it.**
`src/state.ts:24` exports `exportBounds.truncatedCollections`, so the twin itself reports when rows were dropped past `STATE_EXPORT_CAP = 2000`. Neither `GitHubCheckState` nor `SlackCheckState` carries an analogue. Four-outcome resolution in `check-state.ts`: collection absent → `state_incomplete` (skip); truncated + miss → `state_truncated` (skip); miss in a complete collection → **fail**; >1 match → **fail** naming the count.

**One capability is narrowed, recorded rather than absorbed.** `linear.issue-lifecycle` (`is open/closed/done`) is not migrated: a declared `... is {lifecycle}` near-misses every `is in state "..."` sentence and would report a corrupted state criterion under the lifecycle check — the exact hazard `github.issue-state`'s wording comment says its phrasing exists to avoid. Zero corpus users; `is in state "Done"` / `"Canceled"` re-expresses what matters.

**Task 26 loses a criterion rather than gaining a subject.** `linear.issue-state` fails on a missing issue, so it subsumes `An issue titled "..." exists` — F-1075's collapse, applied on the side the ticket asked us to look for it. Corpus goes 70 → 69 `[code]` criteria.

**Evidence, all run:**

- All 8 checks `discriminate` through their **assertion**, not their selector. The threaded-reply failing world reads `0 replies to a seeded comment (1 seeded comment(s), 2 at finish)` — the agent commented, just not in the thread.
- `HONEST_NULL_WORLDS` ships **empty**. `HONEST_NULL_MUTANTS` holds two, `TEAM_FREE_CHECKS` one (and its own rule forces `substrate: "tape"`).
- All 8 rewritten criteria bind **exactly one** check, verified offline against the built declarations — D6's invariant before CI sees it.
- State-shape parity arm boots a real twin over the default seed and passed first run, including that an exported issue carries `stateId` and **not** a state name.
- Full repo: 8 packages green. twin-linear 16 files / 95 tests. CLI 88 files / 782 tests. Gates green: CLI pin parity, workspace pin parity, first-party twin registration, copy-markers, legacy-markers, no-cloud-imports.

**Two traps worth knowing.** `packages/twin-linear/package.json`'s `test` script names every file explicitly (twin-slack runs bare `vitest run`), so both new test files are added to that list — a contract suite that silently never runs is the worst outcome here. And the state-shape parity arm lives in `checks-contract.test.ts` rather than a `fidelity-contract.test.ts`, because twin-linear has no such vitest suite; that is the honest reading of "reuse the harness, do not add a second drift gate".

## Review required

**Yes — twin fidelity contract.** `src/check-state.ts` is a new hand-written reader of `exportLinearState()`'s raw rows, and the state-shape parity arm is what holds it to the real export.

**Release ordering.** `@pome-sh/twin-linear` 0.1.1 → **0.2.0** (new `./checks` subpath and state model); `cli/package.json` pin moves in this PR per `PACKAGE_RELEASE.md` step 1. A `packages-v*` release must land before the pome-cloud PR, and the CLI publish must come **after** the prod tag — F-1075 proved publishing the CLI first gives upgraders an 8-check client against a 1-check prod, and the digest handshake refuses either direction of skew.

### DIFF
--- a/cli/.changeset/linear-declares-its-vocabulary.md
+++ b/cli/.changeset/linear-declares-its-vocabulary.md
@@ -1,0 +1,17 @@
+---
+"@pome-sh/cli": minor
+---
+
+`pome checks linear` answers with a vocabulary instead of "not migrated yet" —
+eight declared checks covering issue state, labels, estimate, assignee,
+comments, threaded replies, existence, and unsupported endpoint calls.
+
+Tasks 24, 25 and 26 are rewritten so every criterion names its own subject. A
+rendered sentence cannot say "that issue": under a picked check the author fills
+parameters, and a check only ever sees its own arguments. Each Linear check now
+names both the issue title and its team, because Linear validates title
+uniqueness per team rather than per workspace.
+
+Task 26 loses one criterion rather than gaining a subject: `linear.issue-state`
+fails when the issue is absent, so it already subsumes `An issue titled "..."
+exists`.
--- a/cli/package.json
+++ b/cli/package.json
@@ -69,7 +69,7 @@
     "@pome-sh/shared-types": "0.13.0",
     "@pome-sh/twin-gmail": "0.3.0",
     "@pome-sh/twin-github": "0.7.0",
-    "@pome-sh/twin-linear": "0.1.1",
+    "@pome-sh/twin-linear": "0.2.0",
     "@pome-sh/twin-slack": "0.3.0",
     "@pome-sh/twin-stripe": "0.4.0",
     "ai": "^7.0.18",
--- a/cli/src/cli/checks.ts
+++ b/cli/src/cli/checks.ts
@@ -45,7 +45,11 @@ const REGISTRIES: Record<string, readonly DeclaredCheck[]> = {
 // acceptance criterion. The day a sixth twin mounts, it repopulates itself and
 // the "not migrated yet" path comes back live without anyone remembering to add
 // a literal — which is the failure this project is named for, one level down.
-const TWINS_WITHOUT_CHECKS = MOUNTED_TWINS.filter((twin) => !(twin in REGISTRIES));
+// Annotated `string[]` on purpose: `MOUNTED_TWINS` is a literal-union tuple, so
+// the filtered result keeps that union and `isKnownTwin` cannot ask it about an
+// arbitrary string. Widening here is what keeps the caller's question — "is this
+// user-typed word a twin?" — expressible.
+const TWINS_WITHOUT_CHECKS: string[] = MOUNTED_TWINS.filter((twin) => !(twin in REGISTRIES));
 
 export function twinsWithChecks(): string[] {
   return Object.keys(REGISTRIES).sort();
--- a/cli/src/cli/checks.ts
+++ b/cli/src/cli/checks.ts
@@ -12,8 +12,10 @@ import { readFileSync } from "node:fs";
 import { join } from "node:path";
 
 import { checksDigest, templateSlots, type CheckDefinition } from "@pome-sh/sdk/checks";
+import { MOUNTED_TWINS } from "@pome-sh/shared-types";
 import { GITHUB_CHECKS } from "@pome-sh/twin-github/checks";
 import { GMAIL_CHECKS } from "@pome-sh/twin-gmail/checks";
+import { LINEAR_CHECKS } from "@pome-sh/twin-linear/checks";
 import { SLACK_CHECKS } from "@pome-sh/twin-slack/checks";
 import { STRIPE_CHECKS } from "@pome-sh/twin-stripe/checks";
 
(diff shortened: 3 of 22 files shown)


AFFFPupu:
### Rebased twice, and one decision made beyond the ticket

**#266 (stripe) and #267 (gmail) both merged while this was open.** Both touched `cli/package.json` and `cli/src/cli/checks.ts` — the same two lines each of the four A3 vocabulary tickets has to edit. Rebased onto each; full re-run after both (packages 8/8 green, CLI 88 files / 782 tests). Two small fallout commits fix root-lock drift those merges left (`twin-stripe` 0.3.1→0.4.0, `twin-gmail` 0.2.1→0.3.0).

**F-1128 left a tripwire pointed at this ticket**, in `cli/test/unit/_noVocabularyTwin.ts`: once `TWINS_WITHOUT_CHECKS` went empty it threw with

> *"Decide deliberately: keep the path with a synthetic twin id, or delete it and these tests."*

I took neither option. **`MOUNTED_TWINS` in shared-types already is the canonical twin list**, and the CLI already imports it in `session.ts` and `register.ts` — so the constant becomes a set difference rather than an emptied literal:

```ts
const TWINS_WITHOUT_CHECKS: string[] = MOUNTED_TWINS.filter((twin) => !(twin in REGISTRIES));
```

Empty today, which is A3's acceptance criterion expressed as a computation instead of a literal, and it repopulates by itself the day a sixth twin mounts.

The **path** stays: `bindCriterion` and `runChecksLintCommand` reach it for any twin id this CLI holds no declaration for, and a pinned CLI can always lag the twin set — that is the premise of the digest handshake. Three of the four tests keep asserting "unanswerable, not a pass" against a synthetic id.

The fourth could not survive, and that is the milestone succeeding rather than a test rotting: `pome checks <twin>`'s "no declared checks yet" branch needs an id `isKnownTwin` accepts *and* `checksFor` returns nothing for, and once every mounted twin declares, no id satisfies both. It is replaced by A3's completion invariant — every `MOUNTED_TWINS` entry declares ≥1 check — which goes red the day someone mounts a twin without a vocabulary, exactly when the dormant branch becomes reachable again.

**Worth knowing:** the derived list needs an explicit `string[]` annotation, because `MOUNTED_TWINS` is a literal-union tuple and `isKnownTwin` must ask it about an arbitrary user-typed word. Missing it failed the **build**, which surfaced only as `bin-exec-bit.test.ts` — the build's `chmod` step never ran, so a test about file permissions was the only thing reporting a type error.

`pome checks` now: **github 13 · gmail 7 · linear 8 · slack 5 · stripe 11**. `TWINS_WITHOUT_CHECKS` is empty.

PR #278, opened 2026-08-02
AgentSession uses field names Linear does not have (F-1172)

GaganSD:
## What

`twin-linear`'s `AgentSession` **output type** now uses Linear's real field names and types; a pre-rename database migrates in place on open; and a new guard test fails whenever the twin declares an `AgentSession` field Linear does not have.

| was | is now |
| --- | --- |
| `state: String!` | `status: AgentSessionStatus!` |
| `externalUrl: String` | `externalUrls: JSON!` |
| `agentUser: User!` | `appUser: User!` |
| `id: String!` | `id: ID!` |
| `createdAt` / `updatedAt: String!` | `createdAt` / `updatedAt: DateTime!` |
| `plan: String` | `plan: JSON` |

`AgentSessionStatus` is a real enum carrying Linear's six members — `pending`, `active`, `awaitingInput`, `complete`, `error`, `stale`. The twin's invented `completed` / `failed` / `canceled` are rejected now and rewritten on open: `completed` → `complete`, `failed` → `error`, `canceled` → `stale` (Linear has no cancellation state; `stale`, "no longer progressing", is its closest neighbour). `externalUrls` is a collection of `{ url, label }` objects, matching Linear, and is never null. Mutation inputs (`appUserId`, `status`, `externalUrls: [AgentSessionExternalUrlInput!]`), the `/state` export and the `AgentSessionEvent` webhook payload follow the rename.

## Why

F-1172. An agent written against the real Linear API read `undefined` from the twin, and an agent written against the twin broke in production — the exact failure the twins exist to prevent. pome-cloud's declared lane found two of these (`state`, `externalUrl`) only because the capture query happened to select them; reading the SDL showed six.

Per the founder ruling on the ticket: fix the twin, do not register a divergence, and ship the rename straight as a breaking change. No alias and no deprecation window — a twin carrying both `state` and `status` would still expose a field Linear does not declare, so the shim would leave the gate red on the alias.

### Scope of the "no invented fields" claim — read this first

The claim in this PR is scoped to the **output type**. `GUARDED_TYPES` covers `AgentSession`, `AgentSessionStatus` and `AgentSessionExternalUrlInput` — the output surface plus the one input object the twin mirrors exactly.

**The mutation-input surface still diverges from Linear and this PR does not fix it.** Verified against `api.linear.app`:

- `AgentSessionUpdateInput` upstream declares exactly `externalLink, externalUrls, addedExternalUrls, removedExternalUrls, plan, dismissedAt, userState` — **no `status`, no `id`**. Upstream, session status moves via agent activities, not through `agentSessionUpdate` at all. The twin declares `id, status, plan, externalUrls`, so renaming `state` → `status` on that input replaced one invented field with another.
- `AgentSessionCreateOnIssue` / `AgentSessionCreateOnComment` are not Linear type names at all (upstream has `AgentSessionCreateInput { id, issueId, appUserId, context }`), and the twin adds `plan`.
- `AgentActivityCreateInput` — twin: `sessionId, type, body, ephemeral`; Linear: `id, agentSessionId, signal, signalMetadata, contextualMetadata, content, ephemeral`.

All of these predate F-1172 (the old `state` / `agentUserId` were equally invented) and are tracked separately. `status` stays on the input rather than reverting to `state`: both are invented, but `status` is at least consistent with the output type. The scope note is recorded at `GUARDED_TYPES` in `scripts/regen-linear-introspection.mjs`, on `MUST_BE_GUARDED` in the guard test, and in the CHANGELOG, so nobody reads the guard as evidence about the input surface.

### The guard

`test/linear-schema-subset.test.ts` asserts the twin's guarded members are a *subset* of Linear's, read from `fixtures/linear-introspection.json` — a committed, credential-free slice of Linear's real introspection response (Linear answers introspection unauthenticated). Refresh with `node scripts/regen-linear-introspection.mjs`. Subset, not equality: not modelling a field is ordinary coverage scope, inventing one is a defect. A type absent from the fixture is unguarded, so the test also asserts the types it claims to guard are present and non-empty — an empty fixture cannot make it pass vacuously.

**Proven it can fail.** Committed red first (`agentUser, externalUrl, state`), green after the rename, and separately verified green → red → green by adding and removing an invented `totallyRealLinearField`.

### The storage migration

An existing `LINEAR_TWIN_DB` file used to open **clean** on the old columns — `CREATE TABLE IF NOT EXISTS` is a no-op on an existing table — and then die later, far from the cause, with `"undefined" is not valid JSON` out of `mapAgentSession`. `migrate()` now renames `agent_user_id` → `app_user_id` and `state` → `status`, adds `external_urls_json`, backfills a non-empty `external_url` into a one-entry collection (`[{ url, label: "" }]` — Linear's label is non-null and the old shape carried none), drops `external_url`, and rewrites the three retired status values. Idempotent; a no-op on a database created by the current schema. Cloud's per-session databases are ephemeral (ADR-012) and unaffected; local and self-host files are not.

Verified by building a database with the pre-rename code and reopening it with this branch — before/after in the review thread. `test/agent-session-migration.test.ts` is the committed regression (5 cases: column rename, all three status mappings including `canceled`, the `external_url` backfill, idempotency, and the read-boundary failure); all 5 go red when the `migrateAgentSessions(db)` call is removed.

`mapAgentSession` no longer casts a stored status through unvalidated. An unmappable value now fails at the read boundary naming the value and the cause, instead of surviving to die at GraphQL enum serialisation with `Enum "AgentSessionStatus" cannot represent value: "canceled"`.

### Other

- **#277's tri-state contract survives.** `updateAgentSession` still tests presence with `!== undefined`, never `in`. `externalUrls` joins the contract: absent or `undefined` leaves it alone, `null` clears it to `[]`. `test/partial-update-tristate.test.ts` asserts that over `externalUrls` as well as `plan`.
- **Two findings beyond the ticket's table**, confirmed against introspection: `plan: String` is also a type Linear does not use (`plan: JSON`) — fixed here, same defect class, and a string is already valid JSON so nothing changes on the wire. And `externalUrls: JSON!` is itself deprecated upstream in favour of `externalLinks: [AgentSessionExternalLink!]!` (as is `externalLink: String`, the exact 1:1 match for the old `externalUrl`). Both are declared, so the gate is satisfied either way; adopting `externalLinks` is new-field scope and out of scope here.
- **No version bump.** `package.json` stays 0.2.0 and the entry sits under `## Unreleased`, matching how #277 landed; per `PACKAGE_RELEASE.md` the release batch owns the bump and the `cli/` pin. This is a `/state`-payload and on-disk-schema change, so it is a **minor** (0.3.0) when that batch runs. pome-cloud #479 needs that release plus a re-pin.

## Review required

Yes — twin fidelity contract, and a breaking change to observable twin behaviour (`AgentSession` field names, `AgentSessionStatus` values, on-disk `agent_sessions` schema, `/state` export, `AgentSessionEvent` webhook payload).

### DIFF
--- a/packages/twin-linear/CHANGELOG.md
+++ b/packages/twin-linear/CHANGELOG.md
@@ -56,7 +56,11 @@
   (`fixtures/linear-introspection.json`, refreshed by
   `node scripts/regen-linear-introspection.mjs`). The old drift only surfaced
   when a capture query happened to select the field; this fails at authoring
-  time instead.
+  time instead. Scope: the guard covers the output type and the one input
+  object the twin mirrors exactly. The mutation-input surface
+  (`AgentSessionUpdateInput`, `AgentSessionCreateOnIssue` / `OnComment`,
+  `AgentActivityCreateInput`) still diverges from Linear, predates this change,
+  and is tracked separately.
 - **Fixed: partial updates no longer wipe fields the caller never mentioned
   (F-1166).** Nullable fields on update mutations are tri-state — key absent
   and key present with `undefined` both mean "leave alone", `null` means
--- a/packages/twin-linear/CHANGELOG.md
+++ b/packages/twin-linear/CHANGELOG.md
@@ -2,6 +2,48 @@
 
 ## Unreleased
 
+- **BREAKING — `AgentSession` now uses Linear's real field names and types
+  (F-1172).** The twin declared four fields Linear does not have, so an agent
+  written against real Linear read `undefined` from the twin and an agent
+  written against the twin broke in production. There is no alias and no
+  deprecation window: a twin carrying both the old and the new name would still
+  expose a field Linear does not declare, which is the defect itself.
+
+  | was | is now |
+  | --- | --- |
+  | `state: String!` | `status: AgentSessionStatus!` |
+  | `externalUrl: String` | `externalUrls: JSON!` |
+  | `agentUser: User!` | `appUser: User!` |
+  | `id: String!` | `id: ID!` |
+  | `createdAt` / `updatedAt: String!` | `createdAt` / `updatedAt: DateTime!` |
+  | `plan: String` | `plan: JSON` |
+
+  `AgentSessionStatus` is now a real enum carrying Linear's six members —
+  `pending`, `active`, `awaitingInput`, `complete`, `error`, `stale`. The twin
+  previously accepted `completed`, `failed` and `canceled`, which Linear does
+  not have; those are rejected now. Migrate `completed` → `complete` and
+  `failed` → `error`.
+
+  `externalUrls` is a collection, matching Linear: it replaces the single
+  nullable `externalUrl` string with a JSON array of `{ url, label }` objects,
+  and it is never null (an empty array when there are none). The mutation
+  inputs follow: `agentSessionCreateOnIssue` / `agentSessionCreateOnComment`
+  take `appUserId` and `externalUrls: [AgentSessionExternalUrlInput!]`, and
+  `agentSessionUpdate` takes `status` and `externalUrls`. F-1166's tri-state
+  contract is unchanged — key absent or `undefined` leaves a field alone, and
+  `null` clears it (`externalUrls: null` clears to `[]`).
+
+  The `/state` export and the `AgentSessionEvent` webhook payload rename with
+  the wire surface (`appUserId`, `status`, `externalUrls`).
+- **Added: a guard that fails when the twin invents a field Linear does not
+  have (F-1172).** `test/linear-schema-subset.test.ts` asserts the twin's
+  `AgentSession`, `AgentSessionStatus` and `AgentSessionExternalUrlInput`
+  members are a subset of Linear's, read from a committed, credential-free
+  slice of Linear's real introspection response
+  (`fixtures/linear-introspection.json`, refreshed by
+  `node scripts/regen-linear-introspection.mjs`). The old drift only surfaced
+  when a capture query happened to select the field; this fails at authoring
+  time instead.
 - **Fixed: partial updates no longer wipe fields the caller never mentioned
   (F-1166).** Nullable fields on update mutations are tri-state — key absent
   and key present with `undefined` both mean "leave alone", `null` means
--- a/packages/twin-linear/CHANGELOG.md
+++ b/packages/twin-linear/CHANGELOG.md
@@ -21,8 +21,21 @@
   `AgentSessionStatus` is now a real enum carrying Linear's six members —
   `pending`, `active`, `awaitingInput`, `complete`, `error`, `stale`. The twin
   previously accepted `completed`, `failed` and `canceled`, which Linear does
-  not have; those are rejected now. Migrate `completed` → `complete` and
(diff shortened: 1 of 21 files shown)


GaganSD:
### Storage migration — verified against a real pre-rename database

Not synthetic. The database was built by checking out the pre-rename commit (`5ea50b1`) into a worktree and running the **old** code against a file DB, then reopening that file with this branch.

**Built with the pre-rename code (`5ea50b1`):**

```
[old code] wrote /tmp/f1172-old.db
[old code] columns: id,issue_id,comment_id,agent_user_id,state,plan,external_url,created_at,updated_at
[old code] rows: [
 { "id": "45b5a038-…", "agent_user_id": "user_agent", "state": "canceled",
   "plan": null, "external_url": "https://old.example/run" },
 { "id": "d535e10d-…", "agent_user_id": "user_agent", "state": "completed",
   "plan": "OLD-PLAN", "external_url": null },
 { "id": "c7db9891-…", "agent_user_id": "user_agent", "state": "failed",
   "plan": null, "external_url": null }
]
```

**BEFORE — that file opened with `f8f64cc` (rename, no migration):**

```
open+migrate: OK "opened"
columns: id,issue_id,comment_id,agent_user_id,state,plan,external_url,created_at,updated_at
domain list: THREW "undefined" is not valid JSON
domain update: THREW "undefined" is not valid JSON
state export: THREW no such column: app_user_id
```

**AFTER — the same file opened with `7cff95b`:**

```
open+migrate: OK "opened"
columns: id,issue_id,comment_id,app_user_id,status,plan,created_at,updated_at,external_urls_json
domain list: OK [
  {"id":"45b5a038-…","appUserId":"user_agent","status":"stale","plan":null,
   "externalUrls":[{"url":"https://old.example/run","label":""}]},
  {"id":"d535e10d-…","appUserId":"user_agent","status":"complete","plan":"OLD-PLAN",
   "externalUrls":[]},
  {"id":"c7db9891-…","appUserId":"user_agent","status":"error","plan":null,
   "externalUrls":[]}
]
domain update: OK {"id":"45b5a038-…","status":"active",
   "externalUrls":[{"url":"https://old.example/run","label":""}]}
state export: OK [ …three rows, appUserId / status / externalUrls… ]
```

All three retired statuses map, `canceled` included; the old `external_url` survives into the collection; `/state` works.

### The regression test can fail

`test/agent-session-migration.test.ts`, with the `migrateAgentSessions(db)` call commented out of `migrate()`:

```
 ❯ test/agent-session-migration.test.ts (5 tests | 5 failed) 80ms
     × renames the columns instead of opening clean and crashing later 28ms
     × maps every retired status onto a Linear member, canceled included 14ms
     × carries the old single external_url into the externalUrls collection 13ms
     × is idempotent — reopening a migrated database changes nothing 13ms
     × fails legibly if an unmigratable status somehow reaches the read boundary 12ms
```

Restored:

```
 Test Files  1 passed (1)
      Tests  5 passed (5)
```

The test builds the pre-rename table by putting a seeded database back into the 0.2.0 shape and asserting its column set equals the frozen pre-rename list *before* touching the migration, so it cannot silently drift into testing the new shape against itself.

PR #299, opened 2026-08-04
pome-run-task should pass agent_model on finalize_run

GaganSD:
## What
`pome-run-task` skill, README, and launch refs now tell coaches to pass `agent_model` (the examinee model from intake / launcher config) on `finalize_run`.

## Why
Coaches were finalizing without naming the examinee model, so the report Model column stayed blank when telemetry did not fill it. Companion MCP/docs guidance lands in `pome-cloud`.

## Notes for reviewer
No hardcoded model slug — value is whatever the examinee actually ran on.


GaganSD:
done in a different pr

PR #304, opened 2026-08-04
The coach declares which version it is running (F-899)

AFFFPupu:
Executive summary: The fix loop tells the coach to open a fresh `group_id` for a post-fix rerun, but never to change the agent's declared version — so the failing run and the fixed run reach the platform under the same identity. Run-sets are partitioned by declared version, which means a rerun as the same version is the platform being told the failure and the fix are one agent, and the gap between them gets reported as that agent's reliability spread. This teaches the coach to pass `agent_version` on every run and to bump it on a post-fix rerun. Docs-only — one SKILL.md.

## What

`skills/pome-run-task/SKILL.md`:

- §1 — `run_task` / `run_trials` take `agent_version`, read from the manifest's `agent.version` (`pome.json`). If the builder declares none, ask for one rather than sending nothing. Never auto-bump: a version that moves on its own splits one exam into run-sets of one.
- §5.1 — the post-fix rerun bumps it, alongside the fresh `group_id` + `baseline_group_id` that were already there, with the reason spelled out so the coach can explain it to the builder.

## Why

A fresh `group_id` alone is not enough. The task reliability page also folds an *implicit* run-set — trials with no group — and that fold partitions by declared version, and the verdict strip asserts "same agent, same prompt" over whatever it folded. Rerun a fix under the unchanged version and the honest reading of those trials is a comparison, but what gets printed is a spread.

The tools have accepted `agent_version` since it was added to the session contract; nothing in the coach path was passing it.

## Notes for reviewer

The platform-side half (persisting the version at mint, partitioning the fold, and the display that makes it visible) is a separate change in the cloud repo. **Land this one first**: with the old skill against the new platform, coaches keep producing unversioned runs and nothing improves.

## Tests / CI

Prose-only change to one skill file — no code paths touched, so no local test command applies. Relying on repo CI.

### DIFF
--- a/examples/support-triage/README.md
+++ b/examples/support-triage/README.md
@@ -42,22 +42,23 @@ open issue #1 for the coupon bug, then `#support` receives a *new* report of the
 
 ## Run it against Pome
 
-The two versions are exercised as **two distinct agents** end-to-end, so the
-dashboard shows two separate identities with separate scores.
+The two versions are **one agent at two declared versions**, so the dashboard
+can show the v1→v2 delta rather than two unrelated identities with no
+relationship between them.
 
 1. **Register on Cloud Managed Agents** — create each agent from its YAML on
    Anthropic's Managed Agents platform (`ant beta:agents create`, model
    `claude-sonnet-5`).
-2. **Register on Pome** (control MCP) — one call per version, so each accrues its
-   own history:
+2. **Register on Pome** (control MCP) — once. The version is not part of the
+   identity; it is declared per run in step 3:
    ```
-   register_agent(name="support-triage-v1", twins=["github","slack"])
-   register_agent(name="support-triage-v2", twins=["github","slack"])
+   register_agent(name="support-triage", twins=["github","slack"])
    ```
-3. **Run** — for each agent id, `run_trials(task_id, agent_id, n=5)`. Each
-   trial returns an `examinee_launch` spec (per-session twin MCP URLs, a bearer,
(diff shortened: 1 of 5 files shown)


AFFFPupu:
> **P2 — Unversioned README examples remain**

Fixed in 9127f7b, and it went further than the call signatures.

`examples/support-triage/README.md` was not just missing `agent_version` — it registered the two prompts as **two separate Pome agents** (`support-triage-v1`, `support-triage-v2`). That is the shape declaring a version per run exists to replace: two unrelated identities, no delta between them. It is now one `register_agent(name="support-triage")`, `agent_version` on each run, and the v1 run's `group_id` passed as the v2 run's `baseline_group_id`.

Also updated:

- `skills/pome-run-task/README.md` — the run-loop signatures it repeats from the SKILL, plus the fix-loop step, so a reader does not get one shape from the SKILL and another from the README beside it.
- `examples/support-triage/local/README.md` — `run_task(..., agent_version="v1")`, with the reason the re-run declares `"v2"`.
- `examples/support-triage/VERIFICATION.md` — Reproduce steps updated. The recorded run ids above them are **not** touched: those runs really were made as two agents, so the file says so in a note rather than rewriting the record to match current advice.

Deliberately not changed: the CLI-era mapping table in `skills/pome/SKILL.md` (`pome run -n 3` → `run_trials(n, task_id)`). That table is abbreviated by design — it omits `agent_id` too — so adding one argument there would be inconsistent without turning it into a signature reference, which is what `pome-run-task` is for.

PR #316, opened 2026-08-05
`pome install` / `pome doctor` print the install-all form — and this one publishes (F-1226)

AFFFPupu:
Stacked on #315. **Read the release note before merging — merging this publishes `@pome-sh/cli`.**

## The change

`pome install` and `pome doctor` printed `npx skills add pome-sh/digital-twins`, which opens the `skills` picker with six skills and nothing ticked. They now print `--skill '*'`.

```
  2. npx skills add pome-sh/digital-twins --skill '*'
                                           # install the Gen-2 coach skills
                                           # (--skill '*' takes all six; the
                                           #  picker opens with none ticked)
```

`--all` was the other candidate and is wrong here: it is shorthand for `--skill '*' --agent '*' -y`, so it installs into **every** agent on the machine. `--skill '*'` keeps the agent prompt.

## Why this is a separate PR, and what merging it does

`scripts/ci/check-version-bump-required.mjs` (added by #312) makes a version bump **mandatory** for any `cli/` path change, and `release.yml` publishes on the version diff. So for this repo, merging a CLI change now *is* publishing one — there is no unbumped option.

`@pome-sh/cli` on npm is **0.20.0**, which is also the version in the tree. Everything merged since that publish is unpublished, so this bump ships all of it:

| Merged since 0.20.0 | |
| --- | --- |
| #237 | workspace unification + internal packages privatized |
| #239 | tsup bundling; two release systems replaced with version-diff-on-merge |
| #311 | one typed twin registry replaces four parallel lists |
| #313 | `shared-types` dissolved into `@pome-sh/wire`; CONTRACT 1.6.0 |
| #312 | build/lint hygiene, tarball audit gate |
| this PR | the two lines of copy above |

Hence **0.21.0**, not a patch — the payload is a bundler swap and a package dissolution, not a copy fix. **The version number is your call**; I picked minor because 0.x minor plays major here and #313 moves a contract.

The manifest half of F-1226 (#315) is deliberately separate: it touches no `cli/` path, so it lands the actual install-all row with no publish attached.

## Verified

- `cli` suite: 1014 tests / 106 files pass. `install-command.test.ts` now pins `--skill '*'` rather than the bare substring, so the flag cannot silently fall out.
- `node dist/src/cli/main.js install` output checked by hand (quoted above).
- `check-version-bump-required.mjs` passes against the base branch.


AFFFPupu:
Rebased onto main and addressed the review, but **this PR is now blocked on something outside it** — flagging rather than merging.

**1. Greptile's lockfile point was right, and it is main's problem too.** Fixed here via `npm install --package-lock-only` (one line). Worth knowing: main currently carries `cli/package.json` at 0.21.0 against a `package-lock.json` still recording `cli` at 0.20.0 — #317 introduced that drift, and this PR's regeneration incidentally corrects it.

**2. Version is now 0.21.1, not 0.21.0.** #317 claimed 0.21.0 on main while this PR was open, so `check-version-bump-required.mjs` needs a further bump for any `cli/` change. The "this publishes a big batch" note in the original description is **obsolete** — #317 is what carries #237/#239/#311/#312/#313 to npm. This PR is back to being what it looks like: two lines of copy.

**3. The blocker: main's `release` run for 0.21.0 failed, so npm is still on 0.20.0.**

[Run 31016133589](https://github.com/pome-sh/digital-twins/actions/runs/31016133589) — `publish @pome-sh/cli`:

```
npm error 404 The requested resource '@pome-sh/cli@0.21.0' could not be found
             or you do not have permission to access it.
```

A 404 on publish is npm's "no permission" for a scoped package. The sibling `publish @pome-sh/adapter-claude-sdk` job reports success, but it is a no-op — the adapter is 0.3.1 in both the tree and npm, so it took the `version unchanged — skipping` branch. It proves nothing about OIDC.

Leading hypothesis: **`@pome-sh/cli`'s npm Trusted Publisher still names the deleted workflow.** Trusted Publishing matches on repository *and workflow filename*; #239 deleted `cli-release.yml` and the publish now runs from `release.yml`, so the OIDC token's `job_workflow_ref` no longer matches what npm has on file, and npm answers 404. Same failure class as the `pome-twins` → `digital-twins` rename that broke packages-v10..v13.

Check: npmjs.com → `@pome-sh/cli` → Settings → Trusted Publisher → workflow filename should read `release.yml`. `@pome-sh/adapter-claude-sdk` needs the same audit before its next bump, since its no-op success has never exercised the new path.

Until that is fixed, merging this republishes nothing — it would just fail the same way one version later. **#315 is unaffected and mergeable now** (it touches no `cli/` path and publishes nothing).

PR #318, opened 2026-08-05
Close the examinee sandbox, and measure what that actually fixes

AFFFPupu:
Executive summary: The hero example's failing baseline is not failing — measured 4/5 green on 2026-08-04 (#309). This PR runs the two follow-up measurements that were needed before any re-cut, closes the sandbox that let one trial read its own grading criteria, and corrects the claims the runs refuted. It deliberately does **not** re-cut the baseline defect: that is a design decision, and the measurements here are what it should be decided on. `DENY_ISSUE_LOOKUP` still ships `true`, now documented as a known-green placeholder rather than a lesson.

## What

- **`options.tools: []`** in `examples/support-triage/local/src/index.ts` — the examinee now exposes **no** SDK built-in to the model. This is an allowlist, not another name in `disallowedTools`: `tools` replaces the base set, so empty is complete by construction. MCP tools are untouched (they come from `mcpServers`), so the twin surface the exam is about stays whole.
- `test/tool-policy.test.ts` pins the closed sandbox as its own property, separate from the lesson, because it survives every re-cut of the lesson.
- `VERIFICATION.md` records two new run-sets with their run ids.
- The refuted claims #309 could not reach: `local/src/index.ts`'s *"WHY IT DOES NOT ROT"* block and `local/README.md`'s *"33, not 0"* section.
- **Withdrawn** from the README's troubleshooting: *"the fix is to add [the missing read path] to `ISSUE_LOOKUP_TOOLS`"*. That advice pointed readers away from the actual fix.

## Why

Two questions were open after #309, and both are now measured — `n=5` each, `claude-opus-5` pinned, one `group_id` each, `agent_version` declared, hosted:

**1 · Does this task need a planted defect at all? Yes.** Honest prompt, no denial, closed sandbox: **5/5 at 100**, `pass^5 = 100%`. There is no native reliability gap in this task on this model, so an honest configuration has no red to show and "unreliable → reliable" has nothing to work with.

**2 · Is the open sandbox the whole hole? No — closing it made the baseline greener.** Sandbox shut, denial unchanged: **5/5 at 100**, against 4/5 open-book. All five trials reached issue #1 through two paths `ISSUE_LOOKUP_TOOLS` does not name — `list_issue_comments` (5/5) and `update_issue` (5/5, a write that 404s on a missing issue and succeeds on a present one). `search_code` supplied the root cause, so losing the shell cost the agent nothing it needed. Two trials tried the denied `list_issues`, failed, and routed around it in the same turn.

Shutting three read paths surfaced two more the same afternoon. Completing the deny-list is not a viable baseline — the flaw has to stop being a tool-policy denial.

## Notes for reviewer

- **Merging this will flip the tracking ticket to Done, and it should not be.** The branch name carries the identifier, so the merge auto-closes it, but only one of its three "done when" bullets lands here. The baseline defect itself is untouched and still needs its re-cut. Move the ticket back after merge.
- **`allowedTools` does not restrict — it only auto-approves.** Measured: `allowedTools` naming one MCP tool left **152** tools live, `Bash`, `Read`, `Write` and `WebFetch` among them. `examples/triage-agent`, `examples/pr-summary-agent` and `examples/pr-summary-review` all rely on `allowedTools` to bound their tool surface and carry no `disallowedTools`, so all three are open-book *with* live web access. Not fixed here — each needs its own verification run, and doing it blind is how an unmeasured guard ships. Worth a follow-up before the template gets replicated.
- **No `verified red` stamp is written anywhere**, on purpose. Nothing in this PR earns one.
- The `WebSearch`/`WebFetch` entries in `deniedTools()` are now redundant (`tools: []` already removed them). Kept as a second latch and labelled as such — the two options are independent knobs and the web clamp should not depend on which one a future SDK version reinterprets.
- Worth knowing while reading the smoke gate: under a fully stripped environment (`env -i` with a minimal `PATH`) this examinee exits **0 with no output at all**, via `npm run start` and via `tsx` directly. Not `TMPDIR`, not a model pin, not the OTLP vars — each ruled out separately. Undiagnosed, pre-existing, and not addressed here. `smoke-examples.mjs` tolerates any exit code by design, so it would not catch this shape.

## Tests / CI

Local, in `examples/support-triage/local`:

- `npm test` — 12 passed (2 files), including the two new closed-sandbox properties
- `npm run typecheck` — clean

Live, rather than from the docs: with `tools: []` the SDK's `init` message listed 100 tools, **all** `mcp__*`, zero built-ins — and the run still scored 100 on all four criteria. That is the same run-set as measurement 1 above, so the sandbox close is verified by a scored hosted run and not only by a unit test.

PR CI (`typecheck-test`, `gitleaks + trufflehog`, `dependency review`) — checked after opening.

### DIFF
--- a/examples/support-triage/README.md
+++ b/examples/support-triage/README.md
@@ -200,11 +200,14 @@ The agent searches, finds issue #1, comments on it, and posts *its* link back to
 
 ### If your baseline passes / your fix fails
 
-* **Baseline passes (stays green).** Check the waterfall for a `search_issues` or
-  `list_issues` span that *succeeded* — if one did, the twin exposed a read path
-  the denial list does not name, and the fix is to add it (and to say so in
-  `ISSUE_LOOKUP_TOOLS`), not to weaken the criteria. `test/tool-policy.test.ts`
-  pins the list for exactly this reason.
+* **Baseline passes (stays green).** It does — 4/5 open-book, 5/5 with the
+  sandbox closed. This section used to say the fix is to add the missing read
+  path to `ISSUE_LOOKUP_TOOLS`. **That advice was measured wrong and is
+  withdrawn.** Shutting the first three paths surfaced two more the same
+  afternoon — `list_issue_comments` and `update_issue` — and the twin's read
+  surface is wide enough that the next pass would surface more. Do not extend the
+  list; the flaw has to stop being a denial ([F-1292](https://linear.app/pome-sh/issue/F-1292),
+  numbers in [`VERIFICATION.md`](./VERIFICATION.md)).
 * **Fix fails (stays red).** If a criterion reads `NOT EVALUATED` rather than
   failed, the run is `INCOMPLETE` — the grader could not see that state at all,
   which is a wiring problem rather than an agent problem. `pome run` exits 1 on
--- a/examples/support-triage/VERIFICATION.md
+++ b/examples/support-triage/VERIFICATION.md
@@ -232,3 +232,84 @@ So the duty is the opposite of what was written: a tool-denial baseline carries
 **high** rot risk and must be re-verified per model generation, exactly like the
 pattern-2 policy constant it replaced. A denial is only as strong as the
 enumeration behind it, and enumerations are never complete.
+
+---
+
+## The two follow-up measurements F-1292 asked for (2026-08-05)
+
+The 2026-08-04 numbers above said the baseline is green and named two routes.
+They left two questions open, and both were answered by measurement before any
+re-cut was designed. Same model (`claude-opus-5`, pinned), same task
+(`support-triage-dedup`, 4-criterion shape), same twin images, `n=5` each,
+`provenance: hosted`, `agent_version` declared on both run-sets.
+
+### 1 · Does the task need a planted defect at all? — **yes**
+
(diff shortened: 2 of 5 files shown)


AFFFPupu:
Greptile's finding was right, and it is this PR's own subject matter: asserting `BUILT_IN_TOOLS` is empty proves nothing while nothing asserts it reaches `query()`. Deleting `tools:` from the options object left every test green with the sandbox reopened — a guard disconnected from its subject passes forever, which is the same shape of mistake the measurements in this PR are about.

Fixed in af3d198: the composed options are now an exported `examineeOptions(mcpServers)`, and three tests assert the wiring rather than the constants — `tools` is `[]`, `disallowedTools` is `deniedTools()`, and `mcpServers` is still handed through. That third one earns its place too: if it regressed the examinee would have **no** tools at all, and the run would read as a model failure rather than a wiring one.

Declared as a `function` rather than a `const` arrow on purpose — this file's top-level `await main()` sits above it, and `scripts/smoke-examples.mjs` exists because exactly that temporal dead zone once shipped in `examples/triage-agent`.

15 tests pass, typecheck clean, and a local launch against a dead endpoint still evaluates the module and reaches `query()`.

PR #334, opened 2026-08-07
Model `stack` on both pull-request surfaces (F-1178)

GaganSD:
## What

`twin-github` now models `stack` on both pull-request read surfaces —
`GET /repos/{}/{}/pulls` and `GET /repos/{}/{}/pulls/{}` — the field GitHub
added to its `pull-request` and `pull-request-simple` schemas on 2026-08-02.

## Why

F-1178. Run 30755607225 on `main` was the first real upstream drift the declared
lane has ever caught:

```
FAIL — 0 unexpected twin surface(s), 2 shape-drift match(es)
  ! GET /repos/{}/{}/pulls     — shape drift (missing in twin: stack)
  ! GET /repos/{}/{}/pulls/{}  — shape drift (missing in twin: stack)
```

**Disposition: model it, not an allowlist entry.** The standing test is whether
an agent under evaluation would behave differently because the field is missing.
It would: stacked PRs change how an agent reasons about a PR's relationship to
its base, so an agent asked to review or merge a stacked PR against a twin that
cannot express stacking behaves differently than it does against real GitHub.

## The vendor schema this was built from

Read from the vendored REST description rather than inferred from the name.
Source: `pome-sh/openapi-spec-mcp` at commit `f0d07e7`, `specs/github.json`
(blob `af4eeae3c5c2c501efb87e9315895ad789472227`, confirmed against the remote
at that ref). Both `pull-request.stack` and `pull-request-simple.stack` are
`{"$ref": "#/components/schemas/pull-request-stack"}`, and neither the property
nor the schema exists at `a8f0142` (2026-07-31), matching the ticket's table.

`components.schemas["pull-request-stack"]`, verbatim:

```json
{
  "title": "Pull Request Stack",
  "description": "The stack information associated with a pull request.",
  "type": "object",
  "properties": {
    "base": {
      "type": "object",
      "properties": {
        "ref": { "type": "string", "description": "The base ref of the stack this pull request belongs to." },
        "sha": { "type": "string", "description": "The base SHA of the stack this pull request belongs to." }
      },
      "required": ["ref", "sha"]
    },
    "size": { "type": "integer", "description": "The total number of pull requests in the stack." },
    "position": { "type": "integer", "description": "The one-based position of this pull request within the stack, where 1 is the bottom of the stack." },
    "id": { "type": "integer", "description": "The ID of the stack that this pull request belongs to." },
    "number": { "type": "integer", "description": "The number of the stack that this pull request belongs to." }
  },
  "required": ["base"],
  "nullable": true
}
```

So: `nullable`, optional on its parents, only `base` required inside it, and
five integer leaves. The only other schemas carrying `stack` are webhook
payloads (`webhook-pull-request-stacked`, `webhooks_pull_request_5`), which the
twin does not model.

## How `stack` is populated

From real twin state, not a constant. GitHub models a stack as its own entity
with its own id and number; the twin has no stack table, so `pullRequestStack()`
reads a stack off the state that *does* encode one — the chain of **open** pull
requests in a repository linked `base_ref` -> `head_ref`. A PR whose base branch
is another open PR's head branch is stacked on it, which is what stacking means.

- `base.ref` / `base.sha` — the chain's bottom PR's base ref, and that ref's
  resolved branch head. Carries the same documented pre-resolution `null` as the
  existing `head.sha` / `base.sha` (FDRS-454).
- `size` — the chain length.
- `position` — this PR's one-based index from the bottom.
- `id` / `number` — **the two the twin cannot read from state**, stated plainly
  rather than faked: there is no stack row to carry them. Both derive from the
  chain's bottom PR (`id` via the same `stableNumericId` derivation every other
  twin id uses, `number` as that PR's number). That is deliberate: what an agent
  grouping PRs by stack needs is that every member reports the same pair, which
  deriving from a single member guarantees. The cost is that the twin's stack
  numbers share the per-repo number space issues and PRs draw from, where
  GitHub's do not.
- A PR outside such a chain reports `stack: null` — what the schema's
  nullability is for.

Three further limits, all written up in FIDELITY.md: a stack is inferred from
base targeting rather than declared, so PRs that merely happen to chain read as
stacked; a non-open PR reports `null`; and a forking chain takes the
lowest-numbered branch, since a GitHub stack is linear.

## Type anchor

`@octokit/openapi-types@28.0.0` carries openapi-version 22.0.0 and predates the
field, so `components["schemas"]` has no `pull-request-stack` to alias yet.
`src/upstream-types.ts` declares the shape locally, transcribed field for field,
with a note to replace it with the real alias at the next @octokit bump. The
`AssertNoUncovered` guard covers `stack` in either world, so no `_Allow` union
moved — nothing was registered as a deliberate omission.

## Notes for reviewer

- **TDD**: the three tests went in first and failed for the right reason
  (`expected { id: … } to have property "stack"`), then the implementation
  turned them green. In `packages/twin-github/test/v2-hot-paths-rest.test.ts`:
  - `GET /pulls/:n and GET /pulls both carry stack: null for an unstacked PR`
  - `GET /pulls/:n carries the vendor stack shape for every member of a stack`
  - `GET /pulls carries the same stack objects as the detail surface`
- **The post-publish 1:1 divergence lint stays green.** FIDELITY.md divergence
  #11 is *extended* rather than joined by a new numbered bullet: the lint binds
  bullets to `known-divergences/github.yaml` by bold title, so a new bullet
  would red it until pome-cloud shipped a matching entry. Verified by running
  pome-cloud's own `lint-known-divergences.ts` against this branch's
  FIDELITY.md — `OK — 20 entries 1:1`. `stack` is modelled, not omitted, so it
  does not belong on that bullet's omission list anyway; the sub-caveats about
  the derivation do.
- Versions: `twin-github` 0.9.3 for the twin release and cloud re-pin; the CLI
  inlines the twin into its bundle, so 0.21.13 carries it to npm. No changeset —
  Changesets is retired in this repo in favour of the version-diff lane.

## Review required

Yes — twin fidelity contract. `stack` is a new field on two `hot`/`semantic`
read surfaces, and the `id` / `number` derivation is a judgement call recorded
as a divergence rather than something the twin can read from state.

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -12,7 +12,8 @@ packaging restructure: bump `version` here and in `package.json`, and merging to
   (`GET /repos/:o/:r/pulls` and `.../pulls/:n`), which GitHub added to its
   `pull-request` and `pull-request-simple` schemas on 2026-08-02 (F-1178). An
   agent asked to review or merge a stacked PR against the twin can now see the
-  stack it belongs to. `packages/twin-github` changed, and the CLI inlines it.
+  stack it belongs to, and two PRs in one stack always agree on its identity,
+  size and membership. `packages/twin-github` changed, and the CLI inlines it.
 
 ## 0.21.12
 
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -4,6 +4,16 @@ Entries are hand-written from 0.9.0 on. Changesets was retired with the
 packaging restructure: bump `version` here and in `package.json`, and merging to
 `main` publishes (see `.github/workflows/release.yml`).
 
+## 0.21.13
+
+### Patch Changes
+
+- The bundled GitHub twin now models `stack` on both pull-request read surfaces
+  (`GET /repos/:o/:r/pulls` and `.../pulls/:n`), which GitHub added to its
+  `pull-request` and `pull-request-simple` schemas on 2026-08-02 (F-1178). An
+  agent asked to review or merge a stacked PR against the twin can now see the
+  stack it belongs to. `packages/twin-github` changed, and the CLI inlines it.
+
 ## 0.21.12
 
 ### Patch Changes
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pome-sh/cli",
-  "version": "0.21.12",
+  "version": "0.21.13",
   "description": "Digital-twin testing for AI agents \u2014 run tasks against resettable local or hosted twins and record tool-call traces for evaluation on pome.sh.",
   "keywords": [
     "ai",
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
     },
     "cli": {
       "name": "@pome-sh/cli",
-      "version": "0.21.11",
+      "version": "0.21.13",
       "license": "Apache-2.0",
       "dependencies": {
         "@ai-sdk/anthropic": "^4.0.29",
@@ -7318,7 +7318,7 @@
     },
     "packages/checks": {
       "name": "@pome-sh/checks",
-      "version": "0.1.0",
+      "version": "0.1.1",
       "license": "Apache-2.0",
       "devDependencies": {
         "@pome-sh/sdk": "*",
@@ -7342,7 +7342,7 @@
     },
     "packages/sdk": {
       "name": "@pome-sh/sdk",
-      "version": "0.11.3",
+      "version": "0.11.4",
       "dependencies": {
         "@pome-sh/wire": "*",
         "hono": "^4.13.0",
@@ -7369,7 +7369,7 @@
     },
     "packages/twin-github": {
       "name": "@pome-sh/twin-github",
-      "version": "0.9.2",
+      "version": "0.9.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@hono/node-server": "^2.1.0",
--- a/packages/twin-github/CHANGELOG.md
+++ b/packages/twin-github/CHANGELOG.md
@@ -1,6 +1,28 @@
 # @pome-sh/twin-github — CHANGELOG
 
 
+## 0.9.3 — 2026-08-08
+
+`stack` is modelled on both pull-request read surfaces (F-1178). GitHub shipped
+stacked pull requests and added the field to both the `pull-request` and
+`pull-request-simple` schemas on 2026-08-02, and the declared lane caught the
+twin missing it on `GET /repos/{}/{}/pulls` and `GET /repos/{}/{}/pulls/{}`.
+
+The shape is transcribed from the vendored `pull-request-stack` schema, not
+guessed: nullable, with a required `base: { ref, sha }` and optional integer
+`size`, `position`, `id` and `number`. `@octokit/openapi-types@28.0.0` predates
+the field, so `src/upstream-types.ts` declares it locally until that bump lands;
+the `AssertNoUncovered` guard covers it either way.
+
+It is populated from twin state, not a constant. GitHub models a stack as its
+own entity; the twin has no stack table, so it reads one off the chain of OPEN
+pull requests linked `base_ref` -> `head_ref` in a repository. `base`, `size`
+and `position` are exact consequences of that chain; `id` and `number` are
+derived from the chain's bottom PR, so every member of one stack reports one
+identity. A PR that is not in such a chain reports `stack: null`. The limits are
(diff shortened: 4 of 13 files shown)


GaganSD:
Pushed `a820a89` addressing the two independent review passes. Summary of what changed and why.

### 1 (high, must-fix) — forked stacks handed out one `stack.id` with two different answers

Confirmed and fixed. The old walk started at the PR being serialized and took the first match at each step, so with bottom A and two open children B, C (`B.number < C.number`), asking A gave `[A,B]` and asking C gave `[A,C]`, both deriving `id`/`number` from A. Two PRs, one `stack.id`, different `size` and membership. The cycle case had the same defect — A and B each resolved to a *different* bottom, so it was never just a termination question.

Root cause: the identity is shared across members, so the answer cannot depend on where the walk started. `pullRequestStack` now resolves the whole **connected component** (breadth-first, both edge directions, seen set) and then requires it to be one unambiguous line — no node with two parents or two children, exactly one bottom, and the ordered walk must cover the component. A fork or a cycle reports `stack: null` from **every** member rather than an identity two of them would dispute.

I took the "must not hand out a shared identity" branch rather than trying to canonicalise a fork, and you were right that lowest-numbered-from-the-parent does not achieve coherence — I checked the alternative of deriving identity from the membership instead, and it still leaves C claiming a 2-member stack whose other member denies being in it.

**The invariant test** (`keeps stack identity coherent: equal stack.id implies equal size, membership and unique positions`) builds a repo holding a clean stack *and* a fork at once, groups the whole `GET /pulls` response by `stack.id`, and asserts per bucket: all members agree on `size`; the bucket's own length equals that `size` (no absent siblings); positions sort to exactly `1..size`; and one agreed bottom `number` and `base`. The fork test asserts null **from every member** (bottom, left, right) as asked.

### 2 (must-fix) — closed middle link silently reported a live chain as null

Fixed, and made the stated choice you preferred: linkage traverses pull requests of **every state**, membership stays the **open** members. A closed middle is skipped in `size`/`position` but does not sever the PRs above and below it; a non-open PR reports `null` for itself; fewer than two open members is no stack. Tested both directions (`links through a closed middle PR and counts only the open members`, `reports no stack once only one open member is left`).

### 3 — cross-repo collisions

Verified rather than assumed: `PullRequestRow` carries `head_repo_id` and `base_repo_id`, and the twin **does** model fork PRs (`resolveHeadRef` accepts `owner:branch`). Edges now match repo id as well as ref name. I proved the test catches the old behaviour by weakening the predicate back to name-only — it fails with `expected { base: { ref: 'main', … } } to be null`, i.e. a stack invented purely from a branch-name collision.

### 4 — N+1 on the list surface

`listPullRequests` reads the table once and threads it through `serializePullSimple`. Measured over a repo with 8 stacked PRs, one `GET /pulls`:

| | full-table `pull_requests` reads |
| -- | -- |
| before | **9** (N+1) |
| after | **1** (independent of N) |

It passes the **unfiltered** list on purpose: a PR's stack is a fact about the repo, so neither the state filter nor the page may change it. Locked by `does not let the state filter or pagination change a PR's stack`.

### 5 — FIDELITY.md overstated the guarantee

Corrected. The "lowest-numbered branch" claim is gone; divergence #11 now states the invariant in the terms above plus four limits — inferred-not-declared, linear-only (fork/cycle → null), linkage-spans-states while membership is open-only, and unnamed when the base sha is unresolved — each with a test behind it. The two tool rows and the matrix rows no longer say "open-PR base chain" either. pome-cloud's own `lint-known-divergences.ts` still reports `OK — 20 entries 1:1` against this branch.

### 6 — `sha` cast

Cast removed. An unresolved base sha now reports `stack: null` rather than publishing a null into a field the vendor types non-null. It is not reachable through the public routes today (PR creation requires both branches to resolve, and the seed has no branch-without-commits path), so it is a one-line guard rather than a tested case — flagging that rather than claiming coverage.

### 7 — test strength

The convergence test now asserts **concrete values** on the list surface (base ref/sha, size, position, bottom number, shared id) before checking that the two surfaces agree field for field, so it no longer passes for a constant. New tests: fork-from-every-member, the identity invariant, cycle, closed middle, drained stack, fork-name-collision, filter/pagination stability. Six of the ten F-1178 tests fail against the previous implementation; the other four guard behaviour it already had.

### Verification

`npm ci`, `npm run build`, `npm run typecheck`, full `npm test` (all 10 workspaces green, twin-github 429/429, CLI 1022/1022), `test:contract`, `test:pack`, `probe:twins`, `probe:examples`, `smoke:examples`, `typecheck:examples`, `fidelity:parity`, `gate:checks-tarball`, and every lint gate in the invariants table plus the version-bump gate.

PR #336, opened 2026-08-07
Three seed-to-finish deltas, so a restraint task can assert restraint (F-1304)

AFFFPupu:
## Why

F-1304 measured the exam half of the task corpus — the 24 `restraint` +
`adversarial` files — and found them resting on one `[code]` assertion each, with
the real verdict falling to a single `[model]` line. That is the grader F-1301
proved an examinee can talk out of a finding.

Repairing it needs criteria that say *"the examinee did not disturb what it was
told to leave alone"*. Every existing check that could express one **reads a
single world** — so on a restraint task, whose correct finish state IS its seed
state, it is true before the examinee starts. The discrimination gate calls that
`already_satisfied`; F-1296 calls it a free point. A one-world check cannot tell
a careful agent from an absent one. **A delta can.**

This PR is the vocabulary only. No task file changes here — those land after
pome-cloud pins the published `@pome-sh/checks@0.2.0`, or the new sentences bind
nothing and the corpus gate goes red.

## What

| check | replaces | for |
| -- | -- | -- |
| `slack.no-new-message-in-channel` | `no-message-posted` (unwinnable on a seeded channel) | task 21's message half |
| `github.no-commit-status-changed` | `commit-status … is failure` (true on the seed) | task 18's forged green build |
| `github.issue-triage-unchanged` | `issue-exactly-one-label` + `issue-assignee` (both pinned non-discriminating since F-1075) | task 03's standing obligation |

`@pome-sh/checks` 0.1.1 → **0.2.0**. Minor, not patch: `GITHUB_CHECKS` goes
15 → 17 and `SLACK_CHECKS` 5 → 6, so a consumer pinned to 0.1.x would grade a
corpus that binds sentences it cannot resolve.

## Reviewer notes

**The Slack one is the interesting one — three earlier shapes failed, each on a
different gate**, and all three are written onto the check so none gets
re-proposed:

1. `no-message-posted` counts a channel's messages *including seeded ones* →
   false before the examinee starts. F-1303 deleted it as the corpus's only
   `failRestsOnUnpassable` row.
2. Repointing it at a channel the seed leaves empty → passes even with the
   exported `messages` array deleted. P2 catches that as a clean bill over unread
   state.
3. A `no-message-containing` needle scan → binds, passes on the seed in the right
   cell, and **still** dies on P2 in five places including `channels[].messages`.

The invariant: **a scan cannot distinguish "nothing matched" from "nothing was
read"**, and a prohibition that cannot distinguish those is the negative
false-pass D4 forbids outright. The `state_incomplete` skips in these three
checks are not defensive boilerplate — they are the reason these exist rather
than any of the above.

**What is deliberately NOT here:** widening `TAPE_ASSERTABLE_TOOLS` to
`close_issue` / `merge_pull_request`. Those would give `triage-agent/02` and
07/09/18 a genuinely independent tape substrate, but the stamping happens in the
twin **server** (`routes.ts`), so it is the runtime axis — a snapshot rebuild and
a prod `TWIN_GITHUB_SNAPSHOT_ID` flip, not a checks publish. `close_issue` also
shares `PATCH /repos/:owner/:repo/issues/:number` with label and assignee
updates, so it needs a conditional stamp the recorder does not currently support.
Filed rather than smuggled in.

## Verification

- twins workspace: all 10 packages green (`npm run test --workspaces`)
- `lint:task-class`, `lint:legacy-markers`, `lint:code-health`, `lint:no-catch`,
  `gate:no-eval` all pass
- each new check has its `checks-contract.test.ts` fixture, its
  `HONEST_NULL_MUTANTS` reason, and its missing-repo predicate fixture

Merging publishes `@pome-sh/checks@0.2.0` via `release.yml`'s version-diff lane.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
     },
     "cli": {
       "name": "@pome-sh/cli",
-      "version": "0.21.11",
+      "version": "0.21.12",
       "license": "Apache-2.0",
       "dependencies": {
         "@ai-sdk/anthropic": "^4.0.29",
@@ -7318,7 +7318,7 @@
     },
     "packages/checks": {
       "name": "@pome-sh/checks",
-      "version": "0.1.0",
+      "version": "0.1.1",
       "license": "Apache-2.0",
       "devDependencies": {
         "@pome-sh/sdk": "*",
@@ -7342,7 +7342,7 @@
     },
     "packages/sdk": {
       "name": "@pome-sh/sdk",
-      "version": "0.11.3",
+      "version": "0.11.4",
       "dependencies": {
         "@pome-sh/wire": "*",
         "hono": "^4.13.0",
--- a/packages/checks/CHANGELOG.md
+++ b/packages/checks/CHANGELOG.md
@@ -1,5 +1,41 @@
 # @pome-sh/checks
 
+## 0.2.0
+
+Three new declarations, all of them SEED-TO-FINISH DELTAS, and that is the whole
+point of the release rather than a coincidence of batching. Minor rather than
+patch because the published surface grows: `GITHUB_CHECKS` goes 15 -> 17 and
+`SLACK_CHECKS` 5 -> 6, so a consumer pinned to 0.1.x grades a corpus that can
+now bind sentences it cannot resolve.
+
+F-1304 measured the exam half of the twins task corpus and found its restraint
+and adversarial tasks resting on one assertion each, with the real verdict
+falling to a single `[model]` line. Repairing that needs criteria a task can
+carry for "the examinee did not disturb what it was told to leave alone" — and
+every existing check that could express one READS A SINGLE WORLD, so on a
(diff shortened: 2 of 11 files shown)


AFFFPupu:
Recut decision (2026-08-12, approved by Ao): closing without merge. Five days behind a main that landed four twins merges plus the criterion-grammar language-table gate (F-1458/F-1495) — and delta vocabulary is exactly what that gate governs. The three seed-to-finish deltas in this diff now have dedicated owners in the One Working Curriculum project: positive tape assertion → F-1338 (M0, first), slack message delta → F-1340 (M2), github issue-scoped delta → F-1341 (M2). Branch kept as salvage and referenced from those tickets: ao/f-1304-exam-half-graded-by-one-model-line. Note for the recut: the title carried the parent id F-1304 — each recut PR carries its own sub-issue id.

PR #338, opened 2026-08-07
twin-linear's agent-session mutation inputs take Linear's field names (F-1176)

GaganSD:
Closes F-1176.

## The decision came first

Recorded as a `[DECISION]` comment on F-1176 before any code: **option 1, full fidelity**, with one scoped known-divergence.

**Live re-verification.** Re-introspected `https://api.linear.app/graphql` unauthenticated rather than trusting the ticket. The ticket's four lists are exactly right. Two extras it did not have:

- `id` on `agentSessionUpdate` is a **mutation argument**, and non-null upstream — `agentSessionUpdate(id: String!, input: AgentSessionUpdateInput!)`. The twin declared it nullable. Second divergence on the same mutation.
- `AgentActivityContent` is a **union, introspectable member by member**: `{type, body, bodyData}` for thought/response/elicitation, `{type, body, title, bodyData}` for prompt, `{type, body, reasonCode, bodyData}` for error, `{type, action, parameter, result, resultData}` for action. So `type` + `body` → `content: {type, body}` is a rename with one level of nesting, not the "different design" the ticket assumed. `AgentActivitySignal` = `stop | continue | auth | select`.

**Two stale notes corrected.** `scripts/regen-linear-introspection.mjs` claimed `AgentSessionCreateOnIssue` / `OnComment` "are not Linear type names at all". They are — both exist upstream under those exact names. (`AgentSessionCreateInput` also exists, separately; that is where `appUserId` lives upstream.) Separately: `externalUrls`, `externalLink`, `externalLinks`, `type` and `workspaceDiffFiles` on the `AgentSession` output type are **deprecated but still declared** — they appear only under `includeDeprecated: true`. The regen script already passes it, so the fixture is right, but anyone re-checking by hand without the flag will wrongly conclude Linear removed them.

**Blast radius, measured.** `cli/tasks/` (42 files): **0**. `examples/` (8): **0**. Anything driving `agentSessionUpdate` with `status`, anywhere: **0**. Every caller is inside `packages/twin-linear`. The corpus migration option 1 was sized against is empty, and with zero customers there is no shim to write — so option 1 is materially cheaper than the ticket assumed.

**Against option 2 (names only).** Its premise is that the 1:1 mapping runs out; the introspected content union shows it does not. It would leave `status` on the update input — the single most consequential divergence in the ticket — and register a gap that cost near zero to close. It also leaves the guard dishonest: the four types would pass against an allowlist rather than against upstream.

**Against option 3 (register wholesale).** Writes down "our mutation inputs are not Linear's" as policy for the one surface where an agent's arguments cross the wire, which is the failure the twins exist to prevent and the thing F-1172 refused to do on the output side. Worse, it ratchets the drift guard onto a wrong baseline, freezing `sessionId` / `type` / `body` / `status` as protected twin vocabulary. Registering is for divergences you cannot close.

## The guard's starting red

Extending `GUARDED_TYPES` was the first commit-worthy act, and it failed on all five newly-covered types immediately — the ticket's starting evidence:

```
FAIL AgentSessionUpdateInput declares no field Linear does not have
  AgentSessionUpdateInput declares members Linear does not: id, status
FAIL AgentActivityCreateInput declares no field Linear does not have
  AgentActivityCreateInput declares members Linear does not: body, sessionId, type
FAIL AgentSessionCreateOnIssue declares no field Linear does not have
  AgentSessionCreateOnIssue declares members Linear does not: appUserId, plan
FAIL AgentSessionCreateOnComment declares no field Linear does not have
  AgentSessionCreateOnComment declares members Linear does not: appUserId, plan
FAIL AgentActivity declares no field Linear does not have
  AgentActivity declares members Linear does not: body, session, type

Tests  5 failed | 7 passed (12)
```

All nine guarded types now pass on merit, with no allowlist entry.

## What changed

| input | removed | now |
| -- | -- | -- |
| `AgentSessionCreateOnIssue` | `appUserId`, `plan` | `issueId`, `externalUrls` |
| `AgentSessionCreateOnComment` | `appUserId`, `plan` | `commentId`, `externalUrls` |
| `AgentSessionUpdateInput` | `id`, `status` | `plan`, `externalUrls` |
| `AgentActivityCreateInput` | `sessionId`, `type`, `body` | `agentSessionId`, `content`, `signal`, `ephemeral` |

`agentSessionUpdate(id:)` is now `String!`.

**Status is no longer settable.** Linear has no field to set it with; it follows the emitted activity, which is Linear's documented model ("updated automatically based on the agent's emitted activities. No manual state management is required."). `agentActivityCreate` moves it: `thought`/`action` → `active`, `elicitation` → `awaitingInput`, `response` → `complete`, `error` → `error`, `prompt` → `pending`. Linear does not publish that table and introspection cannot show behaviour, so it is the **one** invented thing here and it is registered in `REFERENCE-DIVERGENCES.md` as twin-owned, not passed off as upstream truth. Deriving nothing would freeze every session at `pending`, which is worse fidelity than a named guess.

**`AgentActivity` follows the input** onto `content` / `agentSession` / `signal`, `user: User!`. Accepting Linear's `content` while emitting an invented `body` would round-trip to neither Linear nor itself. `agent_activities` gains `content_json` and `signal` and drops `body`; an existing `LINEAR_TWIN_DB` file migrates on open, with `action` rows backfilling an empty `parameter` because the old shape had nowhere to keep one.

`appUserId` survives on the **domain** command only, never over GraphQL — the twin creates sessions internally when an issue is delegated or an app is @mentioned, and upstream that identity arrives with the credential rather than in the input.

## The guard can fail, and that is now a test

Not a manual observation. `it.each` splices an invented member into the twin's **own** SDL via `printSchema`, rebuilds it through `buildSchema` exactly as `src/graphql/schema.ts` does, and asserts the guard names exactly that member — for `status` on the update input, `appUserId` on both creates, `sessionId` on activity create, `body` on `AgentActivity`, `externalUrl` on `AgentSession`, plus an invented enum member. Each case also asserts the real schema reports `[]` first, so the test cannot pass by comparing nothing. If the check is ever softened into a no-op, these go red.

Also added `test/agent-session-inputs.test.ts`: every invented field is rejected over the wire with the GraphQL error naming it, `id` is required as an argument, the full six-row status table is exercised end to end, `signal` round-trips and an undeclared signal is rejected, and malformed `content` (missing body, unknown key, non-`AgentActivityType`, action without parameter) is a loud error rather than a silent stub.

## Migration

No corpus change was needed. In-repo callers migrated in this PR: `partial-update-tristate.test.ts` seeds a plan through `agentSessionUpdate` instead of the create input and no longer sets `status`; `domain/{agents,normalize,rows,index,linear-domain}.ts`, `graphql/{schema,mutation-inputs,resolvers,formatters}.ts`, `types.ts`, `db.ts` and `state.ts` follow the rename.

## Left open, deliberately

Two scalar-level divergences, written up in `REFERENCE-DIVERGENCES.md` rather than fixed, because the guard is name-based by F-1172's design: input `plan` is `String` where upstream is `JSONObject`, and output `content` is `JSON!` where upstream is the `AgentActivityContent` union.

## Gates

`npm run typecheck` exit 0. `npm test` exit 0 — 10 suites, all passing. `knip`, `lint-code-health`, `check-copy-markers`, `no-catch-and-continue`, `check-twin-leaf-portability`, `check-twin-chunk-laziness`, `check-workspace-pins-match-workspace`, `lint-no-cloud-imports`, `capture-mcp-tools-list --check --offline` and `check-version-bump-required` all clean. Versions: twin-linear 0.4.0, `@pome-sh/cli` 0.21.14 (the CLI inlines the twin into its bundle).

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -4,6 +4,20 @@ Entries are hand-written from 0.9.0 on. Changesets was retired with the
 packaging restructure: bump `version` here and in `package.json`, and merging to
 `main` publishes (see `.github/workflows/release.yml`).
 
+## 0.21.15
+
+### Patch Changes
+
+- **Breaking for agents driving Linear agent sessions.** The bundled Linear
+  twin's four agent-session mutation inputs now take Linear's field names, and
+  only Linear's (F-1176). `AgentSessionUpdateInput` drops `id` and `status`,
+  `AgentSessionCreateOnIssue` / `OnComment` drop `appUserId` and `plan`, and
+  `AgentActivityCreateInput` takes `agentSessionId` + `content` in place of
+  `sessionId` / `type` / `body`. Session status is no longer settable — Linear
+  has no field for it, so status follows the emitted activity. No task in the
+  corpus drove these mutations, so nothing in `cli/tasks/` or `examples/`
+  changed.
+
 ## 0.21.14
 
 ### Patch Changes
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pome-sh/cli",
-  "version": "0.21.14",
+  "version": "0.21.15",
   "description": "Digital-twin testing for AI agents \u2014 run tasks against resettable local or hosted twins and record tool-call traces for evaluation on pome.sh.",
   "keywords": [
     "ai",
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
     },
     "cli": {
       "name": "@pome-sh/cli",
-      "version": "0.21.14",
+      "version": "0.21.15",
       "license": "Apache-2.0",
       "dependencies": {
         "@ai-sdk/anthropic": "^4.0.29",
--- a/packages/twin-linear/CHANGELOG.md
+++ b/packages/twin-linear/CHANGELOG.md
@@ -1,6 +1,45 @@
 # @pome-sh/twin-linear — CHANGELOG
 
 
+## 0.4.0 — 2026-08-08
+
+**Breaking.** The four agent-session mutation inputs now take Linear's field
+names, and only Linear's (F-1176). F-1172 fixed the `AgentSession` output type
+and left the inputs inventing fields; verified against live introspection at
+`https://api.linear.app/graphql`:
+
+| input | removed | now |
+| -- | -- | -- |
+| `AgentSessionCreateOnIssue` | `appUserId`, `plan` | `issueId`, `externalUrls` |
+| `AgentSessionCreateOnComment` | `appUserId`, `plan` | `commentId`, `externalUrls` |
+| `AgentSessionUpdateInput` | `id`, `status` | `plan`, `externalUrls` |
+| `AgentActivityCreateInput` | `sessionId`, `type`, `body` | `agentSessionId`, `content`, `signal`, `ephemeral` |
+
+`agentSessionUpdate(id:)` is now `String!` — upstream `id` is a mutation
+argument on this mutation, never a field of the input.
+
+**Session status is no longer settable.** Linear declares no `status` on
+`AgentSessionUpdateInput`; status follows the emitted activity, as Linear's agent
+guide describes ("updated automatically based on the agent's emitted
+activities"). `agentActivityCreate` now moves it: `thought`/`action` → `active`,
+`elicitation` → `awaitingInput`, `response` → `complete`, `error` → `error`,
+`prompt` → `pending`. Linear does not publish that table, so it is registered as
+a twin-owned divergence in `REFERENCE-DIVERGENCES.md`.
+
+`AgentActivity` follows the input: `content: JSON!` and `signal` replace `type`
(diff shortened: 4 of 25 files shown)


AFFFPupu:
Heads-up from today's F-1511 landing: version numbers are now pipeline-written. A PR that edits a published package's `version` fails the new gate (`check-release-note-required.mjs`); the PR carries `## Unreleased (patch|minor)` CHANGELOG prose instead, and `allocate-version.yml` mints the number after merge — RELEASING.md has the whole model.

For this branch: drop the cli 0.21.14→0.21.15 bump and put the changelog text under a `## Unreleased (patch)` heading at the top of `cli/CHANGELOG.md`; twin-linear's own private version/CHANGELOG are gate-exempt (keep or drop, your call).

I converted #405 and #402 to the new shape mechanically today, and attempted the same here but **aborted without touching this branch**: a rebase onto current main conflicts in 20 files, 10 of them `twin-linear/src` — F-1468/F-1480 rewrote the projection/validator layer since this branch's 08-07 base, so the resolution is semantic and yours. Rebase vs recut is your call; the note-conversion itself is one small commit once you're on current main.

PR #371, opened 2026-08-11
twin-linear declares `extensions` and answers it the way Linear does (F-1385)

AFFFPupu:
## What this does

`twin-linear` declares `extensions` on both `/graphql` surfaces and answers it the way Linear does — **before the twin's authentication check**. GraphQL-over-HTTP's fourth envelope member is what Apollo clients send for automatic persisted queries. The twin declared it nowhere, so once F-1372 flipped this twin to `ignore`, the payload was discarded and the query served, where real Linear answers on it.

Closes F-1385. Last of the three sub-issues under F-1456.

## ⚠️ Read this first — the ticket's premise was wrong

F-1385 was filed, and its `[DECISION]` argued, on the reading that Linear's 400 is **"automatic persisted queries simply switched off"** — an unhandled path rather than a contract. I re-measured `https://api.linear.app/graphql` on **2026-08-11** before building, and that reading is falsified.

**Linear runs APQ, in verify-only mode.** It checks that `sha256Hash` is the SHA-256 of the `query` it arrived with, and the 400 is that check failing:

```
Q='{__typename}'; H=$(printf '%s' "$Q" | shasum -a 256 | cut -d' ' -f1)

{"query":"…"}                                                          -> 401 AUTHENTICATION_ERROR
{"query":"…","extensions":{"persistedQuery":{"version":1,"sha256Hash":"abc"}}}  -> 400 INTERNAL_SERVER_ERROR
{"query":"…","extensions":{"persistedQuery":{"version":1,"sha256Hash":"$H"}}}   -> 401 AUTHENTICATION_ERROR   ← served
{"extensions":{"persistedQuery":{"version":1,"sha256Hash":"$H"}}}               -> 200 PersistedQueryNotFound
{"query":"…","bogusKey":{"a":1}}                                       -> 401 AUTHENTICATION_ERROR   ← ignored
```

**Why this matters for the build.** The `[DECISION]`'s literal shape — reject every `extensions` payload with 400 — would have made the twin answer 400 for the correct-hash case, where Linear serves. An Apollo client with APQ enabled sends exactly that on its retry. The twin would have failed an agent that works in production: the mirror of the failure class the `[DECISION]` itself names, and the thing `docs/undeclared-route-inputs.md` warns about ("refusing where the vendor accepts scores an agent for a failure it did not commit").

So the ruling's **shape is built as decided** — declare on both surfaces, mirror Linear's answer, reject pre-auth — against a correct measurement of what "the way Linear does" is. The result is better than the `[DECISION]` expected: **no recorded divergence on any measured case**, rather than the fallback "record the divergence" clause.

**"Implement persisted queries" stays rejected, and is now known to be wrong upstream too.** Nothing is registered by either side — a hash-only request answers `PersistedQueryNotFound` even straight after the same hash arrived with its query. There is no store to mirror, only a hash to verify. The fix carries no persisted-query cache.

The `[DECISION]`'s own instruction covers this: *"A future re-measure that returns 200 is a signal to re-decide, not drift to suppress."* Surfacing it, not suppressing it — but flagging it here and on the ticket rather than deciding it silently.

## The contract, as built

| Request | Twin, and Linear |
| --- | --- |
| `extensions` absent, or carrying no `persistedQuery` | served |
| `persistedQuery` whose `sha256Hash` = SHA-256 of `query` | served |
| hash mismatch (incl. right hash, wrong case), `version` ≠ 1, or no hash | 400 `INTERNAL_SERVER_ERROR` |
| `persistedQuery` with no `query` to verify | 200 `PersistedQueryNotFound` |
| `extensions` that is not a usable object | 400 `BAD_REQUEST`, worded per surface |

One measured asymmetry the two surfaces do **not** share, and the reason neither derives from the other: `extensions: null` is **ignored** in a POST body and **refused** in a query string.

## The pre-auth ordering

Mounted on a router wrapped around the engine's session app, the way `withPublicOAuth` already mounts authorize/token/revoke — so `bearerAuth` has not run yet. Reject after the auth check and an agent sending an APQ payload with a stale token sees 401 here and 400 at Linear: the same divergence, harder to see.

Pinned with a **deliberately-bad token**, per the `[DECISION]`. The sharpest assertion is the last one: with the same bad token, an APQ descriptor Linear *is* satisfied by falls through and answers 401 exactly like the control — which a "reject `extensions` outright" fix would fail.

Two deliberate consequences of sitting outside the session, both documented in the source: the gate's answers are off the recorder tape (like the `bearerAuth` 401s it stands in front of), and it never reports a malformed request — anything it cannot parse is handed to the recorded handler and answered through the twin's own error envelope. That fall-through has its own test.

## Reviewer notes

- **`src/graphql/persisted-query.ts`** is the whole contract as one pure function of the parsed envelope — no hono, no request object. Both surfaces read it, which is what stops them drifting apart.
- **`extensions` is declared deliberately permissively** (`z.unknown()` on POST, `z.string()` on GET). Linear words a different 400 for each way the value can be wrong and answers them itself; a zod schema tight enough to reject them here would substitute the twin's `BAD_USER_INPUT` envelope for the vendor's message on every one — a divergence manufactured by our type rule rather than measured. The derived `type` is `null` on POST because the wire really does accept any JSON value there.
- **Declared inputs 120 → 122**, 979 → 981 repo-wide. `route-inputs.json` regenerated with `npm run emit:route-inputs` after `npm run build`.
- **No "Known divergences" bullet was touched.** twin-linear's `FIDELITY.md` has no such section, so pome-cloud's `lint-known-divergences.ts` 1:1 title binding is not in play here. The new material is a `###` subsection under the existing *Declared input surface (F-1179)* heading.
- **`FIDELITY.md`'s input counts were already stale on `main`** ("31 of them required"); the artifact said 32 before this change too. Corrected to the derived numbers while editing the line.

## Left alone, flagged not fixed

- **Real Linear blocks a bare `GET /graphql` as CSRF** (400 unless `x-apollo-operation-name` or `apollo-require-preflight` is present). Every GET measurement here sent that header. The twin models no CSRF prevention at all — a pre-existing gap on a different axis from route inputs. Recorded in `FIDELITY.md` and the doc; no ticket, no scope widening.
- **`INTERNAL_SERVER_ERROR` with `userError: false`** for a bad hash still does not read like a designed contract (Apollo's own vocabulary has `PERSISTED_QUERY_ID_INVALID`). The twin mirrors what was measured, and `FIDELITY.md` says so out loud — an agent meets the measurement, not the intent.

## Verification

Everything below run locally on this branch:

- `npm run test` — **3,422 tests green** across all workspaces (twin-linear 191, +21 new).
- `npm run build && npm run typecheck && npm run gate:route-inputs` — clean.
- Gates: `lint:route-inputs`, `lint:code-health`, `no-catch-and-continue`, `check-twin-leaf-portability`, `check-twin-chunk-laziness`, `check-copy-markers`, `check-first-party-twin-registration`, `check-workspace-pins-match-workspace`, `gate:no-native`, `gate:no-eval`, `gate:bundled-deps`, `lint:dead-code`, `probe:twins` (128 probes), `fidelity:parity -w @pome-sh/twin-linear`, `test:contract` (104 pass / 5 skip).
- **The version gate was confirmed to bite**: it named `@pome-sh/cli` as unbumped before the bump and passes after. `@pome-sh/cli` 0.23.14 → **0.23.15**, `@pome-sh/twin-linear` 0.4.0 → **0.4.1**.
- **The session-mount test was confirmed to bite**: removing `/s/:sid/graphql` from the gate's mounts fails it, and nothing else.


AFFFPupu:
---

## Update — a regression this PR introduced and then fixed (`4f11eb4`)

The first draft of the gate parsed `c.req` directly. That **silently blanked `request_body` on every recorded `/graphql` request**, and the full 3,422-test suite stayed green because nothing in `twin-linear` reads the tape.

The engine's recorder captures `request_body` with its own `c.req.raw.clone().json()`, before invoking the handler. `clone()` throws once the body stream has been disturbed, and `readRequestJson` catches it and records `null` rather than failing — so a middleware that drains the body ahead of the recorder loses the tape with no error anywhere. It is the exact hazard `twin-stripe/src/routes/_helpers.ts` carries an exemption comment about, from the other direction ("It must happen BEFORE the declaration parses").

**Fix:** the gate hands the declaration a `HonoRequest` over `c.req.raw.clone()`, so the original stream is untouched. The recorder's own clone still succeeds, and the handler behind the gate parses as before. The declaration remains the only thing that reads a value by name — the clone changes *which* `Request` is parsed, not *what* may be read off it.

**Now permanently pinned**, since no existing test would have caught it: `test/route-input-declarations.test.ts` asserts the recorded event carries the request body. Confirmed to bite — reverting the clone fails that test and nothing else.

Two follow-on changes this needed:
- `scripts/lint-route-input-declarations.mjs` gains one exemption for `new HonoRequest(c.req.raw.clone())` in `twin-linear/src/twin.ts`, with the reason above. It is the only entry in the list that grants a clone rather than a named read.
- `scripts/lint-route-input-declarations.test.mjs` gains the matching fixture (the suite requires every exempted file to exist in its throwaway tree), and its baseline twin count moves 4 → 5.

**Worth a look beyond this PR:** the engine hazard is general. Any twin middleware that touches a body before `recorder.handle` blanks that twin's tape, silently. Not fixed here — it is SDK-wide and outside F-1385 — but flagged.

PR #375, opened 2026-08-12
test:contract runs every file in contract/, by discovery (F-1353)

GaganSD:
## What was wrong

`contract/run.mjs:29-34` hand-listed three contract test files
(`contract.test.mjs`, `sdk-boot.test.mjs`, `gmail-fault.test.mjs`).
`contract/cli-start.test.mjs` existed but was never on that list, so
`npm run test:contract` silently never ran it. Coverage depended on
two independently hand-maintained lists (run.mjs's enumeration and
ci.yml's separate `node --test contract/cli-start.test.mjs` invocation)
with nothing tying them together.

## What changed

- `contract/run.mjs` now exports `discoverTestFiles()`, which reads
  `contract/` and returns every `*.test.mjs` file, sorted, minus one
  documented exclusion. The build+run pipeline is now guarded behind
  `import.meta.main` so importing `discoverTestFiles()` for tests
  doesn't re-trigger the npm build chain and re-spawn `node --test`
  on itself (that recursion is what happened on the first pass here).
- `cli-start.test.mjs` is the one deliberate exclusion, investigated
  rather than assumed: it boots the CLI's own compiled entry
  (`cli/dist/src/cli/main.js`), and `run.mjs` only builds its own five
  packages (wire, sdk, twin-*), never `cli/`. Folding it in would mean
  giving `run.mjs` a redundant cli build step just so a bare
  `npm run test:contract` stays runnable standalone — not worth it for
  one file. It keeps running as ci.yml's own step, after the heavy
  job's earlier full `npm run build` already built `cli/`. Documented
  the reason in three places: `run.mjs`'s exclusion comment,
  `cli-start.test.mjs`'s own header (which pointed at a `cli-ci.yml`
  that no longer exists — folded into `ci.yml` a while back), and
  ci.yml's step comment.
- Added a row to AGENTS.md's P8 invariants table for this mechanism.
- `contract/run.test.mjs`: the tdd-required regression, written first
  and confirmed failing before the implementation existed. Fixture-dir
  cases pin the discovery/exclusion mechanism; the last case recomputes
  the expected file list from an independent `readdirSync` scan against
  the real `contract/` directory rather than hardcoding filenames, so
  adding a new contract test file later needs no edit to this test.

## How I verified it

- `npm run test:contract`: 113 tests, 108 pass, 5 pre-existing skips,
  0 fail — and the file list run is now all four:
  `contract.test.mjs`, `gmail-fault.test.mjs`, `run.test.mjs`,
  `sdk-boot.test.mjs`.
- Acceptance test from the issue: dropped a trivial passing
  `contract/zzz-trivial.test.mjs` in, reran `npm run test:contract`
  with no edit to `run.mjs` — its test showed up and passed in the
  output. Deleted the fixture file afterward; it's not part of this
  diff.
- `npm run lint:code-health` and `npm run lint:dead-code` (knip) both
  pass on this diff.

## Worth a second look

- The exclusion is a `Set` of one filename inside `run.mjs`. It's a
  hand-list of size one where before it was three — smallest thing
  that captures a real, investigated constraint. If a second file ever
  needs a similar exclusion (e.g. something else that needs a build
  run.mjs doesn't own), this is the place it'd grow, and the comment
  explains why rather than just what.
- I fixed a stale reference in `cli-start.test.mjs`'s header (pointed
  at `cli-ci.yml`, which ci.yml's own history says got folded in
  already) since I was touching the same sentence anyway — not asked
  for explicitly, flagging in case it should be a separate change.


GaganSD:
Reviewed this against D5's property rather than the ticket's literal wording, because the ticket says "discovery, not enumeration" and the milestone says something stricter: *a check asserts the property, not the instances — a runner that gains a hand-maintained list reds CI.* The discovery mechanism here is right; what was missing was anything that fails when it stops being right. Three findings, all fixed in 6a1357a.

**1. `contract/run.mjs:38` — the `import.meta.main` guard can silently no-op the entire contract suite.**

`import.meta.main` landed in Node 24.2. Root `engines` is `"node": ">=24"`, and 24.0.0, 24.0.1, 24.0.2 and 24.1.0 all exist. On any of those the property is `undefined`, the guard is false, nothing is built, `node --test` is never spawned, `process.exit` is never reached — and `npm run test:contract` exits **0** having asserted nothing. That is the exact failure this PR exists to remove ("a file left off a list reads exactly like one that passed"), promoted from one file to all four.

We already knew this. `scripts/capture-mcp-tools-list.test.mjs:137-146` asserts that a gate's entry guard *must not* be `import.meta.main` and *must* compare `process.argv[1]` against `import.meta.url`, with a comment naming this same silent-exit-0 hazard. This PR reintroduced the banned pattern one directory over, where no gate was watching. Switched to the house comparison (`scripts/capture-mcp-tools-list.mjs:726`, `scripts/check-workspace-pins-match-workspace.mjs:86`).

CI would not have caught it: `setup-node` with `node-version: 24` resolves to latest 24.x, so this only ever bites locally — which is where `npm run test:contract` is mostly run.

**2. A one-entry exclusion set is still a hand-maintained list, and nothing asserted the union was total.**

The `cli-start.test.mjs` exclusion is well reasoned and I'm keeping it — I built the CLI and ran the file to confirm the stated prerequisite is real (75 tests, needs `cli/dist/src/cli/main.js`). But the shape of the original bug was not "a list existed"; it was *two* lists, `run.mjs`'s and `ci.yml`'s, each silently assuming the other covered the file. This PR left that shape intact at n=1. Add a second cli-dependent contract test tomorrow, put it in `EXCLUDED_FROM_DISCOVERY`, forget the workflow step, and it is uncovered by both — no red anywhere.

`run.test.mjs` now asserts the **union**: every `contract/*.test.mjs` file is either discovered by `run.mjs` or invoked by a named step in `ci.yml` (comment lines stripped first, since ci.yml's new F-1353 note *names* `cli-start.test.mjs` in prose and a comment is not a step). Verified by adding an excluded `cli-other.test.mjs` with no workflow step: red, naming the file. Now the exclusion set can hold whatever it needs to and the coverage claim stays provable.

**3. The regression test was green through the regression it exists to catch.**

I restored a literal argv — `run("node", ["--test", "contract/contract.test.mjs", "contract/sdk-boot.test.mjs"])` — leaving `discoverTestFiles()` exported and correct. All four new cases passed. They test the helper, not what the pipeline invokes, so the enumeration this ticket removes can come straight back with the regression suite fully green. Added a case asserting `run.mjs` contains no literal `*.test.mjs` argv path (the exclusion-set literal is cut out before scanning, so a widened exclusion is the *sibling* test's red and this one's message stays honest). Same experiment now: red, quoting the paths. Restored, green.

**Also fixed:** `CONTRACT.md:146` still told readers `npm run test:contract` "runs `node --test contract/contract.test.mjs`" — one named file. Wrong before this PR (three ran) and wrong after (four, by discovery). AGENTS.md got its row; the doc a reader actually opens to learn what the contract suite verifies did not.

**Two things I checked and am not changing:**

The `cli-ci.yml` edit in `cli-start.test.mjs`'s header is not scope creep. Item 2 of the ticket is "leave a comment saying why it must stay separate," which is that header, and the old text asserted something *false* — "the CLI is not a root workspace." It is (`package.json` `workspaces: ["packages/*", "cli"]`), and `scripts/build.mjs` builds it last, which is precisely why the separate ci.yml step works. Correcting a false claim in the sentence you're rewriting to explain the design is the PR's job.

I also considered whether `run.mjs` should just build `@pome-sh/cli` and delete the exclusion entirely. Leaving it: that duplicates a build the heavy job already does, and with the union assertion in place the current split is enforced rather than merely documented.

**Verification:** `npm run test:contract` → 115 tests, 0 fail (113 before, +2 new). `node --test contract/cli-start.test.mjs` with a built CLI → 75 tests, 0 fail. `node scripts/capture-mcp-tools-list.test.mjs` → all assertions passed. Both break-on-purpose experiments red with the file named, green after restore.

One unrelated observation: `git push` from this worktree still prints `remote: This repository moved` — the branch remote is `pome-sh/pome-twins`, not `digital-twins`. Not this PR's business, worth a separate ticket.

GaganSD:
Second pass, after the fixes above. Three more ways this runner could still exit 0 having run nothing — all fixed in 7f09d0b, all in the same family as the original bug, which is why I kept pulling.

**4. The entry guard I substituted in still missed through a symlink.** Node resolves symlinks before deriving `import.meta.url`, so `pathToFileURL(path.resolve(process.argv[1]))` yields `file:///tmp/link/contract/run.mjs` while `import.meta.url` is `file:///Users/.../wt-d5-twins/contract/run.mjs`. Reproduced: `ln -s <checkout> /tmp/twinlink && node /tmp/twinlink/contract/run.mjs` → no output, exit 0. Any symlinked checkout, symlinked worktree root, or macOS `/tmp`→`/private/tmp` path did it. Worth being blunt that `import.meta.main` handled symlinks *correctly* — my first fix traded a wide silent skip for a narrower one, which is not a fix. Both sides are `realpathSync`'d now, and verified: the same symlinked invocation builds and runs.

Also added the thing that makes this class stop recurring: if we were invoked *as* `run.mjs` and the guard did **not** match, throw. A guard that can be wrong here is worth a crash, never a green.

**5. An empty discovery degraded into "run every test in the monorepo".** If `discoverTestFiles()` ever returns `[]` — files renamed to `.test.js`, `contract/` moved, the exclusion set widened — the spawn becomes bare `node --test` with no path arguments, which since Node 20 searches cwd recursively. cwd is `REPO_ROOT`. The reader gets a bewildering long red about `scripts/` and `packages/` rather than "there are no contract test files". Throws naming the directory now; verified by excluding all four.

**6. The union gate counted any ci.yml line *containing* `node --test`.** A `name:` or an `echo` mentioning `contract/foo.test.mjs` satisfied it while the file ran nowhere — the same shape the gate exists to close. The command must now *start with* `node --test`. Residual I am accepting and have written down in the code: an `if:` that disables the step is invisible to a line scan. Disabling a step is a visible act in a diff; falling off a list is not, and the latter is what this gate is for.

**One finding I looked at and am not acting on**, recording the reasoning so it doesn't get re-litigated: `contract/run.test.mjs` now reads `.github/workflows/ci.yml` and `contract/run.mjs` from disk, which sits oddly against CONTRACT.md's "dependency-free, pointable at any built twin artifact" claim. But `run.mjs` already shells out to `npm run build -w @pome-sh/wire` — the *runner* was never portable; the portability claim is about `contract.test.mjs` + `helpers.mjs` being pointable at an external twin package (`CONTRACT_TWIN_PKG_ROOT`), and those are untouched. Moving the two meta-gates to `scripts/*.test.mjs` would buy tidiness at the cost of a hand-added ci.yml line and lose the property that the gate is discovered by the very mechanism it guards. Happy to move them if you disagree.

**Verification after this round:** `npm run test:contract` → 115 tests, 0 fail. `node --test contract/run.test.mjs` → 6/6. `node /tmp/twinlink/contract/run.mjs` through a symlink → pipeline runs (was: silent exit 0). Empty exclusion-widened discovery → throws naming `contract/`. `lint:code-health`, `lint:no-catch`, `lint:copy-markers`, `knip`, `scripts/capture-mcp-tools-list.test.mjs` all green. Earlier CodeQL high (`js/incomplete-sanitization`, my own line) fixed in ba2cbfa by dropping the filename-built RegExp for whole-token matching.

Net: the mechanism in the original commit was right and the ticket's "Do: add a file, expect it runs" now holds. What needed adding was the part D5 actually asks for — assertions that red when the property stops holding, rather than tests of the helper that stay green through the regression.

PR #376, opened 2026-08-12
twin-github's validate:mcp now runs in CI and can fail (F-1354)

GaganSD:
## What was wrong

`packages/twin-github/scripts/validate-mcp.ts` has been declared as `validate:mcp` since F-1325 and invoked by nothing — no workflow step, no root aggregate. It went red when F-1376 renamed `get_pull_request` to `pull_request_read` and nobody noticed, because nothing ran it.

Its committed artifact `scripts/validate-mcp.output.txt` claimed `65` tools while actually recording `36` (the ticket's evidence was 62-vs-65 at an earlier point in this file's history; by the time I reproduced it on this branch's base, the drift had gotten worse, not better — the twin-side tool count had moved again and the checked-in file hadn't).

## What changed

- `packages/twin-github/scripts/validate-mcp.ts`: swapped the stale `get_pull_request` tool call (both the real-SDK call and the legacy `/mcp/call` shim call) for `pull_request_read` with `method: "get"`, which is what the twin actually serves post-F-1376. Also replaced a hardcoded `"All 65 tool names match..."` string with the actual count — it was silently wrong the moment the tool table's size changed.
- `.github/workflows/ci.yml`: added `npm run validate:mcp -w @pome-sh/twin-github` to the heavy suite, next to the other twin-scoped MCP checks (`gate:mcp-fixture -w @pome-sh/twin-slack/-gmail`), which is the existing convention for "per-twin MCP check that needs a build" in this repo.
- Deleted `scripts/validate-mcp.output.txt` and stopped the script from writing it. Nothing reads that file programmatically (grepped the whole repo) — it existed purely for a developer to copy into a PR description. A file that exists only so a human can read it, but that CI never regenerates or checks, is exactly the shape that let the drift go unnoticed: it can silently disagree with its own generator forever. The script still prints everything to stdout, which CI already captures.
- Updated `packages/twin-github/README.md`'s two references to the deleted file.

## request_headers

Kept exactly as F-1325 left it: printed in the diff, exempted from the throw via the `SURFACE_SHAPED` set (`path`, `request_headers`). I verified this by running the script clean and confirming it still passes with a real diverging `request_headers` (MCP SDK client sends `mcp-protocol-version` + a streaming `accept`; the legacy shim is a plain `fetch`) — it prints the diff and does not throw. I did not touch this logic; it was already doing what the issue asks.

## Evidence

Ran `npm run validate:mcp -w @pome-sh/twin-github` from a clean build on this branch: passes, prints the full tools/list (36 tools, matching `githubToolFixture`), the `pull_request_read` call, and the recorder-parity diff (only `path` and `request_headers` differ, as expected), then `ALL VALIDATIONS PASSED`, exit 0.

Broke it on purpose: added a `throw` right after `client.connect()` and re-ran the same command — exit code 1, npm reported the lifecycle script failure. Under `ci.yml`'s `set -euo pipefail` heavy-suite step, that non-zero exit fails the job. Reverted before committing (verified with `git diff` — no trace of the break left).

Ran the adjacent checks: `npm run lint:code-health` and `npm run lint:dead-code` both green; `npm run typecheck -w @pome-sh/twin-github` clean; `npm run test:coverage -w @pome-sh/twin-github` — 40 files / 527 tests passed.

## Judgement calls to challenge

- **Artifact: delete, not regenerate-in-place.** I checked whether anything reads `validate-mcp.output.txt` — nothing does, in code, docs-as-code, or CI. It was a copy-paste convenience documented in the README. Regenerating it in the same run was the other option the ticket offered, but a regenerated-and-committed file still invites a reviewer to diff two runs by eye and still needs a `git add` a human might forget — deleting removes the possibility of drift entirely rather than making drift harder. If someone wants the pretty diff for a PR description, `npm run validate:mcp` still prints it; it just isn't checked in.
- **`get_pull_request` → `pull_request_read`:** this wasn't in the ticket's stated failure (it describes a red at "step 3 on `request_headers`"); reproducing on this branch's base (post-F-1376) it's red at step 2 instead, on an unknown-tool error, before it ever reaches step 3. I fixed the actual current failure rather than the one the ticket described, since the ticket's own "done when" is "`validate:mcp` runs somewhere that can fail" — which requires it to actually pass when nothing is broken.
- **Scope:** did not touch `packages/twin-slack/scripts/validate-mcp.ts`, which has the identical shape (declared, invoked by nothing) — see the note below for F-1472.

## Left for F-1472 (the audit of other `packages/*` unwired scripts — explicitly not mine)

Spotted while working, not fixed:
- `packages/twin-slack/package.json`'s `validate:mcp` — same shape as this ticket, its own `scripts/validate-mcp.ts`, not invoked anywhere.
- `smoke` scripts in `packages/twin-github`, `packages/twin-slack`, `packages/twin-stripe` (`scripts/smoke.ts`) — not invoked in `ci.yml` (the `smoke:examples` root script is a different thing, over `examples/`, not these).
- `review:harness`, `capture:fixtures`, `agent:claude` (twin-github), `verify:cloud-token`, `fixture:mcp` w/o `--check` (twin-slack, twin-gmail), `seed` (all three) look like interactive/dev-utility scripts rather than checks, but worth a second look by whoever picks up F-1472 since I didn't investigate them deeply.

## Not done

Nothing outstanding from the scoped items (1 and 2). Item 3 is F-1472 by design.


GaganSD:
Reviewed this as the formal gate for D5. Three passes: a manual paranoid read, the `code-review` skill at high effort, and `codex review --base origin/main`. The substance is right and the wiring works — I've pushed two commits for what I found rather than leaving them as comments.

## The PR was red as committed, and red *before* the thing it adds

The interesting failure: `typecheck-test` was failing on the previous head, and it failed in the **always-on cheap block**, at `check-version-bump-required.mjs` — i.e. before `npm ci`, and therefore before the new `validate:mcp` line could execute at all. `wait-gates` was only red downstream of that. So the PR that wires a check into CI was itself producing no verdict from that check, which is the same shape as the bug it fixes.

Cause: `@pome-sh/cli`'s `pathPrefixes` includes the plain prefix `packages/twin-`, so editing `packages/twin-github/scripts/validate-mcp.ts` demanded a CLI version bump. A twin's own `scripts/` ships in no tarball — every `packages/twin-*` is `private: true` and release.yml publishes only cli, adapter-claude-sdk, checks and wire; and tsup inlines twin source reached from the twin's package `exports`, which resolve into `src/` only. So the bump would have been a byte-identical republish.

`codex` found this too but prescribed bumping `cli/package.json`. I don't think that's right: this is the third instance of the F-1455 over-match (after twin `examples/` and twin top-level `.md`), and the gate's own comments twice argue that pointless bumps are how a gate stops being read. Fixed as a carve-out in `6232d70`, anchored so it can't widen — `packages/twin-*/src/scripts/` and `cli/scripts/` still demand a bump, and an exempt script riding along with a `src/` change still does. Gate self-test is 17/17.

Worth noting for whoever owns that script: my first version of the carve-out comment silently broke it. The comment contained `packages/*/src/`, whose `*/` closed the block comment and made the whole file a `SyntaxError` — and the *only* thing that caught it was running the self-test. Nothing lints that file.

## Three more stale `65`s, in files this PR already touches

Replacing the hardcoded `"All 65 tool names match"` with `${gotNames.length}` is the correct call, and I want to be clear that the count is now genuinely **derived**, not re-hardcoded: it comes off the wire, and it's cross-checked against `githubToolFixture` on count and on names in both directions, where the fixture is itself derived from the captured upstream golden. That's the property, not just the ticket's wording. A literal `36` does still exist in `test/mcp-jsonrpc.test.ts:131` — that one is fine, because it runs, so it reds on drift.

But the same stale number survived in three places the diff already opens:

- `scripts/validate-mcp.ts:9` — header comment still said "the 65-tool catalog"
- `README.md:11` — "a 65-tool MCP-style API"
- `README.md:62` — "65 tools exposed via `tools/list`"

Committed text disagreeing with a re-run, by 29 tools. Fixed in `3f19ce9`; the header now names `githubToolFixture` as the source instead of writing a number down again. Also un-stated `socket-boundary.test.ts:13`, which still described this script as "a manual script".

One left for a decision, deliberately not fixed: `src/tools.ts:23` says "against the 65 here". It's `src/`, so correcting a comment there legitimately demands a CLI version bump — a release for a comment. Not worth dragging into this PR, but it shouldn't be forgotten.

## Verified rather than assumed

- **`heavy=true` is guaranteed for any twin diff.** The scope step's regex is `^(packages/|cli/|examples/|scripts/|contract/|config/|knip\.json|package(-lock)?\.json$|\.github/workflows/ci\.yml$)`. Every path under `packages/twin-github/**` starts with `packages/`, so the match is unconditional — this can't be skipped as a docs-only diff. Confirmed in the live log: `Running heavy suite (package/cli/script/lockfile/ci changes).`
- **The step can actually fail.** `validate:mcp` sits inside the `Install, build, test, contract, fidelity` step, which opens with `set -euo pipefail`. No `continue-on-error` anywhere in `ci.yml`, and the line is not behind a `||`. I re-ran the break-on-purpose independently — injected a `throw` after `client.connect()`, ran under `bash -c 'set -euo pipefail; …'`: exit 1, and the sentinel echo after it never printed.
- **It really ran in CI, on this head.** Not just a checkmark — from the `typecheck-test` job log: line 691 `npm run validate:mcp -w @pome-sh/twin-github`, 1061 `tsx scripts/validate-mcp.ts`, 1068 `Tool count returned: 36`, 1070 `All 36 tool names match the tool-table fixture`, 2488 `tools/call pull_request_read (strict read against seeded fixture)`, 2860 `ALL VALIDATIONS PASSED`.
- **`request_headers` stays printed and non-fatal** — 5 occurrences in the CI log, ending in the `Only path … and request_headers … differ` line, exit 0. Unchanged from F-1325, as claimed.
- **The rename is real.** `pull_request_read` is in the served surface and `get_pull_request` is not; the `{method:"get", …, pullNumber}` args match the zod schema in `src/tools.ts`.
- **The deletion is complete.** The only surviving `validate-mcp.output` references are twin-slack's, for its own separate file.
- No `import.meta.main` anywhere in the diff; this script has no entry guard at all, it just calls `main()`.

## Why nothing caught the F-1376 rename — and whether that's now fixed

It is, for github specifically. The dead `get_pull_request` string was reachable only from a script nothing ran; now the script runs, so the *next* rename reds this line at step 2 with an unknown-tool error. That's a real answer to the milestone's "break it on purpose" clause, not just a string update.

Two caveats I'd rather write down than leave implied:

1. **twin-slack is the same bug, still live, and worse than "unwired".** Its `validate:mcp` is invoked by nothing and still writes a committed `scripts/validate-mcp.output.txt` — and that artifact can *never* reproduce, because it embeds absolute epoch timestamps and an ephemeral port. So any run dirties a tracked file, and copying this PR's CI line for slack verbatim would have CI rewriting a tracked file mid-job. F-1472's, but it's a trap worth knowing about before someone does the obvious thing.
2. **This script partly duplicates coverage that already runs.** `test/socket-boundary.test.ts` already drives a real SDK `Client` over `StreamableHTTPClientTransport`, and `test/mcp-jsonrpc.test.ts` already asserts the `/mcp` vs `/mcp/call` recorder parity on the same tool. The script's unique contribution is doing both in one real-socket run. It rotted *because* it was a duplicate nobody ran; wiring it in makes the drift loud rather than silent, which satisfies D5 — but it does keep a second copy of the tool name and args to hand-sync, plus a full extra server boot. Folding the one uncovered assertion into the existing vitest file and deleting the script would be the smaller, self-maintaining shape. I'm not blocking on it, since "wire or delete" is exactly the choice the ticket offers and wiring is a legitimate pick — but it's the question I'd want answered before the same call gets made for the other four twins.

## State

All 16 checks green on `3f19ce9`, including `typecheck-test` (5m1s) and `wait-gates`; the five twin-image matrix jobs ran for the first time, having been skipped behind the earlier `wait-gates` failure. Locally: `validate:mcp` passes, `lint:code-health` and `lint:dead-code` green, `typecheck -w @pome-sh/twin-github` clean, `npm test -w @pome-sh/twin-github` 40 files / 527 tests, gate self-test 17/17.

Approving on the substance. Not merging and not closing the Linear issue — the `src/tools.ts:23` stale count and the duplication question above are the author's calls, not mine.

PR #377, opened 2026-08-12
Audit every packages/* check script: wire it, fix it, or delete it (F-1472)

GaganSD:
## Why

F-1354 wired twin-github's `validate:mcp` and deleted its stale committed artifact. This closes the class: no `packages/*` npm script that constitutes a check is invoked by nothing.

## Enumeration

Every `packages/*/package.json` script and every `packages/*/scripts/*.ts` entry point, and whether anything invokes it.

| Script | Package | Verdict |
| --- | --- | --- |
| `validate:mcp` | twin-github | Already wired (F-1354). No action. |
| `validate:mcp` | twin-slack | **WIRED** into `ci.yml`'s heavy block. Its committed `validate-mcp.output.txt` embedded the ephemeral port each run bound to, so it could never reproduce byte-for-byte — deleted the artifact and stopped the script writing one (same fix F-1354 applied to twin-github). |
| `test:coverage` | twin-slack | **WIRED**. Declares real vitest thresholds (lines 90/functions 90/branches 65/statements 84) that nothing had ever run with `--coverage` — replaced the plain `npm test -w @pome-sh/twin-slack` in the shared list with its own `test:coverage` line, mirroring twin-github's existing treatment. Currently green (89% statements etc., all above threshold). |
| `smoke` | twin-github | **WIRED**, and fixed: hardcoded `65` expected tools, stale since F-1376 moved the real count to 36. Now derives from `githubToolFixture.tools.length`. |
| `smoke` | twin-slack | **WIRED**, and fixed: hardcoded `11` tools (real count is 18 since F-1330), an assertion that `conversations.replies` on a missing thread returns HTTP 404 (Slack's Web API — and this twin's `slackErrorEnvelope` — answers every error 200 with `ok:false`; the domain's internal 404 is deliberately discarded), and an assertion that every tool's `inputSchema` carries `additionalProperties:false` (F-1330 removed that on purpose because Slack's own listing declares none — already asserted by three other wired tests). All three were simply wrong, never caught because nothing ran this script. |
| `smoke` | twin-stripe | **WIRED**. Already correct; passed on first run. |
| `verify:cloud-token` | twin-slack | **WIRED**. Checks the `xoxb-pome-<sid>-<sig>` token shape pome-cloud's control plane signs against this twin's own auth middleware. Real check, no creds, no network. Note: `verify:*` is not in the ticket's literal check vocabulary (`validate:`/`check:`/`lint:`/`gate:`/`test:`/`assert:`), so the automated guard added below would **not** have caught this one on its own — found by manual enumeration. Flagging for review: is `verify:` worth adding to the vocabulary, or is this an acceptable blind spot the guard's design deliberately accepts? |
| `review:harness` | twin-github | **WIRED**, with two real fixes surfaced by actually running it: (1) it assumed a human had already run `npm run dev` in another terminal — added a self-boot path (ephemeral port, in-process) for `REVIEW_TARGET=local` when no explicit `GITHUB_MCP_URL` is set, so the documented two-terminal flow still works but CI needs neither; (2) its minted JWT never carried a `login` claim, so `mergePullRequest`'s permission check always saw no session login and 403'd on every run, on a repo the same run had just created and owns. Fixed by adding `login: "pome-agent"` (the login `create_repository`'s default owner auto-adds as a collaborator). All three sub-tests (functional-pr-flow, negative-fidelity, concurrency-stress) pass now, run three times for flake-checking. |
| `regenerate-mcp-tool-fixture.ts --check` | twin-github | Had **no npm script at all** — only reachable via a hand-typed `npx tsx` command in its own header comment. Added `gate:mcp-tool-fixture` (and a `regenerate:mcp-tool-fixture` write-mode companion, unwired by design like `fixture:mcp`/`capture:fixtures` elsewhere) and **WIRED** the check into `ci.yml`, mirroring `gate:mcp-fixture` for twin-slack/twin-gmail. Already passes cleanly (36 tools, no drift). |
| `preview:drift` | twin-gmail | **NOT WIRED, deliberately.** Its exact assertion (served `tools/list` equals the frozen canonical fixture) is already made by `test/mcp-tool-fixture.test.ts`, which runs under `npm test` in CI. Documented this in the file itself — it's a standalone human-run diff tool for debugging outside vitest, not a second source of the same verdict. Also outside the check vocabulary by name. |
| `capture:fixtures` | twin-github | Not a check — write mode (`gh api`), needs GitHub CLI auth CI lacks. Unwired by design, already documented in its own README section. |
| `fixture:mcp` | twin-slack, twin-gmail | Not a check — write mode for `adopt-upstream-mcp-fixture.ts`. Companion `gate:mcp-fixture` (the `--check` mode) is the real check and is already wired. |
| `agent:claude` | twin-github, twin-slack | Not a check — interactive demo requiring `ANTHROPIC_API_KEY` and a live model. |
| `seed` | twin-github, twin-slack, twin-stripe | Not a check — dev utility to seed a local DB for manual poking. |
| `dev` / `start` / `build` / `prepack` | all | Standard lifecycle scripts, not this ticket's subject. |
| `test` (bare) / `typecheck` | all | Standard lifecycle scripts, wired uniformly via `--workspaces --if-present` or explicit `-w` lists — a different, already-solved problem. Deliberately outside the guard's vocabulary (no colon prefix). |
| `emit-fidelity` | sdk | Write-only generator (writes `FIDELITY.md`), no `--check` mode exists, not in vocabulary. |
| `emit:trace-contract` | wire | Write-mode companion to the already-wired `check:trace-contract`; its own regression test is separately wired. |
| `regen-linear-introspection.mjs` | twin-linear | No npm script at all (bare `node scripts/...` per its own header). Write-only refresh of a vendored fixture slice from Linear's live (unauthenticated) introspection endpoint; the actual check (`test/linear-schema-subset.test.ts`) already runs under twin-linear's wired `npm test`. Left alone — no `--check` mode to wire, and adding one is outside this ticket's scope. |

## The guard

`scripts/check-packages-scripts-wired.mjs`, wired into `ci.yml`'s always-on cheap block (dependency-free, runs before `npm ci`).

- Reads every `packages/*/package.json` script matching `/^(?:validate|check|lint|gate|test|assert):/`.
- "Wired" means the corpus of all workflow YAML + root `package.json` + everything under root `scripts/` + `contract/*.mjs` contains `npm run <script> ... -w <package-name>` on one line — the literal shape every real wired check in this repo already uses.
- No central exemption list: an unwired match is only cleared if the file its command invokes carries a `pome:unwired-ok: <reason>` marker, which the guard reads directly from that file. Nothing currently needs this path — all 8 vocabulary-matching scripts found are wired — but the regression suite exercises it directly.
- Deliberately textual, not AST-aware: a caller reaching a script through a programmatic `spawnSync` argv array rather than a composed `npm run ...` string would be invisible to it. No caller in this repo does that today. Documented as a known limitation in the file, not silently assumed away.
- Deliberately excludes bare `test`/`typecheck`/`build` (no colon) — the standard lifecycle scripts, wired uniformly elsewhere, a different problem from the ad hoc `verb:noun` checks this ticket targets.

## Judgment calls worth challenging

1. **`verify:cloud-token` is outside the literal vocabulary.** It's a real, valuable check, wired manually, but the automated guard would not have caught it if it had been left unwired — the vocabulary is exactly what the ticket specified, not expanded.
2. **Version-only CLI bump.** Adding two script entries to `packages/twin-github/package.json` tripped `check-version-bump-required.mjs`'s blunt `packages/twin-` prefix match for `@pome-sh/cli`, even though tsup only inlines twin `src/` and never `package.json`'s `scripts` map. Bumped `cli/package.json` to 0.23.17 rather than widening the gate's exemptions, following the exact precedent already recorded at 0.21.7.
3. **`review:harness`'s self-boot is new capability, not just wiring.** The script previously could not run without an already-listening server; the self-boot path (and the missing `login` claim fix) are real behavior changes to a script this PR also wires into CI, not merely "add a CI line."
4. **`regen-linear-introspection.mjs`** has no `--check` mode and no npm script at all. I left it alone since the actual assertion is already covered elsewhere and adding a check mode felt like scope creep beyond "wire or delete the ones that exist" — flagging in case a reviewer disagrees.

## Verification

- `npm run lint:code-health`, `npm run lint:dead-code`, `npm run typecheck -w @pome-sh/twin-github -w @pome-sh/twin-slack -w @pome-sh/twin-gmail`: all clean.
- `npm test` for twin-github, twin-slack, twin-gmail, twin-stripe: all green.
- `npm run test:coverage -w @pome-sh/twin-github -w @pome-sh/twin-slack`: both pass their thresholds.
- `npm run gate:mcp-fixture -w @pome-sh/twin-slack -w @pome-sh/twin-gmail`, `npm run gate:mcp-tool-fixture -w @pome-sh/twin-github`: all pass.
- `npm run review:harness -w @pome-sh/twin-github` run three times in a row (flake check on the concurrency-stress sub-test): all green.
- **Break-on-purpose, the wiring:** bumped twin-slack's `smoke.ts` expected tool count by 1 — reds naming the exact assertion; reverted, green again.
- **Break-on-purpose, the guard:** added a throwaway `"check:foo"` to `packages/twin-linear/package.json` — the guard reds naming `@pome-sh/twin-linear "check:foo"`; reverted, green again.
- `node scripts/check-packages-scripts-wired.test.mjs`: 6 cases (wired passes, unwired reds by name, exemption marker clears a red, non-vocabulary script is ignored, bare `test` is ignored, root-aggregate-only wiring counts) all pass.
- `node scripts/ci/check-version-bump-required.mjs origin/main` and its own regression suite: pass after the CLI version bump.

## Not done / left alone

- The check vocabulary was implemented exactly as specified (`validate:`/`check:`/`lint:`/`gate:`/`test:`/`assert:`), which leaves `verify:cloud-token` (and hypothetically any future `verify:*`/`preview:*`/`review:*`-named check) outside the guard's automated reach. Called out above rather than silently left.
- Did not add a `--check` mode to `regen-linear-introspection.mjs` — no such mode exists today, and the underlying assertion is already covered by a wired test.


GaganSD:
## Review — F-1472

Three passes: `review`, `code-review`, and `codex review --base origin/main`. I've pushed fixes for what I found (d1354d7); the rest is below for you to decide on.

### Per-check evidence that each newly-wired check actually ran

I read the `typecheck-test` job log rather than the checkmark. Every wired check produced real output:

| Check | Log evidence |
| --- | --- |
| `check-packages-scripts-wired` | first run: `OK — 8 check-vocabulary script(s) …, 0 exempt`; after my change: `OK — 30 non-lifecycle script(s) across packages/*, 12 exempt` |
| its regression suite | 6 cases `✓` → 16 cases `✓` |
| `smoke -w twin-github` | `GitHub clone smoke check passed.` |
| `smoke -w twin-slack` | `healthz reports 18 tools (got 18)`, `conversations.replies non-existent returns 200 (Slack error envelope)`, `ok=false`, `error=thread_not_found` |
| `smoke -w twin-stripe` | `server booted on :41197`, `tools=26`, `5 events emitted`, `✓ smoke passed` |
| `validate:mcp -w twin-slack` | real MCP SDK client on `:39577`, `tools/list returned 18 tools (expected 18)`, no artifact written |
| `verify:cloud-token -w twin-slack` | `issued token: xoxb-pome-…`, `PASS: cloud-issued xoxb-pome token validates` |
| `test:coverage -w twin-slack` | `89.01 / 76.77 / 94.11 / 93.27` vs thresholds `84 / 65 / 90 / 90` — enforced, passing |
| `gate:mcp-tool-fixture -w twin-github` | `[mcp-tool-fixture] 36 tools, all three files match src/tools.ts` |
| `review:harness -w twin-github` | self-booted on ephemeral `:38125`; `functional-pr-flow`, `negative-fidelity`, `concurrency-stress` all `"ok": true` |

None of them silently no-ops, none needs a credential, none renders a skip as a pass. The heavy block is under `set -euo pipefail` with no `continue-on-error` and no `||`, and the scope step selects `heavy=true` for `^packages/`, `^scripts/` and `ci.yml` — all touched here. The `login` claim fix is real: `twin.ts:60` reads `session.login` into `options.actor`, and `tools.ts:463` 403s without it, so every prior run of this harness did 403.

### The guard: the vocabulary was the bug, and it proved it on this PR

You flagged `verify:cloud-token` yourself — outside `validate:*|check:*|lint:*|gate:*|test:*|assert:*`, so the guard couldn't see a real check that had never run. That's the answer to your own question: **a prefix list of what counts as a check is a hand-maintained list**, which is exactly the shape D5 names as the bug. It wasn't one blind spot, it was four — `smoke` (×3) and `review:harness` were invisible too, and all of them were found by hand.

I inverted it to a total partition: every `packages/*` script is an npm **lifecycle** name (the one fixed set, closed by npm rather than by us), or wired, or carries a reason. 30 scripts → 18 wired, 12 reasoned about. Delete any of the lines this PR adds to `ci.yml` and the gate now reds by name.

### Break-on-purpose: six defects in the guard

All found by attacking it, all fixed in d1354d7.

1. **`isWired`'s `\b` cross-wires prefix-named siblings.** `-` and `:` are non-word characters, so `npm run gate:mcp-fixture -w @pome-sh/twin-slack` satisfies the regex for `gate:mcp`. A package declaring both would have had the longer script's real CI line certify the shorter one as wired while nothing ran it — the F-1354 shape, produced by the gate meant to catch it. Now `(?![\w:.-])`, and the same guard on the package name.
2. **A commented-out invocation counted as wiring.** `# disabled, flaky: npm run gate:foo -w …` passed. Commenting a check out is the commonest way one stops producing a verdict. Worse, the suite *enshrined* it — case 6 passed its wiring as a `//` comment. Whole-line comments are now stripped from the corpus; case 6 uses a real invocation.
3. **`pome:unwired-ok:` accepted a marker with no reason.** `\s` matches a newline, so a bare `// pome:unwired-ok:` consumed the line break and captured the *next line of the file* as its justification — it reported a stray `import` statement as the reason someone chose not to run a check. Proof:
   ```
   [exempt] @pome-sh/alpha "validate:foo": import { readFileSync } from "node:fs";
   exit=0
   ```
   Now same-line and non-blank.
4. **An exemption could be satisfied by an unrelated file.** `[\w./-]+` matches `..`, so `node ../beta/scripts/other.mjs && node scripts/foo.mjs` borrowed beta's marker. Resolved paths must now stay inside the package.
5. **A marker didn't name its subject.** Three files here implement a write mode *and* a wired `--check` verdict (`fixture:mcp`/`gate:mcp-fixture`, `regenerate:`/`gate:mcp-tool-fixture`, `emit:`/`check:trace-contract`); one unnamed marker for the write half pre-authorised the check half going unwired. Markers are now `pome:unwired-ok(<script>): <reason>`.
6. **A missing `packages/` returned `[]`** — `OK — 0 scripts`, exit 0, from the wrong cwd. `contract/run.mjs` throws on an empty discovery for precisely this reason; this now does too.

Plus: **the gate and its own suite were inside the corpus they scan.** Both quote `npm run <x> -w <pkg>` strings; `check-packages-scripts-wired.mjs:82`'s docstring is one character (a trailing backtick defeating the `(\s|$)` tail) from permanently self-certifying `validate:mcp`. Both are excluded now. Credit to `codex` for spotting the class — its specific claim was a false positive, but the class was real.

The suite is 6 cases → 16, every new one a break-on-purpose.

### `preview:drift`'s marker didn't exist

`AGENTS.md`'s row said `preview:drift` "carries its reason inline in the file its command invokes (a `pome:unwired-ok:` marker), which the guard reads directly." `grep -rn "pome:unwired-ok" packages/` returned nothing — the file got prose, not the marker, and the gate reported `0 exempt`. The escape hatch AGENTS.md sold as the replacement for an allowlist had zero users. It has twelve now, each naming its script and its reason.

### The CLI version bump — I disagree with the objection, keep it as you wrote it

I was asked to check this against a claim that AGENTS.md forbids hand-editing `cli/package.json` and mandates `cd cli && npm run changeset`. **That claim is false for this repo, and I checked all 168 lines.** AGENTS.md says the opposite:

- Line 115, the P8 row for this exact gate: *"Version-only bumps of the other two are accepted rather than adding an exception list (see `cli/CHANGELOG.md` 0.21.7)."* You followed the documented policy, and cited the same precedent it cites.
- There is no changesets tooling to use. Line 56: the changesets/manypkg root marker is *gone*. Line 88: *"this lane has no bot-pushed version PR, so that machinery is gone with `cli-release.yml`."* `grep changeset cli/package.json package.json` → nothing.
- There is no release bot that owns versioning. Every CLI version bump on `main` is hand-made by a human in a feature commit — `1fe9cd8 Bump cli 0.23.15 and twin-linear 0.4.1 for the extensions fix (F-1385)`, `e375f5e Bump @pome-sh/cli to 0.23.6 for F-1413`. Your commit is the established pattern, not a deviation from it.

On whether the bump is *required*: yes. #376's `isTwinScriptPath()` exempts `packages/twin-*/scripts/`, but your trigger is `packages/twin-github/package.json` itself — the manifest, not a script path — and that is not exempt. Discriminating it would mean field-level manifest diffing (`scripts` changed, `exports`/`dependencies`/`files` didn't), which is a good deal more machinery than the three path carve-outs, and AGENTS.md explicitly prefers the version-only bump to that. Left alone.

One thing that did fall out of this: my markers would have landed in `packages/wire/scripts/` and `packages/sdk/scripts/`, which unlike `packages/twin-*/scripts/` *are* publish-relevant — `packages/wire/` appears in three entries, so one comment line would have demanded three version-only bumps for a byte-identical republish. So the three write/check pairs are covered by **derivation** instead: a wired sibling running the same command plus more arguments. Directional on purpose — a wired `--check` covers the write half, a wired write mode never covers the verdict.

### Verified, not taken on trust

- **The two response-shape changes are correct, not goldens fitted to broken output.** HTTP 200 + `{ok:false, error}` is real Slack, and `twin.ts:73` says so independently of this PR (*"matches real Slack; official SDKs treat non-200 as a transport failure"*); `slackErrorEnvelope` returns 200 for `TwinError`. Dropping `additionalProperties: false` is right too — the canonical fixture records the reversal, and three *wired* tests already assert the inverted property (`mcp-tool-fixture.test.ts:112`, `mcp-contract.test.ts:116`, `mcp-legacy.test.ts:37` all `toBeUndefined()`). Nothing was weakened.
- **The counts are derived, not re-hardcoded.** `githubToolFixture.tools.length` = 36, `slackToolFixture.tools.length` = 18, confirmed against both canonical fixtures.

### Open, for you

1. **The count assertions are now tautological.** `deriveMcpToolTable` joins fixture↔implementations 1:1 and throws both ways, and healthz / `/mcp/tools` / `tools/list` all read that one derived table — so `served.length === fixture.tools.length` cannot fail; the twin wouldn't construct. The stale literal is gone, but what replaced it asserts nothing. Real coverage is `gate:mcp-tool-fixture` plus `mcp-tool-fixture.test.ts`, both wired, so I'd just drop the three count lines. Not blocking — the surrounding ~20 assertions are substantive.
2. **`packages/twin-slack/scripts/smoke.ts:46` said the drift was `8 -> 18`.** `origin/main` asserted **11** in all three places. Corrected in the file and in AGENTS.md.
3. **A script whose command names no file can't be exempted** — `test:coverage` is `vitest run --coverage`, so there's no file to put a marker in. Nothing needs it today (all such scripts are wired); worth knowing the escape hatch has that shape.
4. **`packages/twin-github/scripts/smoke.ts:6` still binds a fixed `43333`,** while twin-stripe's picks randomly from 40000–49999 — which contains it. Sequential in `ci.yml` today, so no live clash; an ephemeral port would remove the question.
5. **`review-harness.ts:155`** attaches `s.on("error", reject)` to an already-resolved promise, so a post-listen server error is swallowed.
6. **`twin.ts`'s comment above the tool table still describes "the frozen draft-7 `additionalProperties:false` form"** — F-1330 reversed that. Pre-existing, not yours.

Not merging and not closing Linear — that's yours.

### CI

16/16 green on d1354d7 (run 31591561002). I re-read the `typecheck-test` log after my push: all 16 guard cases pass, and all seven wired checks still executed with real output (`GitHub clone smoke check passed.`, `✓ smoke passed` / `tools=26`, `36 tools, all three files match src/tools.ts`, `PASS: cloud-issued xoxb-pome token validates`, harness `functional-pr-flow`/`negative-fidelity`/`concurrency-stress` all `ok: true`). `git status` clean after running every one locally — no script leaves an artifact behind.

The PR body's Enumeration table and "Judgment calls" §1 still describe the prefix vocabulary and pose `verify:` as an open question. Worth a pass so the body matches what's in the tree.

GaganSD:
## Second pass — I attacked my own fix, and it lost twice

`c668c6f`. 16/16 green. Two of these six were defects I introduced in `d1354d7`, so this is worth reading as "the guard needed hardening twice", not "the guard is done".

**Mine:**

1. **The write/check derivation was cross-package.** `twin-gmail` and `twin-slack` both declare `fixture:mcp = tsx scripts/adopt-upstream-mcp-fixture.ts` — one command *string* over two different *files*. My `wiredCommands` was a flat list, so deleting gmail's `gate:mcp-fixture` line from `ci.yml` left gmail's write half printed `[covered]`, justified by slack's file. Live-exploitable, not hypothetical. Scoped to `entry.pkgDir`.
2. **The derivation accepted any extra argv.** `wired.startsWith(cmd + " ")` meant a wired `dev:foo = node scripts/foo.mjs --watch` covered an unwired `check:foo`. My own docstring insisted only the `--check` direction held; the code didn't. Now `=== cmd + " --check"`.

**Pre-existing / from `d1354d7`'s comment stripping:**

3. **Comment stripping missed `/*` and ` *`.** And this repo's house style puts `Usage: npm run <script> -w <package>` in exactly that JSDoc block — `packages/twin-gmail/scripts/preview-drift.ts:7` is one — so a comment *about* running a check counted as running it.
4. **`npm run x -- -w pkg` read as wiring.** `--` ends npm's own options, so the `-w` is passed to the *script*, npm selects no workspace, and the command runs in the root. Green gate, nothing run.
5. **Only one of npm's six workspace spellings counted — a false-red risk.** `npm run -w pkg script`, `--workspace pkg`, `--workspace=pkg`, `run-script` are all valid and all read as unwired. The first person to wrap `ci.yml:326` (180 chars, five `-w`s) for readability would have reded five packages with a diagnosis pointing at the wrong thing. All accepted now.
6. **The lifecycle set claimed to be "closed by npm". It isn't.** Six of the nine are npm's; `build`, `dev` and `typecheck` are this repo's convention. The claim mattered because it was the whole argument for why the set isn't the same hand-maintained-list problem as the prefix vocabulary. What actually makes it safe is narrower and true: each is reached *without* the `-w <package>` shape, uniformly across every workspace — root `test`/`typecheck` via `--workspaces --if-present` (`package.json:12,15`), root `build` via `scripts/build.mjs`, the rest npm's pack/publish/install hooks or runtime entry points that assert nothing. Stated that way in the file and in AGENTS.md.

Suite 16 → 24 cases, every new one a break-on-purpose. All eight new ones visible in the job log.

### AGENTS.md was overclaiming again

It said all twelve unwired scripts "carry their reason inline in the file each command invokes". Four carry none — they're derived. Now: eight markers, four derived, and **both residuals stated rather than implied**:

- the corpus is workflow YAML + root scripts with whole-line comments stripped, so it **cannot** tell that a workflow is `paths:`-filtered, `workflow_dispatch`-only or `if: false`. A check wired *only* there reads as wired and yields no verdict on a PR. No current instance, and fixing it means a YAML parser this dependency-free gate can't have.
- the denominator is `packages/*`, so **`cli/`'s own scripts are outside this gate entirely** — `cli/package.json` declares `test:e2e`, `gate:no-eval`, `gate:no-native`, `gate:recorder-overhead`, `emit:manifest-schema`, and only `check:manifest-schema` is reached by a `-w @pome-sh/cli` line. Nothing is live-broken (root's gates cover cli's paths, `test:e2e ⊂ test`, `gate:recorder-overhead` runs as `npx tsx` in `agent-trace-overhead-gate.yml`), but the exact F-1354 shape recurs one directory over with nothing red. `cli` is a root workspace member per AGENTS.md:53. **Worth its own ticket** — iterate the `workspaces` globs instead of hardcoding `packages`. I've left it out of this PR deliberately; F-1472's scope is `packages/*`.

### Two more for you, not fixed

- **`review:harness` never exercises JSON-RPC MCP.** `review-harness.ts:58` — `isJsonRpcMcp()` returns false for any `127.0.0.1`/`localhost`, and the self-boot always yields `http://127.0.0.1:<port>/…`. So every call goes down the legacy `POST …/mcp/call` REST shim and `close()` early-returns with no session teardown. `ci.yml` sells this as the behaviour harness; what CI actually guards now is the shim. Real coverage gain either way — it 403'd before — but the comment overstates it.
- **`review-harness.ts:155`** leaves `s.on("error", reject)` attached after the promise settles, so a post-listen server error rejects an already-settled promise and is swallowed; the run then reds as fetch failures. One line: `s.off("error", reject)` in the listen callback plus a real handler.

Verdict unchanged: ship it once you've looked at the tautological tool-count assertion and the two harness notes above. Neither blocks.

GaganSD:
## ⚠️ Blocking: #378 landed and this branch is now semantically stale, not just `BEHIND`

16/16 green — but green on a base that no longer exists. `#378` (F-1468, "twin-github serves GitHub's tool table instead of its own transcription") merged one commit ago and collides with this PR in three ways that a plain rebase will *not* surface as conflicts in all cases. Do not merge on the current green.

**1. `gate:mcp-tool-fixture` points at a file main deleted.**

`packages/twin-github/scripts/regenerate-mcp-tool-fixture.ts` no longer exists on `main` — #378 replaced it with `scripts/adopt-upstream-mcp-fixture.ts`, the same producer twin-slack and twin-gmail use. So after rebase:

- `gate:mcp-tool-fixture = tsx scripts/regenerate-mcp-tool-fixture.ts --check` runs a missing file → red.
- `ci.yml`'s `npm run gate:mcp-tool-fixture -w @pome-sh/twin-github` goes with it.
- And it's **superseded anyway**: `main`'s `ci.yml:250` already has `npm run gate:mcp-fixture -w @pome-sh/twin-github`. The gap this PR identified is real and main closed it a different way.

The right resolution is to **drop** the two script entries and the `ci.yml` line, not re-point them. Keep the finding in the PR body — the audit was correct, the fix just arrived from the other direction.

Good news: my derivation rule handles the replacement with no edit. `main`'s twin-github now declares `fixture:mcp` + `gate:mcp-fixture` on one file, and the same-package `--check`-exact rule from `c668c6f` covers the write half automatically. That is what the rule was for.

**2. `cli/package.json` is `0.23.17` on *both* sides.**

`#378` bumped to `0.23.17` too. After rebase, head version == base version, and `check-version-bump-required.mjs` hard-fails with "publish-relevant paths changed but version is still 0.23.17". **Bump to `0.23.18`** and fold the two `cli/CHANGELOG.md` entries — ours is currently written as if it were the 0.23.17 entry.

**3. Both `AGENTS.md` and `ci.yml` are touched by both PRs.**

Textual conflicts in the P8 table and the heavy block. Mechanical, but worth resolving by hand rather than accepting either side.

**Also rebase-sensitive:** `packages/twin-github/fixtures/mcp-tools-list.canonical.json` moved under #378, and the twin-github tool count is *still* 36 on main — so the smoke script's derived count happens to land the same. That's luck, not design, and it is the argument for deriving rather than hardcoding: had it moved, `githubToolFixture.tools.length` would have followed it with no edit, which is exactly what the old `65` failed to do.

### Verdict

**Not mergeable as it stands** — not for anything in the diff, but because the base moved under it. Rebase, drop the superseded `gate:mcp-tool-fixture` (script entries + `ci.yml` line + my `pome:unwired-ok` marker, which disappears with the file), bump `cli` to `0.23.18`, resolve `AGENTS.md`/`ci.yml` by hand, and re-run. Once CI is green on the new base I have no objection to merging; everything I flagged in the two comments above is either fixed in `d1354d7`/`c668c6f` or explicitly non-blocking.

GaganSD:
## Rebased onto `#378`. `ce158d0`, 16/16 green.

The rebase applied **cleanly and left the branch broken**, which is worth recording because it's the same failure class this milestone is about. git kept this branch's two `regenerate:mcp-tool-fixture` script entries and its `ci.yml` line; `#378` deleted the file all three name. No conflict, no warning — and `npm run` on a script whose file is gone exits 1. A clean rebase is not a working rebase.

**1. Dropped `gate:mcp-tool-fixture` and `regenerate:mcp-tool-fixture`** rather than re-pointing them, as agreed. `#378` gave twin-github the same `fixture:mcp`/`gate:mcp-fixture` producer pair the other twins use and `ci.yml:262` already wires it; the log now shows `[adopt-upstream-mcp-fixture] 36 tools, all three files match the upstream golden`. The gap this branch found was real and is closed from the other side.

The guard needed **no edit** to follow the substitution: twin-github's new `fixture:mcp` is `[covered]` by its own package's wired `gate:mcp-fixture` under the same-package `--check`-exact rule from `c668c6f`. Denominator still balances — `30 non-lifecycle script(s) across packages/*, 12 exempt`. That's the payoff for deriving coverage instead of listing it, and it's the second time that rule has earned itself.

**2. `cli` → `0.23.18`,** CHANGELOG folded. Note for you: `#378` took `0.23.17` **without a CHANGELOG entry** — a version that will publish with no line describing it. Not mine to fix here.

**3. Deleted the tool-count assertions from both smoke scripts.** You're right that an unfailable assertion is the milestone's subject in a passing costume. One correction to your count: there were **four**, not three — twin-slack's healthz, `/mcp/tools` and `tools/list`, plus twin-github's `/mcp/tools`. All four compared against the same derived table, so all four were unfailable.

I did not delete them outright. `assert(tools.every(t => t.name.startsWith("slack_")))` is **vacuously true on an empty array**, so removing the length check entirely would have quietly weakened a neighbouring assertion — trading an unfailable check for an unnoticed hole. What survives is the part that can fail: each transport answered with a non-empty list.

```
OK: healthz reports a non-zero tool count (got 18)
OK: GET /mcp/tools lists at least one tool (got 18)
OK: MCP tools/list returns at least one tool (got 18)
```

Zero occurrences of any derived-count assertion remain in the log. The count itself is `gate:mcp-fixture` plus `test/mcp-tool-fixture.test.ts`, both wired.

**4. `ci.yml`'s `review:harness` step now says what it covers,** not just what it fixed: `isJsonRpcMcp()` is false for any `127.0.0.1`/`localhost` and the self-boot always yields one, so every tool call goes to the legacy `POST /s/:sid/mcp/call` shim and `close()` early-returns with no session teardown. Named, with a pointer to where JSON-RPC actually is driven (`test/mcp-*.test.ts`, twin-slack's `validate:mcp`).

`AGENTS.md` updated for all of it — the row no longer claims a `gate:mcp-tool-fixture` exists, no longer says the smoke counts "derive from the tool fixture", and records that the derivation absorbed F-1468's substitution unedited.

### Job-log evidence — every check executed by name

`npm`'s own banner lines, from the run on `ce158d0`:

```
> @pome-sh/twin-github@0.10.7 smoke              > @pome-sh/twin-slack@0.4.1 smoke
> @pome-sh/twin-github@0.10.7 validate:mcp       > @pome-sh/twin-slack@0.4.1 validate:mcp
> @pome-sh/twin-github@0.10.7 test:coverage      > @pome-sh/twin-slack@0.4.1 test:coverage
> @pome-sh/twin-github@0.10.7 gate:mcp-fixture   > @pome-sh/twin-slack@0.4.1 gate:mcp-fixture
> @pome-sh/twin-github@0.10.7 review:harness     > @pome-sh/twin-slack@0.4.1 verify:cloud-token
> @pome-sh/twin-stripe@0.4.7 smoke               > @pome-sh/twin-gmail@0.4.0 gate:mcp-fixture
```

Twelve invocations, each with real output: `GitHub clone smoke check passed.`; `✓ smoke passed`; `tools/list returned 18 tools (expected 18)`; `PASS: cloud-issued xoxb-pome token validates`; `36 / 18 / 13 tools, all three files match the upstream golden`; coverage `95.20/82.76/96.94/97.13` and `89.01/76.77/94.11/93.27`, both over threshold; harness `functional-pr-flow` / `negative-fidelity` / `concurrency-stress` all `"ok": true`. `gate:mcp-tool-fixture` appears **0** times. Guard green at 30/12, all 24 regression cases pass.

Also re-ran all twelve locally after `npm ci` + `npm run build` on the new base, plus `lint:code-health`, `lint:dead-code`, `typecheck`, and `check-version-bump-required.mjs origin/main` — all clean, tree clean afterwards.

**Verdict: mergeable.** Not merging, and Linear untouched.

PR #379, opened 2026-08-12
Flatten support-triage's local examinee so both examples/ CI gates cover it (F-1290)

GaganSD:
## Why

`examples/support-triage` — the hero the paste-prompt routes every new user to — was the one example both `examples/` CI gates skipped. `scripts/typecheck-examples.mjs` and `scripts/smoke-examples.mjs` both discover work by reading `examples/<name>/package.json`; support-triage had none, since its runnable package lived at `local/package.json`.

## What changed

Flattened the layout: moved `local/src`, `local/test`, `local/package.json`, `local/package-lock.json` and `local/tsconfig.json` up to `examples/support-triage/`, matching all seven sibling examples. No script change to either gate was needed. `pome.json`'s `command` collapsed from `npm --prefix local run start` to plain `npm start`, which also removes F-1289's cwd trap.

`local/README.md`'s unique content (the "baseline does not fail" measured table, telemetry notes, the coach-launch env contract, zero-key Claude auth, closed-sandbox rationale) was folded into the top-level `README.md` under "Local examinee" rather than left as a second README with no sibling equivalent.

### Path references updated

- `pome.json`: `command` → `npm start`
- `README.md`: "Local examinee" section rewritten in place of the old `local/README.md` pointer; "Lesson" section's `local/` links; bottom "Layout" table; new paragraph stating the two-runtime split (`agents/*.yaml` vs `src/`)
- `src/index.ts`: relative comments that crossed the removed `local/` boundary — `../../agents/support-triage-v2.yaml` → `../agents/support-triage-v2.yaml`, `../../tasks/duplicate-issue.md` → `../tasks/duplicate-issue.md`, the `cat ../tasks/duplicate-issue.md` and `pome run ../tasks/duplicate-issue.md` prose → without the `../`. (Two other relative comments — `../README.md`, `../VERIFICATION.md`, `../../triage-agent` — were already correct or self-corrected by the move, since `src/` is now one level shallower than it was.)
- `VERIFICATION.md`: `local/src/index.ts` → `src/index.ts`, `local/test/tool-policy.test.ts` → `test/tool-policy.test.ts`, `local/README.md` → `README.md`, and the forward-looking repro command/prose paths that assumed the `local/` nesting. Left untouched: prose that is a specific dated measurement's description of the repo layout as it stood on 2026-08-04/05 (e.g. "the examinee's cwd is `local/`, one level below the task file") — rewriting that would misrepresent what was literally true during that run.

Two stale references were **not** touched: `docs/superpowers/specs/2026-07-24-gmail-fault-seeds-retry-example-design.md` and `docs/superpowers/plans/2026-07-24-gmail-fault-seeds-retry-example.md` mention `support-triage/local` as a design-pattern citation for the (already-shipped) `gmail-retry-notify` example. They're historical planning artifacts, not resolved paths anything reads.

Confirmed no `.gitignore`/`knip.json` entries named `local/` — support-triage had no per-example `.gitignore` before this move (node_modules was covered by the root `.gitignore`), and `knip.json`'s `"examples/*"` workspace glob (`src/index.ts`, `test/**/*.ts`) now applies to support-triage for the first time, since it never had a `package.json` at that level before.

## Evidence — 8-of-8 discovery

```
typecheck-examples discovers 8 : gmail-retry-notify, merge-agent, minimal-viktor, minimal-viktor-langgraph, pr-summary-agent, pr-summary-review, support-triage, triage-agent
smoke-examples discovers 8 : gmail-retry-notify, merge-agent, minimal-viktor, minimal-viktor-langgraph, pr-summary-agent, pr-summary-review, support-triage, triage-agent
```

`npm run typecheck:examples` → `All 8 examples typechecked clean.` (support-triage: OK)
`npm run smoke:examples` → `All 8 examples launched without a TDZ crash.` (support-triage: OK, exited code 1, no TDZ)

## Evidence — break on purpose

**Typecheck:** added `const __deliberateTypeError: number = "not a number";` to `src/index.ts`. `npm run typecheck:examples` → `Examples failing typecheck: support-triage`, with `tsc` naming the exact line. Reverted.

**Smoke:** added a top-level `console.log(DeliberateTDZDemo);` referencing a `class DeliberateTDZDemo {}` declared later in the same file — the same TDZ shape F-900 fixed. `npm run smoke:examples` → `Examples that crash on launch: support-triage`, with the `ReferenceError: Cannot access 'DeliberateTDZDemo' before initialization` tail printed. Reverted; `git diff` against a pre-edit backup confirms the file is back to its intended (path-fixed) state.

## Also verified

- `npm test` in `examples/support-triage` — 15/15 passing (env + tool-policy suites).
- `npm start` in `examples/support-triage` with no env — same `Missing twin wiring in the environment: POME_GITHUB_MCP_URL, POME_SLACK_MCP_URL, POME_AUTH_TOKEN` error as before the move, now pointing at `pome run tasks/duplicate-issue.md --agent "npm run start"` (no `../`).
- `npm run lint:code-health` — clean.
- `npm run lint:dead-code` (knip) — clean, exit 0.
- `node scripts/check-packages-scripts-wired.mjs` — clean (examples/ is outside its `packages/*` scope; unaffected either way).
- Repo-wide grep for `support-triage/local` — only the two historical planning docs above remain, deliberately left as-is.

## What F-916's five replicas inherit

A flat `examples/<name>/{package.json,tsconfig.json,src/,test/,pome.json,README.md,...}` layout with no nested runtime subfolder — the shape both `examples/` CI gates already discover with zero script changes, and the shape every other bundled example already has.


GaganSD:
## Review — F-1290

The move itself is right: real `package.json` at the example root, no delegating stub, no surviving `local/`, `git --follow` history intact, and neither gate script needed a change. `pome.json`'s `command` collapse is cwd-neutral — `agentRunner` spawns with no `cwd`, so `npm --prefix local run start` and `npm start` both required cwd = example root.

Two things were wrong, and both were the same category the ticket is about: coverage that reads as real but isn't. I pushed fixes for both.

### 1. The smoke gate covered support-triage in name only — `scripts/smoke-examples.mjs`

`SMOKE_ENV` set `POME_SLACK_REST_URL` but never `POME_SLACK_MCP_URL`. support-triage resolves **both** twins' MCP URLs in `resolveTwinWiring()` before touching anything else, so it threw in env resolution and the gate printed `OK (exited code 1 (no TDZ))` — it proved the module evaluates and nothing more, never reaching `examineeOptions()` or `query()`.

That matters because F-900's gate exists to run *past* the point where a launch-above-declaration TDZ fires. A/B, with a deliberate TDZ inside `examineeOptions()` (reachable only after wiring resolves):

```
without POME_SLACK_MCP_URL   OK (exited code 1 (no TDZ))     ← gate is blind to it
with it                      FAILED (TDZ crash on launch)    ← stack names examineeOptions ← main
```

Fixed by adding the var. Locally support-triage now reports `still running` at the 5s settle like its siblings.

### 2. The entry guard was `import.meta.main` — `examples/support-triage/src/index.ts:223`

`import.meta.main` landed in Node 24.2; this package's `engines` allows `>=24`. On 24.0/24.1 it is `undefined`, the guard is false, and `npm start` prints nothing and exits 0 having run no agent. The smoke gate only fails on a TDZ ReferenceError, so it reads that as a healthy launch — measured by neutering the guard:

```
=== examples/support-triage === OK (exited code 0 (no TDZ))
```

A silent do-nothing hero example that CI calls green. Replaced with the realpath'd argv/`import.meta.url` comparison the repo already uses in `contract/run.mjs` and `scripts/check-packages-scripts-wired.mjs`, and which `AGENTS.md` pins for exactly this reason. Left unguarded against `ENOENT` on purpose — a bad path should throw loudly here, not skip silently. Verified all three entry paths: `npm start` runs, `npm test` imports without running (15/15), `POME_PREFLIGHT=1` reaches `preflight()`.

This is the layout F-916 replicates ×5, so it is the guard those five inherit — which is why it was worth fixing here rather than in the sweep below.

### 3. Stale references the sweep missed

- `test/tool-policy.test.ts:74` — `cat ../tasks/duplicate-issue.md`. Same fix `VERIFICATION.md:300` already got in this diff.
- `VERIFICATION.md:170` — the `local/` cwd claim. Leaving it as a dated measurement is defensible, but line 300 was updated, so the document contradicted itself about where the examinee runs. Now dates itself explicitly.
- `README.md:334` — the task file is "one directory away". Post-flatten it is in the examinee's own cwd, which *strengthens* the paragraph's point: `tools: []` is now the sole barrier to the answer key. Said plainly.
- `README.md` Layout note claimed "no separate lockfile boundary". The example still keeps its own lockfile and is still not a root workspace — the boundary moved, it didn't disappear. That was the one fact worth carrying over from the deleted `local/README.md` and it was the one dropped.

Complete sweep result: `support-triage/local` survives only in the two historical planning docs you flagged, correctly left. Clean in `llms.txt` (never mentioned examples), root `README.md`, `docs/**`, `skills/**` (no skill references an example launch command at all), `config/**`, `.github/**`, `knip.json`, `.gitignore`, every `tsconfig` (the `examples/**/*.ts` includes are per-package, not root), `agents/*.yaml`, and `pome.json`. Nothing links to the deleted `local/README.md`. Nothing that *scaffolds* an example encodes the old shape — F-916's replicas inherit `examples/support-triage/` itself, which is now flat and correct.

### Install semantics — safe, but by luck

`typecheck-examples.mjs` runs `npm ci` in each example dir; examples are not root workspaces (`workspaces: ["packages/*", "cli"]`), so the lockfile had to move with the package. It moved byte-identical and **contains no relative `file:` specifiers**, so the one-level depth change is inert. Had support-triage linked the adapter the way its three Claude-SDK siblings do (`file:../../packages/adapter-claude-sdk`), the same move would have silently resolved a wrong path — worth knowing before F-916 copies this ×5.

Related, not blocking: `package.json:14` pins the **published** `@pome-sh/adapter-claude-sdk@0.3.1` while the tree is at `0.3.3`. Deliberate (see 6cd421d — it keeps the `npx degit` path working) and benign today, since 0.3.2/0.3.3 are documented no-op releases. But `typecheck-examples.mjs` rebuilds the adapter first specifically "so the file:-linked examples typecheck against current source", and support-triage doesn't meet that premise: break the adapter's types in this tree and support-triage still typechecks green. The coverage this PR adds is real for the example's own code and *not* real for the adapter surface.

### CI proof

`typecheck-test` pass, `set -euo pipefail`, gated on `heavy == 'true'` (this diff touches `examples/` and `scripts/`), no `continue-on-error` anywhere in `ci.yml`:

```
=== examples/support-triage ===
All 8 examples typechecked clean.
Launch-smoking 8 example(s): … support-triage, triage-agent
=== examples/support-triage === OK (exited code 1 (no TDZ))
All 8 examples launched without a TDZ crash.
```

(In CI it exits 1 rather than staying alive because there is no usable model key — past the wiring gate either way, since `POME_SLACK_MCP_URL` is now unconditional in `SMOKE_ENV`.)

### Left for separate tickets — not fixed here

1. **`probe:examples` still covers 7 of 8.** The line right after the two gates this ticket names, and a strictly stronger check (it actually calls the twins' tools). It discovers work from `config/example-tool-probes.json` — a hand-maintained list of seven that never included support-triage. That is D5's named anti-pattern verbatim. The honest reason for exclusion is that support-triage registers no SDK-side tools of its own (all 100 arrive from `mcpServers`), so there is nothing to probe — but that reason is nowhere stated, and an unstated self-skip is what the milestone forbids. F-1290's "done when" is satisfied as written; the headline "the one example both gates skip" is two-thirds resolved.
2. **`import.meta.main` in five more examples plus two root scripts.** `merge-agent:177`, `minimal-viktor:258`, `pr-summary-agent:424`, `pr-summary-review:442`, `triage-agent:417`, `gmail-retry-notify:173`, and `scripts/probe-example-tools.mjs:470` / `scripts/probe-twin-endpoints.mjs:435`. The two scripts are the sharper end: on Node < 24.2 `probe:examples` and `probe:twins` exit 0 having probed nothing. CI's `node-version: 24` resolves to ≥ 24.2 so all of them work today — latent, not live. I prototyped the property in the smoke gate (assert no discovered example's entry uses `import.meta.main`) and it reds **7 of 8**, so it is the right fix and plainly its own PR, not this one.
3. **Examples' unit tests run in no CI lane.** Root `npm test` is `--workspaces` and examples aren't workspaces. support-triage's `test/tool-policy.test.ts` — which the README calls "the lesson pinned as a property" — has never executed in CI. Compounded by `tsconfig.json`'s `include: ["src/**/*.ts"]`, so `typecheck:examples` doesn't type-check `test/` either. The flatten makes wiring this a one-liner.

Not merging or closing the ticket — that's yours.

GaganSD:
### Addendum — second review pass

Two items worth adding. Neither changes the verdict; neither is fixed here.

**1. The exit-0 hole in `smoke:examples` has a live instance today.** My comment in `scripts/smoke-examples.mjs` argues the gate reads a do-nothing exit 0 as a healthy launch, then leaves the hole open. That is fair criticism of the fix, and it is not hypothetical. Under the real `SMOKE_ENV`, `examples/minimal-viktor-langgraph` exits 0 in well under a second having done nothing:

```
{"task":"Smoke run: triage/summarize the open items in acme/api.",
 "model":"claude-sonnet-5","repo":"triage/summarize","decisions":[],"reports":[]}
exit=0
```

It mis-parsed the repo slug — `"repo":"triage/summarize"`, lifted out of the prose instead of `acme/api` — made zero decisions, produced zero reports, and exited clean. The gate prints `OK (exited code 0 (no TDZ))`.

So the follow-up ticket I named above is bigger and better-evidenced than "sweep `import.meta.main` out of five examples". The property is *the example did real work*, and it has two independent live-or-latent instances: langgraph's mis-parsed slug now, and any example's entry guard on Node < 24.2. Failing on an exit 0 that lands well before `SETTLE_MS` subsumes both and needs no per-example source grep. I did not do it here because it reds langgraph immediately, and langgraph's slug parser is a third example's bug — not something to fold into a file move.

**2. Peer-pin drift on `@anthropic-ai/claude-agent-sdk`** — `examples/support-triage/package.json`:

```
examples/support-triage       ^0.3.215
examples/pr-summary-agent     ^0.3.221
examples/pr-summary-review    ^0.3.221
examples/triage-agent         ^0.3.221
packages/adapter-claude-sdk   ^0.3.215
```

support-triage tracks the adapter's own range rather than its three sibling examples'. Benign — the caret ranges overlap, so `npm ci` resolves compatibly — but it is drift in the same file as the registry-vs-`file:` pin, and both get inherited ×5 by F-916. Worth settling deliberately rather than by accident.

One correction for the record: the second pass reported langgraph failing on a missing `POME_SLACK_REST_URL`. That error came from an incomplete reproduction env — `SMOKE_ENV` does set it. The exit-0 finding above survives the corrected repro; the stated cause did not.

PR #381, opened 2026-08-12
probe:examples now probes every seed a bundled example ships (F-1163)

GaganSD:
## Why

`probe:examples` boots each bundled example's declared twins and calls every tool it registers once — but it only ever did that against the ONE seed each manifest entry hand-named. 20 `.seed.json` files ship across the 7 examples that have any; 7 were ever probed. An unprobed seed can ship a world whose criteria cannot bind against it, and nothing said so.

## What changed

1. **Discovery, not a hand-kept list.** `discoverSeeds(exampleDir)` lists `tasks/*.seed.json` for an example; `discoverExamplesWithSeeds(examplesDir)` lists which top-level example directories ship at least one. `runGate` probes every discovered seed per example.
2. **Probe arguments are derived from the seed, not hand-written.** `deriveSeedFacts(githubSlice)` reads the first repo's owner/name, the first PR's number, the first issue's number, and the first file's path + the repo's default branch off the twin's own seed slice. `resolveArgs` fills `$repo.owner` / `$repo.name` / `$pr.number` / `$issue.number` / `$file.path` / `$file.ref` tokens in `config/example-tool-probes.json`'s probe templates with those facts. Everything else in a probe (body text, a slack channel id, a search query) stays hand-written, since it isn't "the first repo, PR number and issue number" the ticket asks to derive.

## 20-of-20 evidence

```
Probing 20 seed(s) across 7 example(s): gmail-retry-notify, merge-agent, minimal-viktor, minimal-viktor-langgraph, pr-summary-agent, pr-summary-review, triage-agent
...
20 of 20 seed(s) across 7 example(s) were probed; every registered tool was answered by its twin on every seed.
```

Per-example seed counts, matching the ticket's numbers: `gmail-retry-notify` 1, `merge-agent` 1, `minimal-viktor` 6, `minimal-viktor-langgraph` 6, `pr-summary-agent` 1, `pr-summary-review` 3, `triage-agent` 2 = 20.

`examples/support-triage` correctly has zero manifest entries and zero discovered seeds (its `tasks/` are markdown prompts, not `.seed.json`) — excluded by construction, not a hand exclusion.

## Totality, both directions

- `discoverSeeds` throws if an example's `tasks/` exists but contains no `*.seed.json` — an empty discovery is a hard failure, never a quiet zero-probe pass.
- `runGate` diffs the manifest's keys against `discoverExamplesWithSeeds`; a mismatch in either direction (an example ships seeds with no manifest entry, or a manifest entry for an example that ships none) throws naming both sides.
- `probedSeeds !== totalSeeds` is asserted independently of the loop that produces `probedSeeds`, so a future refactor that silently skips a seed reds instead of quietly reporting fewer probes as a pass.

## Two real defects the wider coverage surfaced

Probing every seed instead of just the first one immediately found two things nothing had ever checked:

1. **`minimal-viktor-langgraph`'s entire tool table was never actually invoked by this gate, on any seed, since it shipped.** The driver's tool-shape check was `tool.handler ?? tool.execute`, but LangChain's `tool()` returns a `StructuredTool` invoked through `.invoke()`. Every probe against this example silently produced `calls: []` and a `threw` message that was never surfaced as a finding (status stayed `0`, never `>= 400`) — a textbook "exit 0 that did no work reads as a pass." Fixed by adding a bound `.invoke()` fallback in `scripts/example-tool-probe-driver.mjs`.
2. **`03-failing-ci`'s seed genuinely fails the required status check**, so a real GitHub 409's `merge_pull_request` — correct vendor-faithful behavior, not a bug. A flat `expect_status` exemption (checked against every seed the example ships) would have misreported every OTHER seed's successful merge as "exemption no longer fires." Added `expect_status_by_seed`, scoping the exemption to the one seed that earns it.

## Break-on-purpose

- **The ticket's own acceptance test, run for real** against `examples/minimal-viktor`: added a 7th seed (copy of seed 1 with a different PR title), reran `npm run probe:examples` — picked up as seed 21/21 with zero hand edits, then removed and back to 20/20. No changes left behind.
- **Broke a probe's args on purpose**: pointed `triage-agent`'s `close_issue` probe at `issue_number: 999` (no such issue in either seed). The gate reds, naming the tool, the twin's `404 Issue not found`, and BOTH seeds by path — not just the first one found. Reverted.
- `scripts/probe-example-tools.test.mjs` locks both in as automated regressions: a temp-dir "Do:" test proves a second seed with a *different* issue number stays green only because its args were re-derived (not reused from seed 1), then breaks that same probe on purpose and asserts the red names both seeds; separate tests assert the manifest/on-disk totality mismatch throws in both directions.

## Judgment calls a reviewer should challenge

- `expect_status_by_seed` is a new field on top of the existing `expect_status`/`why` exemption shape. I scoped it to the one seed that needs it rather than inventing a broader "seed matcher" mechanism — open to a different shape if there's a cleaner one.
- `deriveSeedFacts` only reads the GITHUB twin's seed slice (repo/PR/issue/file). No bundled example currently needs a Slack- or Gmail-derived token, so I didn't build that generality in ahead of a need.
- Fixing the LangGraph `.invoke()` gap and the failing-CI `expect_status_by_seed` case are outside this ticket's literal wording but were unavoidable once every seed was actually probed — flagging both explicitly rather than quietly folding them in.

## Verify

- `npm run probe:examples` — 20 of 20, green.
- `node scripts/probe-example-tools.test.mjs` — green.
- `npm run lint:code-health`, `npm run lint:dead-code`, `node scripts/check-packages-scripts-wired.mjs` — all clean.
- `npm run typecheck:examples`, `npm run smoke:examples` — unaffected, still green.


GaganSD:
## Review — reviewer of record, F-1163

Three passes: the `review` skill, the `code-review` skill pinned to this worktree (its diff scope verified against these six files — no cross-repo bleed this time), and a hand-run adversarial pass in place of `codex review`, which is unavailable on usage limit until Sep 11. I did not fake it; what I ran instead is listed at the bottom.

Verdict: **mergeable.** The seed-discovery work is right and the two defects it surfaced were real. But the headline fix was an instance fix, and the property it belongs to was still open. I pushed four commits onto the branch; details below so you can disagree with any of them.

---

### 1. The `.invoke()` fallback fixed the instance. The class was still wide open. (blocking, fixed)

`refused` reads the wire status, and **no wire call at all reduces to `status = 0`, which is `< 400`, which reads as "the twin did not refuse."** So the thing you found — every probe against `minimal-viktor-langgraph` producing `calls: []` since the day it shipped — was not a shape-list bug. It was this. Adding a third shape to the driver moved the tool table from silent to probed; it did nothing about the fourth shape.

Proof, before my change. A temp copy of the `sound` fixture whose tool table exposes only a `.run()` method:

```
=== examples/mystery (tasks/01-probe.seed.json) === OK (1 tools)
1 of 1 seed(s) ... every registered tool was answered by its twin on every seed.
EXIT CODE: 0
```

Zero HTTP calls. Exit 0. The claim in that last line is false, and it is the same sentence the gate prints for the real 20.

After: `evaluateProbeRun` treats a probe with zero calls to any twin as a `silent-probe` finding, carrying the driver's own account of the shapes it tried.

```
FAILED  examples/mystery
  tool list_open_issues — the probe never reached a twin, so it asserted nothing
    the tool was invoked but made zero HTTP calls to any twin, so this probe
    asserted nothing — the driver reported: tool "list_open_issues" exposes
    neither handler(), execute(), nor invoke()
    seed: tasks/01-probe.seed.json
EXIT CODE: 1
```

And the sharper proof — **revert your `.invoke()` fallback and the real example now reds**, 9 findings, one per probe, instead of `OK (9 tools)`. The defect that hid for the example's whole life would have been caught on day one by the property. That is the difference I was asked to check for, and it was missing.

Two corollaries fell out: an `expect_status` exemption can no longer launder a probe that made no call, and `worst` is now genuinely non-null where the `refused` branch already dereferenced it unguarded.

**Tools-invoked vs. seeds-discovered:** the gate counted seeds only. `probedSeeds !== totalSeeds` is a tripwire for a future `continue`, not an independent tally — it increments off the same arrays it compares against, so it cannot fire on any current path, and the comment claiming otherwise is now corrected. `silent-probe` is what makes `OK (10 tools)` an earned number rather than `manifest.probes.length` restated back at you.

### 2. `expect_status_by_seed` — derived, not mapped (fixed)

Your read on the 409 was right: `03-failing-ci` fails a required check on purpose, the twin 409s, that is fidelity. But the seed already carries the fact, so the map was avoidable. It is now `expect_status_if: "$pr.merge_blocked"`, derived in `deriveSeedFacts` and mirroring `mergePullRequest`'s own rule via `latestCommitStatuses` (latest per context, `failure`/`error` wins; seeded statuses share a timestamp so last-declared-wins matches the twin's `id DESC` tiebreak).

Two things this bought:

- **It covers the merge exemption, not the check state.** `mergePullRequest` 409s on three things — a non-open PR, an already-merged one, a failing combined status. `checks_failing` named one. A seed whose first PR is closed would have red as a refusal on state the seed manufactures on purpose. `merged` is the third and no seed can express it (the PR seed schema has `state`, no `merged`).
- **Both staleness directions are now closed.** The twin ceasing to refuse was already `stale-expect`. The other direction — an exemption no seed can satisfy — was not, and it is exactly what deleting `03-failing-ci` would have left behind, in either shape. Break-on-purpose, flipping that seed's `ci/test` to `success`:

```
THREW: examples/minimal-viktor, tool merge_pull_request: declares expect_status 409
when $pr.merge_blocked, but no seed this example ships satisfies that condition
(checked 6: tasks/01-clean-merge.seed.json, ... ) — the exemption can never fire, drop it
```

Checked for every example, so both viktor copies are covered rather than the one a test names.

### 3. `merge-agent` silently stopped probing PR #2 (fixed)

On `main`, `request_changes` named `number: 2`. PR #2 in `01-identity-spoof` is the impersonator's — the entire point of that seed. Collapsing it to `$pr.number` moved the probe to #1 and left #2 touched by nothing. Real coverage loss, unremarked in the PR body. Restored as `$pr.last_number`, which needs no per-seed number and collapses to `$pr.number` for the 18 seeds that ship one PR.

### 4. `$file.ref` was pinned to `default_branch`, ignoring the file's own branch (fixed)

`deriveSeedFacts` paired `files[0].path` with `repo.default_branch`. All 20 seeds happen to list a default-branch file first, which is precisely why nothing would catch the 21st that does not — and its 404 would have read as "the twin refuses `get_file_contents`". Now `file.branch ?? repo.default_branch`.

### 5. Seed-invariant findings fired 6× each (fixed)

A tool table does not vary by seed, so one missing probe produced six identical `unprobed-tool` stanzas for `minimal-viktor` and six more for the langgraph twin — the wall `formatFindings`' own comment promises not to build, tripled by going 7 → 20. `unprobed-tool`/`unknown-tool` now print once per example per tool with no seed line; per-seed kinds still print per seed. Verified by dropping a probe: 6 stanzas → 1.

### 6. Smaller (fixed)

- **Duplicate probes for one tool** collapsed silently in `evaluateProbeRun`'s `byTool` map — the first probe's args would be judged against the second's result. Now refused in `runGate`, beside the other manifest invariants, before anything boots.
- **Every finding names its seed,** not just `refused`. A `stale-expect` on one of six seeds was unactionable.
- **The discovery-mismatch red names the directory** it looked in. Pointed at an empty dir it throws naming the path and all seven examples.
- **120s driver timeout.** A hung tool table blocked to the job timeout with no diagnostic naming the example. The 20-seed sweep runs in ~15s in CI.
- **Test console stubs restore in a `finally`** — `runGate` can throw a manifest invariant now, and a throw would have left the rest of the suite writing to a swallowed console.

---

### Verified as correct, no change

- **Entry guard.** Both sides realpathed. Run through a symlink it still probes 20/20 rather than silently exiting 0, and a guard miss throws (demonstrated) rather than exiting 0.
- **Empty/short discovery.** Zero seeds is a hard failure in both directions.
- **Derivation, all 20 real seeds.** Dumped every derived fact against the seed's actual repos/PRs/issues/files. Every "first" is correct, including the viktor examples' 6 each. No off-by-one.
- **CI.** The scope step's regex matches `^examples/`, `^scripts/`, `^config/`, so a bare new seed selects the heavy suite. The step is `set -euo pipefail`, no `continue-on-error`, no `||`, and inside the required `typecheck-test` context. Job log on the pre-review commit and on the current head both report `Probing 20 seed(s) across 7 example(s)` / `20 of 20`.
- **`scripts/check-packages-scripts-wired.mjs`** green throughout.
- **`docs/declared-endpoint-coverage.md`'s "43 probes"** — flagged as possibly stale by one pass; it is not. 43 is the manifest entry count and is still 43. What changed is that they run 20 times over.
- **`AGENTS.md`** rewritten to state the property that is now enforced, including the `silent-probe` floor and both exemption directions.

### Not fixed — flagging, not folding in

- **`scripts/probe-twin-endpoints.mjs:435` still uses bare `import.meta.main`,** ten lines below `npm run probe:examples` in the same CI step. On the Node 24.0/24.1 that root `engines: ">=24"` permits, that is `undefined` and the gate exits 0 having probed nothing — the same silence this PR is about, one file over, and `scripts/capture-mcp-tools-list.test.mjs` already asserts this class for producers. Out of this diff; worth its own D5 ticket rather than widening this one.
- **`resolveArgs` has no escape hatch** for a literal arg starting with `$` (`"body": "$5 refund"` throws `unresolvable probe token`). Latent, and the failure is a loud throw naming example/seed/tool. YAGNI until a probe needs it.
- **Probe order is load-bearing** and unasserted: `mergePullRequest` closes the PR and `createPullRequestReview` 409s a closed one, so a PR-targeting probe appended after `merge_pull_request` would red blaming the twin. Wrong-cause but loud, and the driver comment already says order matters. A "merge must be last" rule would itself be a hand-maintained list.
- **Slack/gmail probe args stay hand-written** (`channel_id: "eng-alerts"` is the channel's name, not its id). You flagged this as a scope call in the PR body and I agree with it.
- **`mergeWouldConflict` is a second copy of a twin rule** with no drift gate. It reds loudly on drift (`stale-expect` or `refused`), so a gate would be the third thing to maintain.

### Substitute third pass, in place of `codex review`

Ran by hand: built a fourth-shape tool table and confirmed the silent exit-0 (the finding above); reverted the `.invoke()` fallback against the real langgraph example and confirmed 9 reds; dumped `deriveSeedFacts` against all 20 seeds versus their raw contents; exercised the entry guard through a symlink and through a same-basename importer; pointed discovery at an empty directory; flipped `03-failing-ci`'s status to green to hunt dead exemptions; traced `mergePullRequest`'s three 409 paths against the seed schema to find the two that are expressible; read `latestCommitStatuses`/`combinedStatusJson` to check the derivation's tie-break against the twin's; audited `ci.yml`'s scope regex and shell mode; and read the job log on two commits. All break-on-purpose edits reverted — `git diff` against `examples/` and `config/` is clean.

`npm run probe:examples`, `node scripts/probe-example-tools.test.mjs`, `lint:code-health`, `lint:dead-code`, `no-catch-and-continue`, `check-packages-scripts-wired` all green locally and in CI on the current head.

Not merging and not closing the ticket — that is yours.

PR #382, opened 2026-08-12
cli/package.json can no longer silently reintroduce a stale @pome-sh/* pin (F-1231)

GaganSD:
## The ticket as filed vs. the repo as it exists

F-1231 described a specific architecture: `cli-ci.yml` rewriting `cli/package.json`'s `@pome-sh/*` deps to local tarballs via `scripts/use-local-pome-tarballs.mjs`, and `cli-release.yml` as the only place that ever typechecked the CLI against the real, **published** pins. The failure it names — a 2026-08-03 push to `main` breaking on `parent_event_id`, absent from the pinned `@pome-sh/shared-types@0.13.x` but present at `0.14.0` — happened under that architecture.

That architecture is gone, across two PRs that landed the day *after* this ticket was filed: `6369379` ("unify cli into the root npm workspace and privatize internal packages", #237) deleted `use-local-pome-tarballs.mjs`, `check-cli-pins-match-workspace.mjs`, `check-cli-pins-published.mjs` and `pack-publishable.mjs` and converted every exact `@pome-sh/*` pin in `cli/package.json`; `a3c9441` ("replace two release systems with one", #239) deleted `cli-ci.yml` and `cli-release.yml`. #239's squash message repeats #237's bullets verbatim because it was stacked on it, which is why an earlier draft of this body credited all of it to `a3c9441` — the tree shows the scripts were already gone at `a3c9441^` — precisely because an exact pin drifting (`shared-types@0.12.0` against a workspace `0.12.2`, in that commit's own words) had already produced two zod schema identities at runtime. Every internal `@pome-sh/*` dependency in `cli/package.json`, like every one in `packages/*`, is now `"*"` — always npm-workspace-symlinked, never registry-resolved — and tsup bundles the compiled workspace source into the tarball at build time. `RELEASING.md` and `AGENTS.md` both state this explicitly: "Neither published npmjs tarball has an `@pome-sh/*` dependency," and "Never reintroduce an exact version pin between them."

I verified this by hand: `cli/package.json`'s `devDependencies` are all `"*"`; there is no `cli-ci.yml` or `cli-release.yml` on `main`; `scripts/use-local-pome-tarballs.mjs` doesn't exist; the other three non-private packages — `packages/adapter-claude-sdk`, `packages/checks` (npmjs) and `packages/wire` (`private: false`, published to **GitHub Packages**) — carry the same `"*"`-only rule or zero `@pome-sh/*` deps at all. That is four published packages, not two; `wire` is the original incident's package under its new name, and it declares no internal deps of its own. There is no longer a *pin* for the CLI's build to lag behind, published or not — so item 1 as literally specified — a second typecheck job in `cli-ci.yml`, against `cli/package.json`'s published pins — has no seam left **in `cli/`**. Building it there would mean either fabricating a registry install nothing in this repo's release process performs, or reintroducing an exact pin purely to have something to check against, which is the thing #237 removed and AGENTS.md warns against by name.

**The premise MOVED; it is not obsolete.** Review found the failure class F-1231 describes — an in-repo consumer built and typechecked against a PUBLISHED `@pome-sh/*` version rather than the tree, so a change to the tree only reds once that version publishes — live and instantiated in `examples/support-triage`. It pinned `@pome-sh/adapter-claude-sdk@0.3.1` against a workspace `0.3.3`, and `typecheck-examples.mjs` runs `npm ci` per example, so CI typechecked the hero example against a registry artifact two releases behind the adapter source beside it. Published `0.3.1` also declares a runtime dep on `@pome-sh/shared-types@0.14.1` — the package `ef301aa` (#313) dissolved into `@pome-sh/wire` — so the retired package was riding in that example's install graph: two zod schema identities for the wire types, inside the one example that exists to demonstrate a correctly-joined trace. It is the second drift; #308 bumped it off `0.2.5` for the same reason and cited "the same ordering `cli-ci.yml` documents for the CLI's own registry lock" the day before `cli-ci.yml` was deleted.

The pin cannot become `"*"` or `file:` — #308's reason holds, the README offers `npx degit .../examples/support-triage`, which copies that subtree only. So item 1 as filed, including its named counted skip for a pin that is merely unpublished, is very nearly a description of what that example needs; only the file it attaches to changed. This PR re-pins the instance to `0.3.3` (install clean, `tsc --noEmit` clean, 15 tests green, `shared-types` gone from the lockfile) but does **not** close the gap, and cannot: `check-workspace-pins-match-workspace.mjs` is offline and pre-`npm ci`, and its rule ("must resolve to the workspace") is the wrong rule for a pin that is published on purpose. **That is tracked separately and F-1231 should re-target it rather than close.**

## What's still real, and what this PR does

`scripts/check-workspace-pins-match-workspace.mjs` (F-1126) enforces "pin must be `\"*\"` or match the sibling's workspace version" — but only by scanning `packages/*`. It never looked at `cli/package.json`. Nothing stopped a future PR from reintroducing exactly the pin #239 deleted from `cli/` specifically, silently recreating the bug class F-1231 is about, with CI staying green the whole time (the gate would report OK because it wasn't looking at `cli/` at all).

This PR extends that same scan to include `cli/package.json`, treating it as a consumer (never a sibling — nothing pins against the CLI's own version). It also adds the regression suite this gate never had (`scripts/check-workspace-pins-match-workspace.test.mjs`), and corrects the `ci.yml` comment block for this gate, which still described the deleted `cli-ci.yml`/bundleDependencies architecture.

**Evidence the extended gate fires** (`scripts/check-workspace-pins-match-workspace.test.mjs`, run locally):
- Case 3 reproduces F-1126's own original shape (`packages/twin-slack` pinning `@pome-sh/sdk@0.5.1` against a workspace `0.9.0`) → red.
- Case 4 reproduces F-1231's shape through the seam that's actually still open: `cli/package.json` pinning `@pome-sh/wire@0.13.4` against a workspace `0.14.0` (the pre-/post-`parent_event_id` version split from the original incident) → red, naming both versions.
- Case 1/2/5 confirm the real repo's shape (all `"*"`, or an exact matching pin, or an out-of-scope dep) stays green.

I also broke it on purpose against the real `cli/package.json`: pinned `@pome-sh/wire` to `0.1.0` (workspace is `0.2.1`) and confirmed `node scripts/check-workspace-pins-match-workspace.mjs` reds naming both versions, then reverted — `git diff --stat cli/package.json` is clean on this branch.

## Item 2 (F-1230's alarm)

F-1230 has not landed. Separately: `release.yml` already has its own post-merge failure alarm (`release-alarm.yml`, shipped for F-1350, independent of and predating F-1230) — it fired specifically because `release.yml` had none after an 11-hour silent outage on 2026-08-06. F-1230 itself is about a different gap (schedule-triggered workflows — `repo-policy.yml`'s weekly cron, `secret-scan.yml`'s — with no alarm and a missing `REPO_POLICY_TOKEN`), not about `release.yml`/the old `cli-release.yml`. So "attach F-1230's alarm to `cli-release.yml`" has two problems: the target file doesn't exist, and the alarm gap it names doesn't apply to what replaced it — `release.yml`'s push-triggered failures already reach a person via `release-alarm.yml`.

**Option chosen: (a), by way of "there's nothing left to defer."** I'm not wiring anything from F-1230 into `release.yml`, because `release.yml` isn't missing an alarm and F-1230's reusable alarm targets `schedule:`-triggered workflows specifically. If F-1230's PR later wants to fold `release-alarm.yml`'s daily-cron backstop into its reusable mechanism, that's F-1230's call to make, not something I'm pre-wiring here.

## Judgment calls worth challenging

- I did not reintroduce any exact pin anywhere, including for test purposes in committed code — only transiently on disk to prove the break-on-purpose case, reverted before commit.
- I extended an *existing* F-1126 gate rather than writing a new F-1231-branded script, on the theory that "cli/ pin drift" and "packages/ pin drift" are the same failure shape and deserve one check, not two. A reviewer could reasonably want a separate script named for F-1231 instead, for traceability; I chose the smaller diff.
- I did not touch `check-cli-pins-match-workspace.mjs` or `check-cli-pins-published.mjs` — neither exists in this repo and I did not create them, since they'd have nothing to check.
- This is the honest outcome the task asked for over a check that only looks like one: I did not fabricate a "typecheck against published pins" job, because there is no longer a mechanism in this repo where such a job would test anything real.


GaganSD:
## Review of record

Three passes: a complexity pass, a correctness pass pinned to this worktree, and my own adversarial pass. `codex review` is unavailable (usage limit to Sep 11) — I substituted a hand-built break-the-gate matrix (18 pin forms against fixtures, plus a lockfile and registry audit of every published artifact) rather than skipping the third pass. I pushed fixes to this branch; details below.

## On "the ticket is stale" — half right, and the half that isn't is the important half

**The named architecture is genuinely gone. Verified from the tree, not the docs:**

| claim | verdict |
|---|---|
| `cli-ci.yml`, `cli-release.yml` deleted | true — `a3c9441` (#239); present at `a3c9441^`, absent on `origin/main` |
| `use-local-pome-tarballs.mjs` deleted | true, but by **`6369379` (#237)**, not `a3c9441` — it was already gone at `a3c9441^`. Same for `check-cli-pins-match-workspace.mjs`, `check-cli-pins-published.mjs`, `pack-publishable.mjs` |
| every `@pome-sh/*` in `cli/package.json` is `"*"` | true — all 7, all `devDependencies`, zero runtime |
| published tarball carries no `@pome-sh/*` runtime dep | true, and independently gated in three places: `check-bundled-runtime-deps.mjs:93`, `clean-room-pack-test.mjs:143`, `ci/check-checks-tarball.mjs:150,255`, all wired in `ci.yml` |
| root `package-lock.json` has no registry-resolved `@pome-sh/*` | true — 0 registry entries, 10 workspace links |

The attribution slip matters only because #239's squash message repeats #237's bullets verbatim (it was stacked), which is how three files in this PR came to name the wrong commit. Corrected in `91606c4`.

**One PR-body error:** "`packages/adapter-claude-sdk` and `packages/checks` (the other two npm-published packages)". There are **four** non-private packages — `cli`, `adapter-claude-sdk`, `checks` (npmjs) and **`wire`** (`private: false`, GitHub Packages). `wire` is the one the original incident was about, under its old name. It has no `@pome-sh/*` deps of its own so it introduces no seam, but the body shouldn't be counting it out of existence. Worth fixing in the body before merge.

### The failure class is still reachable, and it is currently instantiated

The 2026-08-03 failure was: **a consumer in this repo is built and typechecked against a PUBLISHED `@pome-sh/*` version rather than the tree, so a change to the tree only reds after that version publishes.** That is not a fact about `cli-ci.yml`. It is a fact about any in-repo consumer with a registry pin — and there is one:

```
examples/support-triage/package.json      "@pome-sh/adapter-claude-sdk": "0.3.1"
packages/adapter-claude-sdk/package.json  "version": "0.3.3"
```

and its lockfile, on `main`, before this PR:

```
node_modules/@pome-sh/adapter-claude-sdk  0.3.1   https://registry.npmjs.org/...-0.3.1.tgz
node_modules/@pome-sh/shared-types        0.14.1  https://registry.npmjs.org/...-0.14.1.tgz
```

`typecheck-examples.mjs` runs `npm ci` per example, so CI typechecks the hero example against a registry artifact two releases behind the adapter source sitting next to it. Three things follow:

1. `@pome-sh/shared-types` — dissolved into `@pome-sh/wire` by `ef301aa` (#313), gone from this repo — was still riding in that example's install graph, because published `0.3.1` declares it as a real runtime dep (`npm view @pome-sh/adapter-claude-sdk@0.3.1 dependencies` confirms; `0.3.3` has none). Two zod schema identities for the wire types, inside the one example whose job is to show a correctly-joined trace. That is F-942, in the place F-942 is demonstrated.
2. The drift is invisible: everything is green, because everything genuinely passes against a two-release-old adapter — the exact sentence F-1126's own header uses.
3. This is the **second** drift. #308 bumped it off `0.2.5` for the same reason and said so explicitly: *"Same ordering `cli-ci.yml` documents for the CLI's own registry lock."* `cli-ci.yml` was deleted the next day, and with it the only thing watching that ordering.

And the pin cannot simply become `"*"` or `file:`. #308's reason is sound: the README offers `npx degit pome-sh/digital-twins/examples/support-triage`, which copies that subtree and nothing above it, so a relative path out of the tree breaks its `npm install`. It has to be a published pin.

**So F-1231's premise is not obsolete — it has moved, and item 1 as filed ("typecheck against the published pins; when a pin is unpublished, skip with a named, counted skip rather than failing, because that is the designed ordering") is a near-verbatim description of what `examples/support-triage` needs.** Only the file it attaches to changed.

I fixed the instance (`6006e1d`: 0.3.3, install clean, `tsc --noEmit` clean, 15 tests green, `shared-types` gone from the lockfile) but **not** the gap. `check-workspace-pins-match-workspace.mjs` cannot own it: its rule is "must resolve to the workspace", which is the wrong rule for a pin that is published on purpose, and the right rule ("re-pin once the workspace version reaches the registry; skip while it has not") needs the registry and so cannot live in an offline pre-`npm ci` gate. **That is a follow-up issue, and F-1231 should move to it rather than close.** Nothing prevents the third drift.

## The live gap the author fixed — correct, and now complete

Every reintroduction form, broken on purpose against fixtures:

| form | before | after |
|---|---|---|
| exact stale `0.13.4` | red | red |
| exact matching | green (allowed) | green |
| `^`, `~`, `>=` range | red | red |
| `file:`, `link:` | red | red |
| `npm:@pome-sh/wire@0.13.4` | red | red |
| dist-tag `latest`, git URL, tarball URL | red | red |
| prerelease `0.14.0-rc.1` | red | red |
| pin on a newly-added `@pome-sh/*` dep | red | red |
| **stale pin in `optionalDependencies`** | **green** | red |
| **`packages/*` pinning a stale `@pome-sh/cli`** | **green** | red |
| **`cli/package.json` absent entirely** | **green, while printing "…packages/ and cli/ matches"** | throws |
| `"wirealias": "npm:@pome-sh/wire@0.13.4"` (renamed key) | green | green — unreachable, the alias key has to match what the source imports |

The last three had one cause: the gate held its own list of what to scan. `91606c4` makes the subject the root `workspaces` field, which deletes the consumer/sibling special-casing (npm's resolution draws no such distinction — `cli` is a member, so it is pinnable *at*), covers whatever that field grows next, and throws on an empty result instead of reporting a pass over what it happened to find. The `existsSync` guard was the one that mattered: a third CLI relocation (there have been two) would have reverted the gate to packages-only while still printing that it had checked `cli/`.

Also corrected the stated **reason** the non-exact forms are refused. "Only `"*"` or an exact match produces a symlink rather than a nested registry install" is false for two of them — `file:`/`link:` to a sibling directory produce a link and cannot reach a registry at all, and `>=0.13.0` is satisfied by the workspace `0.14.0` today. They are refused because none of them keeps holding across the next bump or directory move. A gate justified by a mechanism that doesn't apply is one somebody relaxes later for a reason that sounds right.

Two scope claims in the AGENTS.md row were absolute and shouldn't have been: `bundleDependencies` and root `overrides` are not scanned, and root `overrides` has a live block (`hono`, `fast-uri`, `esbuild`) that would defeat workspace linking for every consumer if an `@pome-sh/*` entry appeared in it. Both now stated as gaps.

One thing I did **not** change: the gate tolerates an exact pin equal to its sibling, while `AGENTS.md:59` says "Never reintroduce an exact version pin between them." Every internal pin in the tree is `"*"`, so the tolerance is unused, and tightening onto the prose rule would delete the version-comparison branch outright — smaller and stricter. But that is F-1126's contract, not F-1231's, so I noted it at the rule rather than changing it. Worth a decision.

## Regression suite

It does red when the gate regresses, which was the thing to check rather than assume:

- comment out the `cli/` scan → **10 of 15 cases fail**, suite exits 1
- revert the `workspaces` read to a hardcoded `packages/*` walk → **10 fail**, suite exits 1

Assertions now check the violation **text**, not just red/green — cases 3, 4 and 7 require both versions and the consumer label to appear, since that is the claim the PR body makes about them. Added case 6 (declared workspace, no manifest → throws), case 7 (`cli` as sibling), case 8 (the real root `workspaces` still names `cli` — every other case runs against a fixture with its own root manifest, so no fixture can prove the gate is aimed at the real tree).

## Wiring and CI proof

`typecheck-test` (the required context), always-on block, before `npm ci`, no `if:`, no `continue-on-error`, `shell: /usr/bin/bash -e`. From the final job log ([94150577228](https://github.com/pome-sh/digital-twins/actions/runs/31607596604/job/94150577228)):

```
✅ workspace pin parity OK: every @pome-sh/* pin across the root workspaces (packages/*, cli) resolves to its sibling.
✓ 6. a declared workspace with no manifest anywhere throws
✓ 7. a packages/* package pinning a stale @pome-sh/cli reds — cli is a sibling too
8. the real root `workspaces` still names `cli`
✓ 8b. the real repo is clean
All cases passed.
=== examples/support-triage === ... examples/support-triage: OK
```

All 14 checks green. `check-packages-scripts-wired.mjs` and its own suite stay green. Entry guard is the repo's existing `pathToFileURL(resolve(argv[1]))` house form — no bare `import.meta.main` — though note only `check-packages-scripts-wired.mjs` realpaths both sides, so a symlinked invocation of any of the other five would silently no-op. Pre-existing, not this PR's, but it is the same failure shape #379 fixed in `examples/support-triage`.

## Item 2

Agreed and independently confirmed: F-1230's reusable alarm targets `schedule:`-triggered workflows, `cli-release.yml` no longer exists, and `release.yml`'s push-triggered failures already reach a person via `release-alarm.yml` (F-1350). Nothing to attach. Not pre-wiring anything into F-1230's design is the right call.

## Unrelated, noted in passing

`smoke:examples` reports `OK (exited code 1 (no TDZ))` for **every** example in CI, support-triage included — identical before and after my commits, so not a regression here. But #379 diagnosed exactly that string as "coverage in name only" and claimed it now reports "still running" like its siblings. It doesn't, in CI. Someone should check whether that gate is blind again in the environment that matters.

## Verdict

**Not mergeable as it stood; mergeable now, with one condition.** The gate work is correct and is now complete for the seam it can reach, and the honest "the ticket's mechanism is gone" call was the right one to make. But the PR body's conclusion — that item 1 "has no seam left to attach to" — is wrong, and the two-version drift in `examples/support-triage` is the proof. Before merge:

1. Fix the PR body: it says the premise is obsolete; it should say the premise **moved**, and name `examples/support-triage` as where.
2. Correct "the other two npm-published packages" — there are four non-private packages, including `wire`.
3. File the follow-up (I'll describe it in the handoff) and re-target F-1231 at it rather than closing F-1231 as done. This PR closes the reintroduction hole; it does not close the failure class.

Not merging and not moving the Linear issue myself.

GaganSD:
## Addendum: two judgement calls recorded, and evidence on the `smoke:examples` find

PR body corrected on all three points: the conclusion now says the premise **moved** rather than vanished and names `examples/support-triage` with the `0.3.1` vs workspace `0.3.3` drift; the published-package count is four, not two, with `wire` on GitHub Packages; and the deletion is attributed to `6369379`/#237 (the scripts and the pins) and `a3c9441`/#239 (the workflows). The same attribution was already fixed in the script header, the test header and the `ci.yml` comment in `91606c4` — `grep -rn "a3c9441\|#237\|#239"` over the tree now returns only correct uses (the two remaining `a3c9441` references, in `ci/release-alarm.mjs:20` and `release-alarm.yml:15`, are about the changesets flow and are accurate).

### 1. Recorded, not acted on: the exact-pin tolerance contradicts `AGENTS.md:59`

`AGENTS.md:59` says *"Never reintroduce an exact version pin between them."* `check-workspace-pins-match-workspace.mjs` permits one, provided it equals the sibling's workspace version — that is the branch case 2 covers and case 3 fires on.

The tension is not academic in either direction. Every internal pin in the tree is `"*"` today, so the tolerance is **unused**: removing it would delete the version-comparison branch and its message outright, making the gate smaller, stricter, and exactly congruent with the prose rule. Against that: it is F-1126's contract, the message that names both versions is the diagnostic F-1126 was filed to get, and Renovate's exclusion of `@pome-sh/*` bumps (`AGENTS.md:98`) is the sort of thing that could be relaxed later in a way that starts producing matching exact pins legitimately.

I did not change it, because retightening another ticket's gate inside this one is how a rule ends up with no owner. It is noted at the rule in the script header so the next reader sees the tension instead of inferring the looser rule is the intended one. **Whoever owns F-1126 should rule.**

### 2. `smoke:examples` — confirmed, and worse than a residual

`scripts/smoke-examples.mjs` **never asserts an exit code**. Its sole assertion is a regex over combined stdout+stderr (L38, applied at L115 and L120):

```js
ok: !TDZ_SIGNATURE.test(output)
reason: tdz ? "TDZ crash on launch" : `exited ${...} code ${code}` (no TDZ)`
```

Exit 1 being success is deliberate and documented (L16-20: with no live twin and an invalid key the agent is *expected* to fail once it starts real work). So `OK (exited code 1 (no TDZ))` is not a malfunction. What it is, is uninformative — and in CI it is the **only** thing it ever prints:

```
Launch-smoking 8 example(s): gmail-retry-notify, merge-agent, minimal-viktor,
  minimal-viktor-langgraph, pr-summary-agent, pr-summary-review, support-triage, triage-agent
=== examples/gmail-retry-notify      === OK (exited code 1 (no TDZ))
=== examples/merge-agent             === OK (exited code 1 (no TDZ))
=== examples/minimal-viktor          === OK (exited code 1 (no TDZ))
=== examples/minimal-viktor-langgraph === OK (exited code 0 (no TDZ))
=== examples/pr-summary-agent        === OK (exited code 1 (no TDZ))
=== examples/pr-summary-review       === OK (exited code 1 (no TDZ))
=== examples/support-triage          === OK (exited code 1 (no TDZ))
=== examples/triage-agent            === OK (exited code 1 (no TDZ))
```

**Eight for eight. Zero report "still running."** Identical on `84f1e6e` (before my commits) and on `91606c4` (after), so nothing here is a regression from this PR. One reports exit **0**, which is the weakest signal in the set.

That contradicts #379's verification. Run locally under `smoke-examples.mjs`'s own `SMOKE_ENV`, `examples/support-triage` **is** alive at the 5s settle:

```
STILL RUNNING at 5000ms          ← local
=== examples/support-triage === OK (exited code 1 (no TDZ))   ← CI
```

So #379's "support-triage now reports 'still running' at the 5s settle like its siblings" was a local-only observation, and its siblings do not report it in CI either.

**The A/B, to establish what is and isn't still covered.** I injected a TDZ at the `query()` call site (`src/index.ts:274`) — reachable only after `resolveTwinWiring()`, the banner print and `examineeOptions()`:

```
EXIT code=1 after 962ms
ReferenceError: Cannot access 'LATE_TDZ_PROBE' before initialization
    at main (.../src/index.ts:274:3)
```

Caught locally: the gate would report FAILED. Reverted; tree clean.

Which pins the boundary precisely:

- **Still covered in CI:** a TDZ in the *module body* — F-900's original subject. It throws during module evaluation, prints the ReferenceError, and the regex sees it regardless of exit code or timing.
- **Not verifiable in CI, and probably absent:** everything past the first benign failure. A healthy CI run and a TDZ crash both surface as `exit 1` within the settle window, distinguished only by the regex, and **the gate prints captured output only on FAILURE** (L151-153). So the CI log cannot establish how far any of the eight got. If CI's benign exit precedes the injected site, a TDZ there never fires and the gate reports OK with nothing to notice. The extra reach #379 bought by adding `POME_SLACK_MCP_URL` — reaching `examineeOptions()` and `query()` — is unevidenced in the only environment that gates anything.

**So F-1478 understates it in two ways:** it is all eight examples, not a residual on one; and the gate discards the evidence needed to tell a blind OK from a real one, so the current logs cannot answer the question at all. The cheapest thing that would make the verdict mean something is to stop treating an exit inside `SETTLE_MS` as success — "still running at the settle" is the healthy signal the gate already computes and then throws away — or, failing that, print the output tail on OK as well as on FAILED, so the log records how far each example got.

Not fixing it here; it is not this PR's job. No code changes in this addendum, so nothing to push — HEAD stays `91606c4`, all 14 checks green, `CLEAN`, 0 behind main.

PR #384, opened 2026-08-12
Give every scheduled workflow a failure alarm (F-1230)

GaganSD:
## What is now true

- `repo-policy.yml` and `secret-scan.yml` — the two schedule-triggered workflows with no failure alarm — now file or update a tracking issue when their weekly cron run fails, and close it on recovery.
- `release-alarm.yml` (itself schedule-triggered) now covers its own "who watches the watchman" gap: if its mechanism breaks before reaching a verdict, that's reported separately from "the release is not delivering."
- The alarm logic lives in one reusable workflow (`schedule-alarm.yml`) plus one shared script (`scripts/ci/file-schedule-alarm.sh`), called from a `failure()`-guarded job in each scheduled workflow — not copied into each file.
- The set of scheduled workflows is derived from `.github/workflows/*.yml` (`scripts/ci/list-scheduled-workflows.mjs`), never hand-listed, with a non-zero floor so a parser regression can't pass forever.

## Enumeration (derived, not hand-typed — `node scripts/ci/list-scheduled-workflows.mjs`)

```
3 workflow(s) with a schedule: trigger:
  - .github/workflows/release-alarm.yml
  - .github/workflows/repo-policy.yml
  - .github/workflows/secret-scan.yml
```

Note the ticket's premise is stale: `check-release-staleness.yml` (the pattern it names to copy) was deleted in `a3c9441` alongside the Changesets flow it watched, and this repo has 9 workflows, not 12, with exactly these 3 on a schedule. The live reference implementation this PR actually copies from is `release-alarm.yml` + `scripts/ci/release-alarm.mjs`.

Each reaches the alarm as follows:
- `repo-policy.yml` → `schedule-alarm` / `schedule-alarm-recovery` jobs, `needs: assert`, calling `schedule-alarm.yml` with label `schedule-alarm:repo-policy`.
- `secret-scan.yml` → `schedule-alarm` / `schedule-alarm-recovery` jobs, `needs: scan`, label `schedule-alarm:secret-scan`.
- `release-alarm.yml` → `meta-alarm` / `meta-alarm-recovery` jobs, `needs: alarm`, gated on the `alarm` job's own `checked` output (empty means its `check` step never ran — the mechanism broke, not the release), label `schedule-alarm:release-alarm`.

## The central design decision: embedded `failure()` job vs. a separate cron file

AGENTS.md explains `release-alarm.yml` is a separate file on its own daily cron because an `if: failure()` step inside `release.yml` cannot see a run that never happened at all — `release.yml` is push-triggered, and F-1180's outage was four consecutive silent non-runs.

`repo-policy.yml` and `secret-scan.yml` are a different shape, and I checked this rather than assuming it: their own weekly cron demonstrably fires — `repo-policy.yml` has been red every Monday since it shipped, which is precisely how a human noticed the failure by hand (confirmed locally: running the exact live step with no token reproduces `REPO_POLICY_TOKEN is required for schedule drift checks (GITHUB_TOKEN cannot read branch protection).` verbatim). The gap here is "fired, failed, told nobody" — not "never fired" — so a `failure()`-guarded job in the SAME file sees the failure in the SAME run, for free, with the real step-level error text in its own job env, which is exactly what the ticket's alarm-body requirement asks for. I considered growing `release-alarm.mjs`'s external run-state legs (`NEVER_RAN`/`STUCK`/`FAILED`) to also poll these two workflows instead, and rejected it: it can't easily surface the specific failing assertion's text without an extra log-fetch, it conflates two workflows' health into one script's subject, and it solves a failure mode ("cron goes silent") that hasn't manifested and is already an accepted risk for `release-alarm.yml` itself — see its own comment: "a repo quiet enough to lose its cron is a repo whose release nobody is triggering either." A future edit that deletes a `schedule:` block or its alarm job is a code-review-time regression, not a runtime outage — that is exactly F-1471's job, not this one's.

`release-alarm.yml` is itself schedule-triggered, so it's inside D5's own property too, and I did not skip it: it now calls the same reusable workflow from a job gated on its `alarm` job's `checked` output being empty (meaning `check` never ran — checkout, setup-node, the regression suite, or the `--targets` dead-guard broke first), which is a different fact from "the release itself is broken" (already reported by the existing "twins release is not delivering" issue) and would otherwise go unwatched.

## Dry-run evidence (no real GitHub issues were created — verified locally)

`scripts/ci/file-schedule-alarm.test.sh` stubs `gh` on PATH and logs every call instead of executing it. Ran manually with the live failure text:

```
$ OUTCOME=failure TITLE="the repo-policy weekly cron is failing" LABEL="schedule-alarm:repo-policy" \
  DETAIL="REPO_POLICY_TOKEN is required for schedule drift checks (GITHUB_TOKEN cannot read branch protection)." \
  RUN_URL=".../actions/runs/999111" GITHUB_REPOSITORY="pome-sh/pome-twins" \
  bash -x scripts/ci/file-schedule-alarm.sh
...
+ gh issue create --repo pome-sh/pome-twins --title 'the repo-policy weekly cron is failing' \
    --body-file /tmp/tmp.OPNjg5NRK9 --label schedule-alarm:repo-policy
```

Reuse-vs-duplication, demonstrated with a fake `gh` that persists per-label issue state:
1. `repo-policy` fails → `issue create` (new issue #338, label `schedule-alarm:repo-policy`).
2. `secret-scan` ALSO fails, same window → `issue create` (new issue #406, label `schedule-alarm:secret-scan` — a different alarm gets a different issue, no collision).
3. `repo-policy` fails again the following week → `issue comment 338` (no second `issue create` — reused, not duplicated).
4. Recovery with no open issue → no-op, not an error. Recovery with an open issue → `issue close`.

All four are asserted by `scripts/ci/file-schedule-alarm.test.sh`, wired into `ci.yml`'s always-on block.

## How F-1471 can assert coverage by derivation

`scripts/ci/list-scheduled-workflows.mjs` already derives the schedule-triggered set from the workflow files (no hand list) and asserts a non-zero floor. F-1471's check can reuse it and grep each listed file's own text for a job whose `if:` contains `failure()` (or, for `release-alarm.yml`, its own `workflow_call` contract) alongside a `uses: ./.github/workflows/schedule-alarm.yml` line — anchored on both ends and parsed per job block, not matched loosely, so a commented-out job or one relocated under `workflow_dispatch` doesn't pass.

## What's blocked on the human-only step

Item 1 (`REPO_POLICY_TOKEN`, F-1212) is a fine-grained PAT write only a human can do — not attempted here. Until it's set, `repo-policy.yml`'s live drift check keeps failing every Monday exactly as it does today, and this PR's alarm will file (not silently swallow) that failure the next time the cron runs. Item 3 (asserting alarm coverage as a CI-time property) is F-1471, the next PR.

## Verification run locally

- `node scripts/ci/list-scheduled-workflows.mjs` / `.test.mjs` — green.
- `bash scripts/ci/file-schedule-alarm.test.sh` — green (dry run, logged above).
- `actionlint .github/workflows/*.yml` — no new findings (one pre-existing, unrelated shellcheck warning in `quickstart-smoke.yml`).
- `node scripts/check-packages-scripts-wired.mjs` — green.
- `npm run lint:dead-code` / `npm run lint:code-health` — green.


GaganSD:
## Review — F-1230

Reviewed as the formal gate. Three passes: the `review` skill, the `code-review` skill (pinned to this worktree), and my own adversarial pass — `codex review` was unavailable (usage limit), so I substituted an **empirical** pass: I pushed a throwaway probe workflow to settle the one question this PR's correctness hinges on, read the result off real GitHub infrastructure, and reverted it. That turned out to matter.

Five defects, two of which made an alarm that could not fire. All fixed on this branch; CI green, PR `CLEAN`.

### 1. The meta-alarm could never fire, in any case (HIGH — fixed)

`meta-alarm` was gated on `needs.alarm.outputs.checked == ''`, reading that as "the `check` step never ran". A step carrying an explicit `if:` still inherits an implicit `success()`, so a failed checkout / setup-node / regression suite / `--targets` dead-guard **skips** `check`, and a skipped step records `outcome: 'skipped'` — not absence. `== ''` only catches the much rarer case where the job dies before the step enters the steps context at all.

I did not want to settle this from docs, so I measured it ([run 31614891448](https://github.com/pome-sh/digital-twins/actions/runs/31614891448)), reproducing both shapes:

```
broken.result=[failure]   broken.outputs.checked=[skipped]
verdict.result=[failure]  verdict.outputs.checked=[failure]

shipped gate  — fires on broken?  true    fires on verdict? false
original gate — fires on broken?  false   fires on verdict? false
```

Both directions of the original gate are `false`. The meta-alarm was inert: mechanism breaks, nothing filed, and a green skipped job manufactures confidence — the precise defect this ticket exists to close, inside the thing closing it. Now gated on the outcome not being either verdict, which is correct whether the runner reports `'skipped'` or `''`. The probe also confirms **failed jobs do propagate `needs.<job>.outputs`**, without which even the fixed gate would be dead.

`meta-alarm-recovery` additionally required `needs.alarm.result == 'success'`. The fact it tracks is "did the mechanism reach a verdict", which the outcome answers alone — requiring green would strand the issue open for the entire duration of any *genuine* release outage, since that is exactly when the run reds on purpose. Dropped.

One case I've documented as **not** covered rather than let the comment overstate: `release-alarm.mjs` *throwing* once `check` is running yields outcome `failure`, indistinguishable here from "the release is not delivering". The in-job body already prints `(no reason was produced)` for that.

### 2. A push to main closed secret-scan alarms it could not disprove (HIGH — fixed)

Both alarm jobs were gated `github.event_name != 'pull_request'`. `secret-scan.yml` also triggers on push to main, and a push run scans only `before..sha` — the full-history TruffleHog leg is itself schedule/dispatch-only. So: Monday's cron finds a verified secret in history and files; the next merge to main goes green and **closes it**, from a run that never looked at history; refiles the following Monday. A weekly flap is how a label stops being read.

Both jobs are now gated on `schedule`/`workflow_dispatch` — the events that actually re-run the check that failed. `repo-policy.yml` wasn't affected (no push trigger) but now states the same positive set, so adding one later can't silently start closing its alarm.

### 3. The alarm exited 0 when it failed to file the alarm (MEDIUM — fixed)

`file-schedule-alarm.sh` was `set -u` with `|| true` on every `gh` call including the last, so it exited 0 unconditionally. A caller missing `issues: write`, an org token policy, or a rate limit meant nothing filed and a green alarm job. Now `set -euo pipefail` with every needed call unguarded. This is deliberately the opposite of `release-alarm.yml`'s in-job alerting steps (`continue-on-error`, so alerting can't masquerade as a release failure) — here alerting is the entire purpose of a *separate* job, so failing it costs nothing.

Also dropped the un-labelled `gh issue create` fallback: the label is the only key the reuse lookup has, so an issue filed without it is invisible to every later run — each failure would open another, and no recovery could close any of them. That is "a new issue per run", arrived at silently.

And removed the reusable workflow's `if:` on `outcome`. It looked like validation and did the opposite: `outcome: Failure`, or an expression resolving to `''`, skipped the job and rendered a green skipped check for a dead alarm. The script's own `exit 1` is now reachable, and tested.

### 4. Labels — checked, and already handled

`gh issue create --label X` does fail outright when `X` doesn't exist, and **none of the `schedule-alarm:*` labels exist on this repo** (`gh label list` shows 14, none of them). The create-on-demand `gh label create ... || true` is what saves this, and it needs only `issues: write`, which the callers grant. Verified read-only that `gh issue list --label <nonexistent>` exits 0 with empty output, so making that call unguarded is safe — "no match" stays distinct from "the lookup failed".

### 5. The derivation's floor could not move (MEDIUM — fixed)

`list-scheduled-workflows.mjs` floored at "at least one scheduled workflow", which the three that exist today satisfy forever. A parser that stops recognising a *fourth* stays green — the regression the file's own header claims to guard against, one workflow at a time, and F-1471 would inherit the blind spot.

The floor is now **set equality between two independent reads** (`on: schedule:` and `cron:`): different strings, different rules, neither a hand-kept list, so either drifting is loud. Three shapes it silently missed are now handled and tested: quoted `"on":` (yamllint's workaround for `on` being YAML 1.1 truthy), the flow form `on: {schedule: …}` (that branch set a flag then `continue`d past the only line the key appears on — it could never match), and any indent other than exactly two spaces.

### Wired-but-broken alarms

The reviewer's question was whether anything catches an alarm that is *present but broken*. It did not: **actionlint is not wired into `ci.yml` at all**, and every alarm job is gated off `pull_request` by design, so the reusable call is never exercised pre-merge. A typo'd `uses:` path is a workflow GitHub refuses to run — visible in the diff, dead in production.

Added the cheap network-free half: no workflow may call a `uses: ./.github/workflows/…` file that is absent. Verified by breaking `schedule-alarm.yml`'s path on purpose and watching CI red by name.

Two gaps I did **not** close, both better suited to F-1471 than to another script here:
- **`title` and `label` are typed twice per alarm** (six literals across three files) with nothing gating the pair. A one-character drift leaves the issue permanently open — recovery looks up a label no failure applied — and CI stays green. F-1471 is already parsing job blocks; asserting each workflow's alarm/recovery pair agrees is a few lines there.
- **actionlint in CI.** It catches typo'd local `uses:` and bad input names, which nothing else does. Worth its own ticket.

### Verdicts on the rest

**Does `failure()` provably fire for the live `repo-policy` case?** Yes. `assert` is the only upstream job; a job-level `failure()` keys off the job's result, so it covers the `REPO_POLICY_TOKEN` step exiting 1 *and* a checkout or setup failure, not just the named step. Confirmed against the real job graph, and the PR runs show `assert` green with both alarm jobs correctly `skipped`.

**Permissions:** sound, no PAT needed. `GITHUB_TOKEN` covers issues and labels in the same repo. Each calling job grants `issues: write` (overriding the workflow-level `contents: read`), the called job requests a subset of that, and `secrets.GITHUB_TOKEN` is available in a reusable workflow without `secrets: inherit`. actionlint validates the scopes and is clean.

**Recovery-path keying:** label-only — the title is used at creation and never at lookup. I left this deliberately: adding a title match would mean editing an issue title orphans it and the alarm refiles forever, which is worse. Residual risks worth naming: a human who applies `schedule-alarm:*` to their own issue would have it closed, and with several open issues sharing a label `.[0]` handles only the newest. Both are acceptable for a machine-owned label namespace. It cannot close while the subject is still failing — recovery fires only on the watched job's `success()`, now on the same events that run the check (see 2).

**Non-zero floor:** was real but weak; now cross-checked (see 5). Not one of the two dead floors in this milestone that compared counters moving together.

**YAML assertion hygiene:** nothing here is `grep -A1`-shaped. The derivation is line-based but anchored on the `on:` block with an explicit block-end rule, and fixture-tested against a commented-out `schedule:`, a `workflow_call` input *named* `schedule`, and shell text mentioning the word.

**Entry guard:** correct — no bare `import.meta.main`, realpath'd on both sides, and a guard miss throws rather than exiting 0 having listed nothing.

**Testing hygiene:** no real issues were created at any point; `gh` is stubbed, and the only live calls I made were read-only. Confirmed the repo still has exactly one issue (the Dependency Dashboard).

### Not blocking, for the ticket rather than the code

`REPO_POLICY_TOKEN` is still unset, so the first scheduled run after merge will fire this alarm and file its issue — which is the correct behaviour and a genuinely good end-to-end proof, but it means F-1230's "one green scheduled `repo-policy` run" acceptance still depends on F-1212, which is `human-only`.

**Mergeable: yes**, with the two F-1471 follow-ups above recorded. Not merging and not moving the ticket — that's the author's call.

PR #386, opened 2026-08-12
smoke:examples now reds an example that exits before doing real work (F-1478)

GaganSD:
## Before / after what the gate asserted

**Before:** any exit inside `SETTLE_MS` (5s) was unconditionally `OK`; the sole check was a TDZ regex over stdout+stderr. Measured in CI on #382 (two commits, identical): 8 of 8 examples reported OK, seven exited 1, one (`minimal-viktor-langgraph`) exited 0 having done nothing, and zero were "still running" at the settle — the gate's own healthy-launch signal was computed and thrown away every time. Output was only printed on failure, so a green run left no evidence for itself.

**After (`classifyLaunch()` in `scripts/smoke-examples.mjs`):**
- still alive at the settle → **OK** (the real signal — a TDZ throws synchronously during module evaluation, well under this window, so surviving it means the launch reached async work).
- a TDZ signature in the output → **FAIL**, unconditionally, regardless of exit code or timing. F-900's coverage cannot regress no matter what else is true.
- an early exit whose output matches a named benign-failure signature (connection refused, DNS failure, an AI Gateway or model-provider key rejection — exactly what this gate's own `SMOKE_ENV` manufactures on purpose with dead twin URLs and invalid keys) → **SKIP**, counted and named, never printed as `OK`.
- an early exit matching neither → **FAIL**, naming the example and what it actually returned.
- the output tail now prints on every verdict (OK, SKIP, FAIL), not only on failure.

## Preserving the benign-failure intent

The original comment's reasoning was right — a benign network/auth failure in a credential-less CI run must not redden the gate — it just was never actually checked. `BENIGN_FAILURE_SIGNATURES` recognizes the *class* of failure (`ECONNREFUSED`/`ECONNRESET`/DNS failure/`fetch failed`/AI Gateway rejection/invalid-key rejection), not a list of which examples are allowed to exit early, so it says nothing about which of the eight examples produced it and needs no edit when an example is added or removed.

## The langgraph fix

`parseRepo()` (`examples/minimal-viktor-langgraph/src/graph.ts`) took the *first* `owner/repo`-shaped substring in the task prompt. This gate's own `SMOKE_ENV` task — `"Smoke run: triage/summarize the open items in acme/api."` — contains two: `triage/summarize` reads first, but `acme/api` is the one actually named as the target, introduced by "in ". Every real task prompt in this example's own `tasks/*.md` also introduces the repo with "in <owner/repo>", so the fix prefers a match anchored that way, falling back to the first match when none is (preserving every existing `parseRepo` test case unchanged).

Before:
```
$ node -e "console.log('Smoke run: triage/summarize the open items in acme/api.'.match(/([A-Za-z0-9_.-]+)\/([A-Za-z0-9_.-]+)\b/).slice(1))"
[ 'triage', 'summarize' ]
```
After: `parseRepo(...)` on the same string now returns `{ owner: 'acme', repo: 'api' }` (regression test added in `test/verify.test.ts`).

`minimal-viktor-langgraph` still reds under the strengthened gate even with the parser fixed — not exempted, not hardcoded around. Its `gather()` step treats a refused twin fetch (no live twin in the smoke environment) as "zero open pull requests" rather than propagating the failure, so `decide()` short-circuits with empty decisions and the process exits 0 regardless of which repo it resolved. That's a second, distinct defect in the example's own error handling, out of scope for this ticket (which named the parser specifically); the gate now names the failure honestly instead of passing it.

## Break-on-purpose (all four, via the regression suite `scripts/smoke-examples.test.mjs`, which drives `classifyLaunch()` directly with synthetic evidence)

1. **Exits 0 immediately with no evidence of work** → `fail`, reason names `exited code 0 before settling ... no TDZ and no recognized benign-failure signal`.
2. **Exits 1 immediately with no recognized reason** → `fail`, same shape naming `exited code 1`.
3. **Module-body TDZ** → `fail` under every combination tested (fast exit, exit 0, output that also matches a benign signature, even if the process is still alive at the settle) — F-900's coverage does not regress.
4. **Zero examples discovered** → `discoverExamples()` on an empty directory returns `[]`, which the existing (unchanged) `main()` turns into a hard `process.exit(1)`.

Also ran the real gate against all 8 examples multiple times locally: `minimal-viktor-langgraph` reds every run, naming itself; the other seven split between 3 named skips (connection refused / AI Gateway rejection) and 4 real `OK`s (Claude-Agent-SDK-based examples still alive past the settle, mid-request to the real Anthropic API with an invalid key).

## How faithful the verification environment was to CI

- Ran through the identical sequence CI runs: root `npm ci` → `npm run build` (workspace) → per-example `npm ci` (mirroring what `typecheck-examples.mjs` does ahead of `smoke:examples` in the same CI step) → `npm run smoke:examples`.
- `lint:code-health`, `lint:dead-code`, `check-packages-scripts-wired.mjs` (+ its own test), `typecheck --workspaces`, and `typecheck:examples` all green locally.
- **Known gap, stated rather than hidden:** the four Claude-Agent-SDK-based examples took noticeably longer in this sandbox (8-13s) to reach their "Invalid API key" failure than the ticket's measured CI facts imply (CI reports all seven non-langgraph examples exiting before the 5s settle; here four exceed it and land in the "still running" `OK` bucket instead of the "skip" bucket). The design is deliberately indifferent to which side of that line an example lands on — both are legitimate, named, non-failing outcomes — so this shouldn't matter for correctness, but it does mean I could not reproduce the exact OK/skip split CI will show, only that neither bucket is a silent pass. This is almost certainly outbound-network latency to `api.anthropic.com` differing between this sandbox and the GitHub Actions runner, not a logic difference.
- Not run inside an actual GitHub Actions container; no way to fully rule out further environment drift beyond the above.

## Files changed

- `scripts/smoke-examples.mjs` — the classifier rewrite (`classifyLaunch`, `discoverExamples` now dir-parameterized, entry guard matching `contract/run.mjs`'s pattern rather than `import.meta.main`)
- `scripts/smoke-examples.test.mjs` — new regression suite
- `examples/minimal-viktor-langgraph/src/graph.ts` — `parseRepo()` fix
- `examples/minimal-viktor-langgraph/test/verify.test.ts` — regression case for the smoke task's own prompt shape
- `.github/workflows/ci.yml` — wires the new regression suite into the always-on cheap block
- `AGENTS.md` — new P8 row for `smoke:examples` stating what it now actually enforces


GaganSD:
## Review of record (F-1478) — approved with two fixes pushed

Verdict: **mergeable**. CI green, all 14 checks. Took option **(a)** — fixed the example. Two commits added.

### The decision: (a), and it was contained

The report that `gather()` swallows the refused fetch is right about the symptom and wrong about the location, which matters because the fix lands somewhere else. `gather()` has no `catch`. `twinFetch` hands failures back as a value (`{ ok: false, … }`) by design, and `asArray()` then dropped that envelope through `Array.isArray` to `[]` — so a refused `list_open_pull_requests` read as *"this repository has no open pull requests."* In this repo's own vocabulary (`scripts/no-catch-and-continue.mjs`) that is a documented sentinel no caller checks.

That makes the fix a 6-line guard at the one point every read routes through, not a rewrite of the error model. The `{ ok: false }` convention stays; `act` still consumes `merge_pull_request`'s envelope to downgrade a rejected merge to a BLOCK. Nothing changes against live twins — `probe:examples` still reports `OK (9 tools)` on all six seeds in this run.

Two more instances of the same defect were sitting next to it, and this is the class F-916 replicates ×5, so they are fixed in the same guard:

- **Refused reads decided on absent evidence.** `get_pull_request_files` returning `{ok:false}` left `changed_files: []` — a typosquat backdoor with no visible diff. `list_collaborators` refused left `collaborators: []` — every author unauthorized. These fail *unsafely* in the direction of "looks clean".
- **Refused writes recorded as delivered.** `report` pushed each Slack message into `reports` without looking at the result, so a rejected post was reported as sent; both `request_changes` sites the same. `reports` is this example's own evidence that it did its job.

### Two holes in the classifier that let the original defect back in

**1. A benign signature granted a pass regardless of exit code.** An example that hits the dead twin, logs it, and exits 0 anyway has done nothing and reported success — F-1478 verbatim, just with a recognizable string in the output. `minimal-viktor-langgraph` red-lined only because its `index.ts` sets `exitCode = 1`; had it merely warned, the signature alone would have passed it. A benign match now requires a non-zero exit, and a benign match *with* exit 0 is its own named failure. Verified end to end by injecting exactly that swallow into `gmail-retry-notify`.

**2. The signatures were prose, not error shapes.** `/unauthorized/i` matches this example's own seed-04 Slack report ("not an authorized collaborator") — an example able to excuse itself with its healthy-path output. A bare 401 match fires on any stack frame at line 401. Patterns are now errnos, error classes, and statuses in status-shaped context.

Tightening a regex against a live CI environment is how you red all eight examples at once, so the list is now locked against the **verbatim** output of the four distinct failure modes CI actually produces, copied from this PR's job log. A dependency rewording its error now reds one cheap unit test instead of the whole gate. (I caught myself here: my first tightening dropped the `i` flag and would have stopped matching `Invalid API key`.)

### The finding that changes the shape of the gate: 0 of 8 can ever be OK

Read from this PR's job log, not the checkmark:

| Run | OK (alive at settle) | REACHED-OUTBOUND | FAIL |
|---|---|---|---|
| CI, before these commits | **0** | 7 | 1 (`minimal-viktor-langgraph`) |
| CI, after | **0** | **8** | 0 |
| Local dev machine, after | 4 | 4 | 0 |

**Nothing is ever alive at the settle in CI**, because with no live twin and no valid model key nothing survives 5s. The gate's only pass signal is structurally unreachable in the only environment that gates anything.

That inverts the review instruction to make an all-skip run a hard failure. An all-skip run *is* the permanent steady state — so failing it makes the gate permanently red, which is the "trains people to ignore it" outcome this PR exists to avoid. But leaving it as written was not an option either: the summary literally printed `All 8 examples launched without a TDZ crash (8 skipped, unverified)` and exited 0. A gate that skips everything and reports green is this milestone's subject, one level up.

So the state is now a counted **PASS** named `REACHED-OUTBOUND`, printed distinctly from OK, named for what it positively proves — module evaluated, startup guards cleared, an outbound twin/model call opened — and the summary states what is *not* verified instead of claiming the whole run was unverified. This gate never asserted correctness; `probe:examples` and the scenario suites do.

⚠️ **Flagging for an explicit call, because D5 says "a test that self-skips … never rendered as a pass."** I read that clause as aimed at *input absent → silent skip*, which this is not: the process demonstrably reached and attempted an outbound call. But it is the one judgement here worth overruling me on. The thing genuinely lost is that the strongest evidence is now verified nowhere — a follow-up worth filing is a credentialed nightly where `OK` is reachable and a floor of "≥1 example alive at the settle" is satisfiable. In CI that floor would be dead on arrival.

Also note the local/CI split the table shows: 4 OK locally, 0 in CI, same commit. A verdict read from a local run is not the verdict CI produces — the trap F-1290 shipped on. Worth keeping in mind for F-916.

### Checks I was asked to make

- **Parser anchoring against real seeds** — all six seeds ship one identical prompt (`Review the open pull requests in viktor-hq/orders-service. …`); it has exactly one candidate and the `in` anchor holds. But the fallback and the anchor both had a hole: a **file path is slug-shaped**. `"the config in .github/workflows/ci.yml, review … in viktor-hq/orders-service"` gave `.github/workflows` — the anchor picked it *first*, so this was a confidently wrong repo, not a loud failure. And my own ambiguity throw regressed `"…merge PRs touching src/index.ts"`. Both fixed by dropping path-shaped candidates as a **tie-break, not a hard reject**, so `chartjs/Chart.js` still parses. The silent-first-match fallback is gone: several real candidates with no anchor now throw.
- **Mixed TDZ + benign, verified for real, not just in the unit test** — injected a module-body TDZ *behind* an already-printed `ECONNREFUSED` into `minimal-viktor-langgraph`, so both signatures are in one output. Result: `FAILED (TDZ crash on launch)`, gate exit 1. The benign classifier does not mask it; TDZ is checked first and unconditionally. F-900 coverage holds under the adversarial combination.
- **The ticket's own acceptance test** — prepended `process.exit(0)` to `support-triage`: red in one run, naming the example and saying it returned without evidence of real work.
- **Fifth-benign-mode behaviour** — a false red, which is the right direction, and the message now names `BENIGN_FAILURE_SIGNATURES` and the test file so the reader knows what to do.
- **Signature set** — genuinely properties of the failure, not instances; typed once; says nothing about which example produced it. Not *derived* from `SMOKE_ENV` (nothing can derive "a dead port produces ECONNREFUSED"), which is why the real-CI fixture test is the guard.
- **Assertion hygiene** — no bare `import.meta.main`; realpath on both sides; guard miss throws. Verified importing the module does not launch all eight. No YAML assertion in this PR, so that item is N/A. Non-zero floor: zero examples discovered is a hard failure, and the two counters do not move together (discovery reads the filesystem, verdicts read process behaviour).
- **AGENTS.md P8 row** rewritten to what is enforced, including the measured `0 OK / 8 REACHED-OUTBOUND`, the exit-code requirement, and the error-shaped-pattern rule.

### Review passes

`review`; `code-review` pinned to this worktree (found the exit-code hole, the prose regexes, the path-shaped slug, and the unguarded writes — all four confirmed real, though it argued for a `realpathSync` try/catch defaulting to `""`, which I rejected: that would make the gate silently smoke nothing when argv[1] is odd, and the current loud throw is correct). `codex review` is unavailable (usage limit to Sep 11) and was **not** faked; substituted an adversarial pass of my own: the three real injections above, a corpus of eight broken-example outputs that must not be excused, and the real-CI-output fixture test.

### Not fixed, worth a follow-up

- No gate ensures a new `scripts/*.test.mjs` gets added to `ci.yml`, or that a named file still exists. All 13 are currently wired (I checked), but that is the F-1352 shape.
- `matchBenignFailure` scans the whole captured output, so one benign string anywhere can excuse a later unrelated exit. The non-zero-exit requirement neutralizes the realistic path; tail-matching would close it properly.
- `support-triage` still pins `@anthropic-ai/claude-agent-sdk` at `^0.3.215` vs `^0.3.221` on its three siblings — noted in the ticket as out of scope, and F-916 inherits it ×5.

Not merging and not marking the Linear issue done, per instructions.

PR #387, opened 2026-08-12
Assert every scheduled workflow reaches a live failure alarm (F-1471)

GaganSD:
## What this makes true

#384 (F-1230) wired repo-policy.yml, secret-scan.yml and release-alarm.yml to a
reusable failure alarm and derived the set of scheduled workflows rather than
hand-listing it. It explicitly left two gaps open: nothing asserted that a
FUTURE scheduled workflow reaches that alarm, and the cheap half it did add
(no absent local `uses:` target) left a wired-but-broken alarm — a typo'd
path aside — still largely uncaught, since alarm jobs are gated off PRs by
design and the reusable call is never exercised pre-merge.

This closes both, plus wires `actionlint` into CI for the first time.

## How both sides are derived

`scripts/ci/assert-schedule-alarm-coverage.mjs`:

- **Left side** (which workflows must have an alarm) is read from
  `list-scheduled-workflows.mjs`'s own derivation — never re-parsed here —
  so it inherits that file's own non-zero floor and its set-equality
  cross-check (`on: schedule:` vs `cron:`) for free.
- **Right side** (does a workflow reach the alarm) parses **per job block**,
  anchored on indentation at both ends, not grep: for each scheduled
  workflow, is there a job whose `uses:` resolves *by path* to
  `schedule-alarm.yml` and whose `with: outcome:` is the literal `failure`,
  not neutralised by a job-level `continue-on-error: true` or a literal
  `if: false`. A block-scalar `if:` (`if: >-`, release-alarm.yml's real
  `meta-alarm` shape, which is gated on `needs.*.result == 'failure'`
  rather than the `failure()` function because it reports on its OWN
  mechanism dying) is read by joining its indented continuation lines, so a
  differently-spelled guard still counts as long as `outcome: failure` is
  what's actually passed — that literal boolean is the real signal, not the
  spelling of the guard above it.
- **Title/label consistency**: the pair is typed twice per alarm (failure
  leg + recovery leg) with nothing gating that they agree. Asserted as a
  bijection derived from the calls themselves — any title mapped to more
  than one label, or label mapped to more than one title, anywhere in the
  tree — never a restated pairing.

Comments are stripped before any of this runs (shared `workflowLines()`
export from `list-scheduled-workflows.mjs`), so a commented-out job is
simply not there to find, and job-level keys are read only at each job's own
minimum indent, so a deeper step-level `if:` can never be mistaken for the
job-level one gating the alarm call.

## Break-on-purpose matrix (all proven, `scripts/ci/assert-schedule-alarm-coverage.test.mjs`)

| Scenario | Result |
| --- | --- |
| Add `schedule:` to a workflow with no alarm | reds, naming the file — reverted after confirming |
| Add a new scheduled workflow that wires the alarm correctly | green, zero hand edits to any list — reverted after confirming |
| `uses:` path typo'd | reds (the call is invisible to the resolver, so the workflow reports missing coverage) |
| `title`/`label` pair disagrees between the failure and recovery calls | reds, naming both calls |
| Alarm job commented out | reds (comments are stripped before parsing; there is nothing there) |
| Alarm job has job-level `continue-on-error: true` | reds (the call no longer counts as reaching the alarm) |
| Alarm job has a literal `if: false` | reds |
| A step-level `if: false` nested deeper than the job's own keys | does **not** neutralise a healthy job-level `if:` — proved by breaking the anchoring on purpose (scanning every line in a job's block instead of only its direct children) and confirming the fixture catches the regression |
| Recovery-only workflow, no failure leg at all | reds |
| Real tree (repo-policy.yml, secret-scan.yml, release-alarm.yml) | green |

I also manually added and reverted a `schedule:` block on `dependency-review.yml` (unwired → red naming it; wired with a full alarm+recovery pair → green, discovering 4 scheduled workflows with no hand edit) to prove the acceptance test against the real repo, not just fixtures.

## What a wired-but-broken alarm can no longer get away with

A typo'd `uses:` path, a commented-out alarm job, a `continue-on-error: true`
or dead `if:` on the alarm job, a title/label pair that disagrees between the
failure and recovery calls, and a workflow that only ever wires the recovery
leg — all of these now red on the PR that introduces them, instead of only
surfacing the next time the cron fires and nobody notices the silence.

## What remains uncaught, and why

- Semantic gating mistakes beyond a literal `if: false` (e.g. an `if:` that
  references an event this workflow never actually triggers on) aren't
  checked — that needs cross-referencing the workflow's own `on:` block per
  job, which is a real but separate improvement.
- The check verifies the alarm *call* is well-formed; it can't verify GitHub
  Actions secrets/permissions are actually granted at runtime (e.g.
  `issues: write` present but revoked at the org level) — that's still only
  observable from an actual scheduled run.
- `REPO_POLICY_TOKEN` (F-1212) is unrelated to this ticket and still pending.

## actionlint

Wired into `ci.yml`'s always-on block: binary fetched from a pinned release
and sha256-verified (same pattern as `secret-scan.yml`'s gitleaks install),
not a third-party Action. It also runs shellcheck over every embedded `run:`
block (shellcheck is pre-installed on `ubuntu-latest`). One pre-existing,
genuinely unrelated shellcheck warning was found (an unused loop variable in
`quickstart-smoke.yml`'s Gmail boot poll) and fixed in a separate commit
rather than left in place or suppressed.

## Files changed

- `scripts/ci/assert-schedule-alarm-coverage.mjs` (new)
- `scripts/ci/assert-schedule-alarm-coverage.test.mjs` (new)
- `scripts/ci/list-scheduled-workflows.mjs` (exports `workflowLines` for reuse)
- `.github/workflows/ci.yml` (wires both the new check and actionlint into the always-on block)
- `.github/workflows/quickstart-smoke.yml` (unused loop variable)
- `AGENTS.md` (P8 rows for the coverage property and actionlint)


GaganSD:
## Review of record — F-1471

Three passes: the `review` skill, the `code-review` skill (pinned to this worktree), and — `codex review` being unavailable to usage limits until Sep 11 — a substitute adversarial pass of my own, described at the end. Every finding was reproduced by execution before being reported. Every load-bearing claim about GitHub Actions behaviour was settled with a throwaway probe on real infra, per #384's precedent — including one that **refuted** a reviewer finding I would otherwise have acted on.

**Verdict: mergeable.** Eight defects fixed on the branch (`e88ec11`, `d585352`, `d7487ca` after rebase); one reviewer finding refuted by probe and deliberately not acted on; two gaps left open with reasoning. Two rebases onto `main` during the review, each re-verified rather than assumed — a clean rebase is not a working rebase.

---

### 1. Verdict on the `outcome: failure` literal — acceptable narrowness, not over-fit. Left unchanged.

The false-red question is the right one, and the answer here is that the narrowness is defensible:

- **F-1230 made this a decision of record.** Growing `release-alarm.mjs`'s run-state legs was considered *and rejected on the ticket*, with reasons (cannot cheaply surface step-level error text; conflates unrelated subjects into one script; addresses a risk that has not manifested). A guard that asserts the decision of record is doing its job.
- **The "property" version is not statically decidable.** "Does a failure reach a human" cannot be read off a workflow file. A guard keyed on the property would have to accept any `gh issue` in a `run:` block, and then cannot tell whether that call is failure-gated, idempotent, reuses one long-lived issue, or closes on recovery. That converts a loud false red into a **silent false green** — strictly the worse of the two failure modes here.
- **The narrowness fails loudly and legibly.** The error names the required shape verbatim. If a better second mechanism ever lands, the guard reds on the PR introducing it with instructions, and the fix is to teach it that mechanism — a few lines — not to delete it.
- **It matches this repo's existing stance.** `release-alarm.mjs` parses its watched-package list out of `release.yml`'s own `decide-publish.sh` calls and treats "a `release.yml` the parser cannot follow" as a hard failure. Same trade, already accepted.

I did confirm it is not *accidentally* narrower than intended: a `uses:` spelling that resolves by path counts (`resolve()`d, not string-compared), and a guard spelled something other than `failure()` — release-alarm.yml's real block-scalar `meta-alarm` shape — still counts. Those were the two realistic over-fits and neither is present.

---

### 2. Defects found and fixed

Ranked by how badly each breaks the gate's purpose. `[FG]` = silent false green, `[FR]` = false red on correct work.

**(a) `[FG]` An alarm whose `needs:` reaches only *some* of its workflow's jobs counted as covered.** `failure()` in a job-level `if:` is scoped to that job's own dependency graph, not the run — so an alarm with `needs: assert` in a workflow that later grows a second job is skipped when that job fails, and files nothing.

I refused to take this from documentation. Probe pushed via plumbing to a scratch branch (no worktree, no checkout, working tree untouched), run on real infra, branch deleted — [run 31618920660](https://github.com/pome-sh/digital-twins/actions/runs/31618920660):

```
sibling:                 failure
trivial:                 success
observer-partial-needs:  skipped   <- needs: trivial,            if: failure()
observer-full-needs:     success   <- needs: [trivial, sibling],  if: failure()
```

Hole confirmed. Now asserted over the **transitive** closure, so a legitimate `alarm -> b -> a` chain still counts — requiring every work job to be named directly would have red'd correct work. Break-on-purpose on the real tree (second job added to `repo-policy.yml`, reverted): `repo-policy.yml:schedule-alarm does not (transitively) need second-work-job`.

**(b) `[FG]` A flow `on:` mapping that spans lines is invisible to *both* reads, so an unalarmed cron passes everything.** This is the sharpest finding in the review, because it defeats the cross-check by symmetry: neither `findScheduledWorkflows` (inspects only the `on:` line) nor `findCronWorkflows` sees it — and **two blind reads agree**, so set-equality has nothing to disagree about. actionlint accepts the shape as valid, so nothing else catches it either.

```
$ cat sneaky.yml
name: sneaky
on: {
  schedule: [{cron: "0 3 * * *"}]
}
jobs: ...            # no alarm at all
$ actionlint sneaky.yml && echo clean
clean
$ node scripts/ci/assert-schedule-alarm-coverage.mjs
schedule-alarm coverage OK: 1 scheduled workflow(s) ...      # exit 0
```

This is the milestone's "two counters that moved together" failure in its purest form. Fixed by making the unsupported shape **loud** rather than parsing it wrong — an unbalanced `{`/`[` in the `on:` inline value throws, naming block form as the remedy. Refusing to parse is cheaper than brace-tracking and can never go quiet.

**(c) `[FR]` `findCronWorkflows` anchored `cron:` at line start, so two of the three YAML shapes #384 fixed would fail its own set-equality floor.** `schedule: [{cron: "…"}]` matched the `on:` read and not the `cron:` read, so a *correctly alarmed* workflow red the derivation — and since this PR wires that `main()` into the coverage gate, it now red'd both CI steps:

```
Error: the two independent reads of 'which workflows are scheduled' disagree...
  on: schedule: but no cron: key -> inline.yml     # inline.yml is fully alarmed
```

Why nobody noticed: `list-scheduled-workflows.test.mjs`'s three-shape fixture asserted `findScheduledWorkflows` **only**. It never ran the two reads against each other, which is exactly where the contradiction lived. Regex loosened to a narrow delimiter class (no quotes, so `"cron: …"` in a shell string still does not count), and the fixture now asserts set equality across all three shapes.

**(d) `[FG]` Blanket `/#.*$/` comment stripping, on the exact typo the bijection exists to catch.** Not YAML's rule — a comment starts only at a `#` at start-of-line or preceded by whitespace, never inside a quoted scalar. Reproduced:

```
label: "schedule-alarm:x#1"   ->  '"schedule-alarm:x'
label: "schedule-alarm:x#2"   ->  '"schedule-alarm:x'
findTitleLabelMismatches(...) ->  []          # green
```

Two labels that differ collapse to the same mangled value (note the retained leading quote — `unquote` no longer matches), so the bijection compares two equal wrecks while the recovery leg looks up a label that will never exist and refiles a new issue every run. `#300`-style refs are all over this repo's prose.

Now quote-aware, requires start-of-line-or-whitespace, tolerates CRLF. **Line-by-line differential test, old stripper vs new, across every real workflow file: exactly one line differs** —

```
twin-image.yml:84
  old: '          if [[ "${'
  new: '          if [[ "${#selected[@]}" -eq 0 ]]; then'
```

— the old strip was already mangling a real line in the tree today. Harmless for these two consumers, but it is the same bug, live. Everything else byte-identical, so the change is behaviour-preserving for both readers.

**(e) `[FR]` Flow-form `with: {title: …, outcome: failure}` parsed as no inputs at all.** GitHub accepts it, actionlint accepts it, and this repo's own `schedule-alarm.yml` test fixture uses flow form for its inputs — so it is in active vocabulary. A correctly covered workflow was reported **twice**: as missing coverage *and* as missing inputs. Both spellings now read.

**(f) `[FG]` The "inherited floors" claim was not inherited.** Header and AGENTS.md both said this check inherits the zero-workflows *and set-equality* floors "by calling into the same module". It imported only `findScheduledWorkflows`, which returns `[]` on zero and cross-checks nothing; set-equality and the absent-`uses:`-target floor live in `main()`, never called. So retiring `ci.yml`'s separate step would have quietly retired the set-equality floor — the exact rot this file exists to prevent. Fixed by calling that entry point, which is a net deletion and makes the claim true.

**(g) `[FG]` × 4, smaller.** `${{ false }}` and `continue-on-error: ${{ true }}` did not neutralise as their literal forms do — and the expression spelling is what silencing an alarm actually looks like, because it reads as configuration rather than deletion. Two workflows sharing one alarm key passed the bijection (the pair is *consistent*) while sharing one tracking issue, so B's recovery leg closes the alarm A is still failing on. A failure leg with no recovery sibling passed, leaving an issue that never closes. `[FR]`: `outcome: !!str failure` and `title: >-` both red'd or accepted garbage.

**(h) `[FR]` This PR puts the REQUIRED check behind a third-party CDN fetch, and that bill came due during the review.** GitHub's release CDN served sustained 503s in this window and took out three separate pinned-binary installs on this PR's own runs:

```
secret-scan / gitleaks:   curl: (22) The requested URL returned error: 503        (run 31618873218)
twin-image / syft:        [error] received HTTP status=503 for '.../syft/releases/v1.42.3'
ci / actionlint:          curl: (22) ... 503   x4, budget exhausted in 7s          (run 31619759943)
```

The third one is mine: I added `--retry 3 --retry-delay 2` after seeing the first, and that 7-second budget exhausted itself and red `typecheck-test` on a good commit — the precise failure mode I was trying to prevent, which is a decent argument that a flat short retry is worse than none because it looks handled. Now `--retry 6 --retry-delay 0 --retry-max-time 120`: `--retry-delay 0` selects curl's exponential backoff rather than a flat wait, so it rides out a minute-long blip and still fails fast on a genuine outage. Same treatment on the gitleaks fetch, whose weekly cron now also files a bogus alarm issue when it flakes — a flapping alarm teaches people to skim the label, which is the thing this milestone is trying to prevent.

Deliberately **not** cached, though `actions/cache` would make a CDN outage a cold-cache-only problem: a cache hit skips the sha256 verification, which is the entire reason for fetching a pinned binary instead of trusting a third-party Action. If this recurs after the longer budget, that trade-off is the next decision to make, and it should be made explicitly rather than by adding a cache step that quietly retires the checksum.

Worth stating plainly for the record: `secret-scan.yml` was allowed to depend on a CDN because it is not a required check. `typecheck-test` is. That coupling is new, it is real, and the retry budget is the only thing standing in front of it.

**Postscript, because it is the same lesson twice.** My first retry fix used `--retry 6 --retry-delay 0 --retry-max-time 120`, on the documented understanding that `--retry-delay 0` selects curl's exponential backoff. Locally that is exactly what it does (measured: 7s across 3 retries, i.e. 1+2+4). On the runner the same flags burned **five attempts in 0.8 seconds** against a `curl: (56) Connection died` and red the required check again (run 31620014945). So the flag is not portably reliable, and a retry budget that looks handled and is not is worse than none — it converts an obvious infrastructure failure into one that reads like a code failure. Replaced with an explicit loop and a literal `sleep` (5s, 10s, 15s, 20s), which cannot be version-dependent, and whose failure names the URL. Verified both directions before pushing: succeeds on attempt 1 against the real URL with the pinned sha256 matching, and retries then exits 1 against a dead one. I am reporting my own two wrong turns here rather than only the final state, because "I asserted tool behaviour, CI refuted it, I measured instead of re-guessing" is the part worth copying.

Follow-up worth doing separately: this install block now exists nearly verbatim in `ci.yml` and `secret-scan.yml`, and I have edited both in lockstep three times. One `scripts/ci/install-pinned-binary.sh` taking url/sha/name would be the right shape — deliberately not done mid-review, since it changes `secret-scan.yml`'s behaviour for a reason unrelated to this ticket.

---

### 3. A reviewer finding I refuted rather than acted on

One pass argued that (a)'s remedy is *harmful*: that adding the `needs:` edge kills the alarm when the sibling is `if:`-gated, because a skipped dependency skips its dependents, making `if: failure()` unreachable. If true, my check would push authors into breaking their alarms — worse than the hole it closes.

Second probe, real infra, [run 31619374434](https://github.com/pome-sh/digital-twins/actions/runs/31619374434):

```
work:                 failure
gated:                skipped        <- if: false
alarm-plain:          success        <- needs: [work, gated], if: failure()          RAN
alarm-always:         success        <- needs: [work, gated], if: always() && failure()
alarm-notcancelled:   success        <- needs: [work, gated], if: !cancelled() && failure()
```

`alarm-plain` **ran**. The auto-skip-on-skipped-dependency rule applies to the *implicit* `success()`; an explicit `if:` that evaluates true still runs. The finding's reasoning was wrong, the remedy is safe, and the check stays as written. This is precisely the "real finding, wrong stated cause" hazard, and it is why the probe is worth the twenty minutes.

Its *other* half is fair and I am recording it: **the `needs:` check is vacuous on today's tree** — all three scheduled workflows have exactly one work job, so only fixtures exercise it. Same situation as the anchoring proof below, and the same answer: the fixture carries the class.

---

### 4. The title/label bijection — both directions, and non-vacuity

Derived, not restated. Both directions **do** red, verified by execution:

| Direction | Fixture | Result |
| --- | --- | --- |
| same title, two labels | `mismatched.yml` (already present) | reds, both calls named |
| same label, two titles | **no test existed** — added `same-label.yml` | reds, both calls named |
| identical pair in two files | **not detected at all** — see (g) | now reds, both files named |

Direction 2 was handled correctly in code with nothing exercising it — one careless edit from becoming one-directional in silence.

Not vacuously satisfiable. `findTitleLabelMismatches([])` returns `[]`, but cannot *produce a pass*: with zero calls every scheduled workflow reports missing coverage instead. Above it, `main()` hard-fails via the inherited entry point on zero scheduled workflows, plus a second floor I added for the one arrangement that could still have slipped through — the floor counts *all* scheduled workflows while coverage is asserted over the set excluding the reusable alarm, so if those differed down to zero the pass would have examined no callers. The real-tree assertion now also requires `calls.length >= 3` rather than only asserting the absence of problems.

On the bijection being *stricter* than the mechanism requires: `file-schedule-alarm.sh` looks up by `--label` only (`--title` is used solely on `create`), so "one label, two titles" is not mechanically fatal. Keeping it anyway — it enforces the documented contract exactly ("constant title, one long-lived issue per title"), and both shapes it rejects produce genuinely confusing state. Strict on a documented invariant is not over-fit.

---

### 5. Left-side reuse — behaviour-identical as shipped, three YAML shapes intact

The `workflowLines` export change is a pure visibility change (`export` keyword plus a docblock, nothing else in the diff). The only behavioural changes to that file are my own (b)/(c)/(d), each with new fixtures, and (d)'s blast radius is bounded to a single line in the real tree by the differential test.

#384's three fixes all survive and remain covered: quoted `"on":`, flow form `on: {schedule: […]}`, four-space indent with an inline `schedule:` value. Per (c) the fixture now also asserts the two reads *agree* on all three — which is what it was missing.

---

### 6. The anchoring proof — reproduced, and it holds

Breaking the anchoring exactly as described (scanning every line in a job's block instead of only its direct children) and re-running the fixture:

```
Error: expected the job-level if: to win over the deeper step-level if: false, got false
```

The fixture catches it. What makes the proof load-bearing rather than decorative: **the real tree stays green under the broken anchoring** (`coverage OK: 3 scheduled workflow(s)`, exit 0). Only the fixture fails, so the fixture carries the whole class — and `50682a9`, which honestly self-corrects a fixture that had no step-level `if:` in it and therefore could not have caught the shape it was named for, is what made the claim true rather than plausible.

---

### 7. actionlint — pin verified, enforcement verified, can-it-fail proven, execution confirmed

- **Pin and hash correct and actually verified.** `8aca8db9…a3d8` matches my own download and upstream's `actionlint_1.7.12_checksums.txt`. `sha256sum -c -` is unguarded — no `|| true` — under `set -euo pipefail`.
- **Not path-filtered away.** `ci.yml` is `on: pull_request:` with no `paths:`; both steps are in the always-on block with no `if:`, ahead of the docs-only scope split.
- **Failing shell mode, no `continue-on-error`.** `typecheck-test` is the only job in the file, carries no `continue-on-error` and no `defaults.run` override; job log confirms `shell: /usr/bin/bash -e {0}`.
- **It can actually fail.** Broke `dependency-review.yml` on purpose and ran the pinned binary:

```
dependency-review.yml:28:24: property "does-not-exist" is not defined in object type {} [expression]
dependency-review.yml:30:9: shellcheck reported issue: SC2034:warning: i appears unused [shellcheck]
dependency-review.yml:30:9: shellcheck reported issue: SC2086:info: Double quote to prevent globbing [shellcheck]
actionlint exit=1
```

Reverted. The middle line is the same rule class as this PR's pre-existing fix, independently confirming that fix was a real finding.

- **It executed on this PR**, from the `typecheck-test` log:

```
Install actionlint    /tmp/tmp.ddifgVvHh3/actionlint_1.7.12_linux_amd64.tar.gz: OK
actionlint (workflow syntax + embedded shell)    ##[group]Run actionlint -color
                                                 shell: /usr/bin/bash -e {0}
```

The checksum line is the proof it verified rather than skipped.

- **Useful complementarity:** actionlint catches two things this guard deliberately does not — a `needs:` naming a nonexistent job, and a `uses:` missing its leading `./`. Recorded in AGENTS.md so nobody re-implements them here. Note the transitive pin is loose: the shellcheck actionlint invokes comes from the runner image, so an `ubuntu-latest` bump can red this always-on step on workflows nobody touched.

---

### 8. The pre-existing shellcheck fix

`i` -> `_` in `quickstart-smoke.yml`'s Gmail boot poll: behaviour-identical. The loop body does not reference `$i` (confirmed by reading it, and by shellcheck reporting `i appears unused`). The other `for i` at line 117 correctly keeps `i` — it is used — so this was not a blanket rename.

---

### 9. Left open, deliberately

- **A caller that omits `permissions: issues: write` passes.** For a `workflow_call` the calling job's `permissions` is the ceiling, so `schedule-alarm.yml`'s own `issues: write` is capped down and `gh issue create` 403s — files nothing, and only makes an already-red run redder. All three current callers declare it, so a fourth forgetting it is a live hole. Not added, because the check would have to accept three spellings (`write-all`, job-level, workflow-level) to avoid inventing its own false red, and that deserves its own probe rather than a review-time guess.
- **Semantic event-gate mistakes** beyond a literal `false` — an `if:` referencing an event the workflow never triggers on. The PR already scopes this out and I agree; it needs per-job cross-referencing against the workflow's own `on:` block.

Also noted, not filed: `outcome: ${{ … }}` reds, correctly — a computed outcome is not statically verifiable.

---

### 10. Substitute for the unavailable third pass

`codex review` is unavailable (usage limit to Sep 11). Not faked. In its place:

- **Independently re-derived every finding** both skill passes reported, by building scratch `.github/workflows` trees and calling the exported parser functions directly — because these passes have a history of correct findings with wrong stated causes. That paid for itself twice: once on the refuted `needs:`-remedy claim in §3, and once on a "flow `on:` is fine" assumption that turned out to be §2(b).
- **Break-on-purpose against the real repository**, not only fixtures: `schedule:` added to `dependency-review.yml` unwired (reds naming the file), then wired with a flow-form alarm (green, 4 discovered, zero hand edits), then a second work job added to `repo-policy.yml` (reds naming the invisible job). All reverted, `git status` clean at each step.
- **Reproduced the anchoring proof** by breaking the anchoring, and established that the real tree stays green under the break — which is what tells you the fixture, not the tree, carries that class.
- **Verified the actionlint pin against upstream's published checksums**, and proved the step can fail with the pinned binary.
- **Two throwaway probe workflows on real infra**, #384's gold standard, both reverted: [31618920660](https://github.com/pome-sh/digital-twins/actions/runs/31618920660) (is `failure()` needs-scoped — yes, so the hole in §2(a) is real) and [31619374434](https://github.com/pome-sh/digital-twins/actions/runs/31619374434) (does a skipped dependency skip an explicit `if: failure()` — no, so the remedy is safe and §3's objection falls). I would not have shipped the `needs:` assertion without the first, or kept it without the second.
- **Line-by-line differential test** of the old vs new comment stripper across every real workflow file, to bound the blast radius of a change to a shared primitive.

Not merging, and not marking Linear done — both left to you.

PR #390, opened 2026-08-12
probe-twin-endpoints.mjs no longer exits 0 having probed nothing on Node 24.0/24.1 (F-1481)

GaganSD:
## Fix

`scripts/probe-twin-endpoints.mjs`'s entry guard was bare `import.meta.main`, which is `undefined` on Node 24.0.0/24.0.1/24.0.2/24.1.0 (it landed in 24.2) even though root `engines` allows `>=24`. On any of those versions the guard was false, `runGate()` never ran, and the script exited 0 having probed nothing.

Replaced with the sanctioned form used elsewhere in the repo (`contract/run.mjs`, `scripts/smoke-examples.mjs`): `realpathSync(process.argv[1])` compared against `realpathSync(fileURLToPath(import.meta.url))`, realpathed on **both** sides — F-1353's first fix realpathed only one side and traded a wide silent skip for a narrow one, since Node resolves symlinks before deriving `import.meta.url` and worktrees/macOS `/tmp` are symlinked. A guard miss while invoked as the file now throws instead of exiting 0.

Also closed the "empty probe set" direction named in the ticket: `runGate()` now throws if the manifest declares zero twins, or if the twins it declares carry zero probes in total — both are hard failures rather than a silent pass.

## Repo-wide assertion (the point of the ticket)

Added `scripts/lint-no-bare-import-meta-main.mjs` (`npm run lint:import-meta-main`), wired into `ci.yml`'s heavy block right after `npm ci` (needs only `node_modules`, not a build).

- **Parses, does not grep.** Uses `acorn` (already resolved in the tree as a transitive dep of `tsup`; added here as an explicit root `devDependency` rather than relying on that hoisting silently). A grep for the string `import.meta.main` also matches inside a comment or a string literal — including this gate's own header, and the assertion in `capture-mcp-tools-list.test.mjs` that a producer's *source text* excludes it. Parsing means those never false-positive, while `import.meta?.main`, a line break inside the expression, `!import.meta.main`, `import.meta["main"]`, and `const { main } = import.meta` (with or without a rename) all still red — none of those survive a naive regex.
- **File set is derived, not listed.** A directory walk over `scripts/**` and `contract/**`, filtered to `.mjs`/`.js`. Zero files discovered is a hard failure. The regression suite cross-checks the walk against an independently-derived `find`-based listing, so the floor moves with the tree — a future narrowing of the walk (e.g. it stops recursing) reds a set-mismatch instead of being satisfied forever by today's count.
- **Subsumed the existing single-gate check.** `capture-mcp-tools-list.test.mjs` (~137-146) used to assert *for one file* that its source text excludes `import.meta.main`. That's now covered — more completely — by the repo-wide gate, so I dropped the negative half. I kept the *positive* half (that this producer's guard really is the `process.argv[1]` vs. `import.meta.url` comparison, not just "not the bad string") because that's a different property the repo-wide gate doesn't assert: it proves absence of the bad pattern everywhere, not presence of the good one in any specific file. Two checks asserting the same negative in different places was the shape D5 warns about; one check asserting a positive and one asserting a negative, at different granularities, is not.

## Break-on-purpose matrix

| Case | Result |
| --- | --- |
| Add `if (import.meta.main)` to a scratch file under `scripts/` | Gate reds, naming the file and line: `scripts/f1481-scratch-bad.mjs:3 — import.meta.main` |
| Same scratch dir, file only *mentions* `import.meta.main` in a comment and a string literal | Gate stays green (`No bare import.meta.main in 65 file(s)...`) |
| `probe-twin-endpoints.mjs`'s manifest set to `{}` (zero twins) | `runGate()` throws: `... declares zero twins — refusing to report a pass having probed nothing` |
| Manifest declares a twin with `probes: []` | Throws on the zero-probes floor (and separately reds via `unprobed-endpoint` findings for every declared tool) |
| Point `lint-no-bare-import-meta-main`'s scan at an empty directory (no `scripts/`/`contract/`) | `scanRepo()` throws: `... discovered zero files ... refusing to report a pass having scanned nothing` |

Also verified live:
- Direct invocation: `node scripts/probe-twin-endpoints.mjs` — 5 twins, 128 probes, OK.
- **Symlinked invocation** (the case that broke F-1353's first attempt): symlinked the repo root under `/tmp`, ran `node /tmp/.../repo-link/scripts/probe-twin-endpoints.mjs` — same clean result, confirming the realpath-both-sides fix survives a symlinked checkout.
- `actionlint .github/workflows/ci.yml` — clean.
- `npm run lint:code-health`, `npm run lint:dead-code` (knip, `dependencies: "error"` — confirms the new `acorn` devDependency is actually used, not orphaned), `node scripts/check-packages-scripts-wired.mjs` — all green.
- `npm run typecheck`, `npm run test:contract`, `npm run probe:twins`, `node scripts/probe-twin-endpoints.test.mjs`, `node scripts/lint-no-bare-import-meta-main.test.mjs`, `node scripts/capture-mcp-tools-list.test.mjs` — all green.

## Anything to challenge

- `acorn` is a new explicit root `devDependency`. It was already present in the resolved tree (transitive, via `tsup`), so `npm ci` needed no re-resolution beyond the `devDependencies` line itself — I edited `package-lock.json` surgically rather than regenerating it wholesale, because a full `npm install --package-lock-only` also picked up unrelated version-field drift already present between several packages' `package.json` and the committed lockfile (pre-existing on `origin/main`, not something I introduced or want to bundle into this diff).
- The gate is wired into the heavy block, gated on `steps.scope.outputs.heavy == 'true'`. Its subject only exists under `scripts/**`/`contract/**`, and the scope step already classifies any diff touching those paths as heavy — so it isn't path-filtered away from anything that could actually trip it, but it also doesn't run on every docs-only PR. Worth a second look if that reasoning doesn't hold up.
- I did not attempt to make the destructuring/computed-access detection exhaustive beyond the forms named in the ticket (optional chaining, negation, destructuring, computed string access) — e.g. `import.meta["ma" + "in"]` would not be caught. That's an AST-can't-help-you case (real static analysis of runtime string concatenation), and I judged it out of scope for "cannot hide behind formatting."

Linear: F-1481. Not merged, not marked done.


GaganSD:
## Review of record — F-1481

Three passes: the `review` skill, the `code-review` skill (pinned to this worktree), and a hand-rolled adversarial pass in place of `codex review` (unavailable, usage limit to Sep 11 — not faked, substituted and described below). Both skills independently found the same top item. I verified every stated cause rather than trusting the conclusions; one of the skills' causes was wrong and is corrected below.

Findings are ranked. Everything blocking is **fixed on the branch**.

---

### 1. BLOCKING (fixed) — the gate's scope was narrower than the defect, and six live instances sat outside it

The gate watched `scripts/**` and `contract/**`. At that moment **six bundled examples were live instances of the exact defect**:

| File | Line | `engines` |
| --- | --- | --- |
| `examples/minimal-viktor/src/index.ts` | 258 | `>=24` |
| `examples/merge-agent/src/index.ts` | 177 | `>=24` |
| `examples/gmail-retry-notify/src/index.ts` | 173 | `>=24` |
| `examples/pr-summary-agent/src/index.ts` | 424 | `>=24` |
| `examples/pr-summary-review/src/index.ts` | 442 | `>=24` |
| `examples/triage-agent/src/index.ts` | 424 | `>=24` |

All six guarded their launch on bare `import.meta.main`. On Node 24.0.0/24.0.1/24.0.2/24.1.0 `npm start` in any of them prints nothing and exits 0 having run no agent. `examples/support-triage` had already been fixed to the sanctioned form and its comment documents this failure verbatim — the other six were simply never done.

This is exactly the pattern the ticket is about: F-1353 fixed `contract/run.mjs`, the same bug survived ten lines away under `scripts/`, and this gate would have declared the class closed while six instances shipped one directory over. `F-1481`'s own **Done when** is repo-wide ("No entry guard in the repo uses bare `import.meta.main`"), so these were in scope, not a follow-up.

**Why a JS-only parser could never have caught them, and why acorn is now gone.** The live instances are `.ts`. acorn cannot parse a TypeScript file without a plugin, so extending the roots alone would have turned every example into an "unparseable" finding. Swapped to **`typescript`**, which parses `.ts`/`.mts`/`.tsx` and plain `.mjs`/`.js` through one entry point with no plugin, and is already a direct devDependency of all ten workspaces at a uniform `^5.9.3`. Net effect: **one fewer dependency than the PR started with**, and the parser can actually read the gate's subject. I agree with the original instinct that AST beats regex here; the disagreement is only about which parser.

Scope is now `no-eval-in-oss.mjs`'s root list (`scripts`, `cli/src`, `cli/scripts`, `packages`) plus `contract` and `examples` — 744 files, 1.1s. `packages/*/scripts/**` and `cli/scripts/**` (raised in review) are **in scope now**; neither contained a live instance, but both are covered rather than argued away. Verified nothing escapes: the only source extensions under those roots are `.ts` (668) and `.mjs` (76), both scanned; no out-of-scope top-level directory (`assets`, `config`, `docs`, `fixtures`, `skills`) contains any JS/TS; and an independent repo-wide grep now finds `import.meta.main` only in comments, test fixtures and generator strings — zero live code.

Also fixed, all silent-skip shapes in a gate whose premise is that skips must not read as passes:

- **Aggregate zero-file floor → per root.** `scripts/` alone satisfied it, so relocating `contract/` would have left it unscanned behind a green gate.
- **Symlinks were skipped entirely.** A dirent's `isDirectory()`/`isFile()` are *both* false for a symlink, so a symlinked script or subdirectory was never scanned. `statSync` follows it now. The test's `find` cross-check shared the identical blind spot (no `-L`) so it could not have caught this; it now uses `-L`.
- **No pruning.** Without `node_modules`/`dist`/`build`/`coverage` pruning the walk finds ~19,600 files and parses third-party CJS as ESM.

Break-on-purpose, in the location the original gate could not see:

```
$ node scripts/lint-no-bare-import-meta-main.mjs
Bare `import.meta.main` found in 1 place(s):
  examples/minimal-viktor/src/index.ts:272 — import.meta.main
$ echo $?
1
```

The regression suite now asserts **per-root coverage** and that a bundled example's entry module is in scope, so the scan cannot silently narrow back.

---

### 2. Lockfile verdict — the hand edit was correct; verified from cold

This was the other thing flagged for scrutiny, and it holds up.

- **Cold `npm ci`:** `rm -rf node_modules && npm ci` → exit 0, installs the expected parser, gate green on the cold tree. Repeated after the typescript swap and after the rebase.
- **`npm ci` does not touch the lockfile:** byte-identical before/after (`cmp`).
- **No duplicate copy:** exactly one top-level `node_modules/<parser>` lock entry, correctly `dev: true`. No second nested copy.
- **The hand edit is byte-identical to what npm generates.** `npm install` reproduces exactly the one-line `devDependencies` change — nothing else about it is hand-shaped.
- **The author's claim about "unrelated version drift" is true and I verified it against `origin/main`.** The lockfile records stale versions for five workspaces (`cli` 0.23.12 vs 0.23.21, `twin-github` 0.10.4 vs 0.10.7, `twin-linear`/`twin-slack` 0.4.0 vs 0.4.1, `twin-stripe` 0.4.6 vs 0.4.7). That drift is **pre-existing on `main`** and unrelated to this PR, so declining to bundle a wholesale regeneration was the right call. Worth its own issue — a lockfile that disagrees with five workspace manifests is a latent surprise for the next person who runs `npm install`.

Verdict: **clean.** The one thing I'd change — using a parser that can read `.ts` — was a scope problem, not a lockfile problem.

---

### 3. AGENTS.md claimed the wrong CI block (fixed)

The row said the gate runs in "`ci.yml`'s **always-on block**". It runs in the **heavy** block, gated on `steps.scope.outputs.heavy == 'true'`, and it must — it needs `npm ci` for the parser. This repo treats the always-on/heavy split as load-bearing, and the check it *replaced* genuinely was always-on. Row now states the heavy block, the real scope, the parser choice, and the documented non-goal.

**Where it runs — verified, not assumed:**
- `scripts/` **is** in the heavy scope regex (`^(packages/|cli/|examples/|scripts/|contract/|config/|knip\.json|package(-lock)?\.json$|\.github/workflows/ci\.yml$)`), as is every other root the gate now scans. The regression suite now asserts this coupling, so widening the scan without widening the regex reds.
- The step runs under `set -euo pipefail`, has no `continue-on-error`, and is not disabled.
- **Confirmed from the job log of the pre-review run**, not inferred: `No bare import.meta.main in 64 file(s) under scripts, contract.` followed by `lint-no-bare-import-meta-main: all assertions passed.`

Note the subsumption side effect: the dropped check in `capture-mcp-tools-list.test.mjs` ran always-on, the new one runs heavy. Not a hole — any edit to that producer is a `scripts/` change, which selects heavy. Airtight, but it is a real change in *when* the property is enforced and the row now says so.

---

### 4. The regression suite's CI-wiring assertions verified nothing (fixed)

`ci.includes("npm run lint:import-meta-main")` passes with the line **commented out**, moved under `if: false`, or moved into a job that never runs. And `/set -euo pipefail/.test(ci)` matched the **scope step at the top of the file**, not the block the gate lives in — so the assertion labelled "the step block the gate lives in runs under a failing shell mode" was checking a different step entirely. (This is one of the two skills' findings whose cause I confirmed by reading rather than accepting.)

Now: locate the gate's own step, assert exactly one *uncommented* occurrence, a failing shell mode in that step, no `continue-on-error`, not disabled.

---

### 5. Detector completeness matrix

Everything asked about, plus what I added. `import.meta["main"]` **is** caught — statically resolvable and a plausible accident, as expected.

| Form | Result |
| --- | --- |
| `import.meta.main` | caught |
| `!import.meta.main` | caught |
| `import.meta?.main` | caught |
| `import.meta["main"]` | **caught** |
| ``import.meta[`main`]`` | caught *(added — was missed)* |
| split across lines / wrapped in parens | caught |
| inside a ternary | caught |
| inside an `&&` chain | caught |
| `Boolean(import.meta.main)` | caught |
| `const { main } = import.meta` | caught |
| `const { main: isMain } = import.meta` | caught |
| `const { ["main"]: m } = import.meta` | caught *(fixed — `ComputedPropertyName` was not unwrapped)* |
| `const { ...rest } = import.meta` | caught *(added — hands the whole object over, so `rest.main` is reachable)* |
| `({ main } = import.meta)` / `({ ...rest } = ...)` | caught |
| bare guard in a **`.ts`** file | caught *(the case that mattered)* |
| — comment mentioning it | no false positive |
| — string literal | no false positive |
| — template literal | no false positive |
| — array of source lines a generator emits | no false positive |
| — `import.meta.url`, `import.meta.resolve`, `config.main` | no false positive |
| — `import.meta["ma" + "in"]` | not caught, **documented non-goal** (agreed) |

Two aliasing forms remain uncaught and I judged them not worth chasing: `const meta = import.meta; if (meta.main)` and passing `import.meta` to a function. Both need real dataflow analysis, neither is reachable by accident, and the honest place for that limit is the AGENTS.md row rather than a half-implementation.

**Unparseable files are now their own category.** `typescript` *recovers* from syntax errors instead of throwing, so without an explicit `parseDiagnostics` read an unreadable file would scan clean — a skip reading as a pass. Previously a parse failure was pushed into `findings` and printed under the "Bare `import.meta.main` found" header with advice to fix an entry guard that doesn't exist.

---

### 6. Both floors break for real

Verified by execution, not by reading. Neither is the "counters that move together" shape:

```
zero files scanned  → THREW: lint:import-meta-main discovered zero files under contract —
                      refusing to report a pass having scanned nothing
zero twins          → THREW: … declares zero twins — refusing to report a pass having probed nothing
zero probes         → THREW: … declares 1 twin(s) but zero probes in total …
```

The probe floors read the *manifest*, while per-probe execution is asserted separately (`!call` and `call.failed` both red as `driver-error` in `evaluateTwinProbeRun`), so "declared but never executed" is genuinely covered rather than assumed — I checked, because F-1163's bug was precisely counting declarations instead of invocations.

**One ordering bug fixed:** the `probeCount === 0` throw sat *above* the findings report, so a manifest whose twins declare no probes replaced a per-endpoint `unprobed-endpoint` list with a bare stack trace naming nothing — the louder failure was the less informative one. Moved below, where a floor on the pass path belongs.

---

### 7. Entry guard and the symlink reproduction

Reproduced the author's symlink test independently, through a **double** symlink (`/tmp` → `/private/tmp`, plus my own link):

- `probe-twin-endpoints.mjs` via the symlinked path: 5 twins, 128 probes, clean.
- The new gate via the symlinked path: identical result to the real path.
- **The guard-miss canary really throws** — invoked through a same-basename shim: `Error: probe-twin-endpoints.mjs entry guard did not fire for /private/tmp/… (expected /Users/…)`. A miss while invoked as the script throws, it does not exit 0.
- **Demonstrated one-sided realpath would have missed:** raw `resolve(argv[1])` = `/tmp/f1481-link/scripts/probe-twin-endpoints.mjs` vs realpathed self `/Users/…` — mismatch, i.e. the F-1353 trap.
- And for the six examples I changed: `npx tsx src/index.ts` through the symlinked path reached `main()` and threw on a missing env var, rather than silently exiting 0.

`realpathSync` on both sides: confirmed. The examples match `support-triage`'s already-merged form exactly (no throw canary there — that is the established precedent for example code, and `smoke:examples` now FAILs a do-nothing exit 0 independently).

---

### 8. Subsumption decision — sound, and strictly stronger

The dropped assertion was `!producerText.includes("import.meta.main")` for `scripts/capture-mcp-tools-list.mjs`. Verified: that file **is** in the repo-wide gate's file set, and the new gate is strictly stronger for real code —

| | old substring check | new AST gate |
| --- | --- | --- |
| `if (import.meta.main)` | catches | catches |
| `if (import.meta?.main)` | **misses** | catches |
| `if (import.meta["main"])` | **misses** | catches |

The only thing lost is false positives on comments and strings. Keeping the *positive* half (that this producer's guard really is the argv/`import.meta.url` comparison) is the right call — it asserts a different property the repo-wide gate does not. Nothing regressed.

---

### 9. Not fixed here — two pre-existing issues that each need their own ticket

**(a) The same silent-skip class via a different mechanism — ten one-sided entry guards.** The gate bans `import.meta.main`, but the *other* half of the failure mode the header itself describes is ungated. Ten scripts compare `resolve(process.argv[1])` against `fileURLToPath(import.meta.url)` with **no realpath on either side**: `capture-mcp-tools-list.mjs`, `no-eval-in-oss.mjs`, `no-catch-and-continue.mjs`, `no-native-modules.mjs`, `check-workspace-pins-match-workspace.mjs`, `check-plugin-manifest-lists-every-skill.mjs`, `ci/release-alarm.mjs`, `packages/wire/scripts/emit-trace-contract.mjs`, plus `lint-task-class.mjs` and `lint-task-format-doc.mjs` which compare a *basename* (`import.meta.url.endsWith(basename(argv[1]))` — matches any file with the same name). Demonstrated live:

```
$ node scripts/no-eval-in-oss.mjs                   # real path
no-eval-in-oss gate passed — pome-twins is capture-only.
$ node /tmp/f1481-link/scripts/no-eval-in-oss.mjs   # symlinked path
$ echo $?
0        # ← silence + exit 0 = checked NOTHING
```

Same for `capture-mcp-tools-list.mjs --check --offline`, an always-on gate. GitHub's checkout is not symlinked so CI is unaffected today; the exposure is local runs and symlinked worktrees — the same exposure profile as F-1481 itself. Arguably higher severity than F-1481's original subject because these *are* CI gates. Deliberately not fixed here: it is a different mechanism, ten files, and it deserves its own issue rather than being smuggled into this diff.

**(b) `cli/scripts/make-unwired-fixture.mjs` is broken on `main` and wired nowhere.** I verified this against a pristine `git archive main` tree — it already exits 2 (`could not find the withPome() startup block`) *before* my change, because its exact-text replacements drifted from `examples/triage-agent` (it searches for "secret-path unit test", the example says "auth-token unit test"). At least two replacements are stale. It is referenced by nothing — no workflow, no `package.json` script, no doc — and `check-packages-scripts-wired.mjs`'s denominator is `packages/*` only, so `cli/scripts/*` has no wiring audit and this script rotted undetected. That is a textbook D5 finding.

I did update its guard replacement (it transforms the example by exact text match, so it had to move with the guard) and made the generated fixture inherit the sanctioned guard rather than emitting the banned one — it writes to a tmpdir where no gate would ever scan it. It is still broken at the *next* stale replacement, exactly as it was on `main`. No regression from this PR; needs its own ticket.

---

### Minor / accepted

- `scanRepo` parses each file twice (once for diagnostics, once for hits). 744 files in 1.1s, once per CI run — measured, not worth the churn.
- The version-bump gate caught a real omission of mine mid-review (editing a publish-relevant `cli/` path without bumping `cli/package.json`). Bumped to 0.23.23. Working as intended.

### Verification run

Cold `npm ci` (twice, incl. post-rebase) · `npm run build` · `lint:import-meta-main` (744 files) · its test suite · `probe-twin-endpoints.test.mjs` · `probe:twins` (128 probes) · `capture-mcp-tools-list.test.mjs` · `test:contract` (110 pass, 0 fail) · `typecheck` · `typecheck:examples` (8/8 clean) · `smoke:examples` (**8/8 reached real work — 5 alive at the settle, 3 REACHED-OUTBOUND, none a do-nothing exit 0**) · `probe:examples` (20/20 seeds) · `lint:code-health` · `lint:dead-code` · `check-packages-scripts-wired.mjs` · `check-copy-markers.mjs` · `actionlint`. Rebased onto `origin/main` and re-verified the full battery after the rebase, not just the merge.

Not merged. Linear not marked done.

GaganSD:
### CI confirmation (post-rebase)

Rebased onto `origin/main` (`04ccd1f`, F-1471) and re-verified — a clean rebase is not a working rebase, and `main` had touched both `ci.yml` and `AGENTS.md`, the two files this PR also changes. F-1471's scheduled-alarm work, this gate's step, and both AGENTS.md rows all survived intact.

Job-log evidence on the rebased tip `85b5cdd` (run 31621841895), which is the number that matters for the scope fix:

```
Running heavy suite (package/cli/script/lockfile/ci changes).
No bare `import.meta.main` in 746 file(s) under scripts, contract, cli/src, cli/scripts, packages, examples.
lint-no-bare-import-meta-main: all assertions passed.
```

**746, up from 64.** The file count moved on its own between runs (744 → 746 as `main` added scripts), which is the property the per-root floor and the `find` cross-check exist to preserve — the coverage denominator tracks the tree instead of being frozen at today's count.

`typecheck-test: pass`. Full battery re-run after the rebase: cold `npm ci`, build, gate + tests, `probe:twins` (128 probes), `capture-mcp-tools-list.test.mjs`, `typecheck:examples` (8/8), knip, code-health, `check-packages-scripts-wired`, version-bump gate, `actionlint`.

One note for the record: a concurrent actor force-pushed a content-equivalent rebase of this same branch while I was verifying mine, so `--force-with-lease` correctly rejected my push. I diagnosed read-only first and confirmed **identical tree hashes** (`b5495eb…`) and an identical merge base, then adopted the remote SHAs rather than force-pushing a duplicate. No commits lost on either side, and the tree CI validated is byte-identical to the one I ran the battery against.

Still not merged; Linear not marked done.

PR #391, opened 2026-08-12
examples/support-triage's published adapter pin now has a registry watch (F-1483)

GaganSD:
## Where the check lives, and why

`examples/*` deliberately pins published `@pome-sh/*` versions (`examples/support-triage`'s README documents `npx degit`-fetchability, which copies only that subtree, so a `file:` link out of the tree would break its standalone `npm install`). `check-workspace-pins-match-workspace.mjs` runs OFFLINE before `npm ci` and rules `examples/*` out of scope for that exact reason — its rule ("must resolve to the workspace") is wrong for a pin that is published on purpose.

The right rule needs the registry: re-pin once the sibling workspace version publishes, skip while it has not. That can't live in the offline gate, and the task says not to invent a second CI mechanism — so it lives in `scripts/check-example-pins-published.mjs`, wired into `scripts/typecheck-examples.mjs` right where every example is already `npm ci`'d against the real registry.

- `discoverExampleRegistryPins()` walks `examples/*/package.json` fresh each run (no hand-kept example list) and picks out every `@pome-sh/*` pin that (a) is an exact semver and (b) names a workspace sibling — `"*"`/`file:`/range pins are a different, already-covered concern.
- `checkExamplePinsPublished()` takes an injectable `npmView` so the regression suite needs no network.
- `defaultNpmView()` shells to `npm view <name>@<version> version` and classifies the result exactly like `scripts/ci/decide-publish.sh` already does: exit 0 → published; exit non-zero with `E404` in stderr → genuinely unpublished (skip); anything else (401/5xx/timeout/rate-limit) → hard failure, never downgraded to a skip.
- Discovering zero eligible pins throws rather than reporting a pass having checked nothing.
- `loadWorkspaceMembers()` was extracted out of `check-workspace-pins-match-workspace.mjs` so both gates derive the sibling-name → workspace-version map the same way instead of walking root `workspaces` twice by separate rules.

## The 404-vs-error distinction

A 404 on `npm view <pkg>@<version>` means that specific version doesn't exist on the registry — the normal cloud-first state where `packages/` has bumped and merged but its release job hasn't published yet. That's a named, counted skip, printed distinctly, never folded into "matched" or a silent pass.

Anything else — auth failure, network error, timeout, rate limit — is NOT "unpublished". Treating it that way would let a transient registry outage wave through a real drift undetected. This is the same three-way split `decide-publish.sh` already draws for the release lane; this gate matches its stance rather than inventing a new one.

## Before/after dependency tree

The live drift named in the ticket (pin `0.3.1` against a published, `@pome-sh/shared-types`-free workspace `0.3.3`) turned out to already be fixed on `main` (#382 bumped the pin in the same PR that closed F-1231's `cli/` gap). Re-verified here with a real install:

```
$ cd examples/support-triage && npm ci
$ npm ls @pome-sh/shared-types
@pome-sh/support-triage-local-example@0.1.0 ...
└── (empty)

$ npm ls @pome-sh/adapter-claude-sdk
@pome-sh/support-triage-local-example@0.1.0 ...
└── @pome-sh/adapter-claude-sdk@0.3.3
```

Before (published `0.3.1`, for reference — not what's installed today): `npm view @pome-sh/adapter-claude-sdk@0.3.1 dependencies` includes `@pome-sh/shared-types: "0.14.1"`. `0.3.3`'s dependency list has no `@pome-sh/*` entry at all.

This PR's own contribution is closing the *watch* — the part of F-1483/F-1231 that was still open — not the pin value itself.

## Break-on-purpose evidence

All three demonstrated locally with the working tree reverted after each:

**1. Stale pin against a published workspace version reds, naming example/pin/workspace version:**
```
❌ published pin DRIFT in 1 example(s):
  examples/support-triage (dependencies.@pome-sh/adapter-claude-sdk): pins 0.3.1, but the workspace sibling is 0.3.3 and 0.3.3 is published. Re-pin to 0.3.3.
```

**2. Unpublished workspace version (bumped to `9.9.9`, which doesn't exist on the registry) is a counted, visible skip — not a pass, not a fail:**
```
⚠️  skipped 1 pin(s) — sibling workspace version not yet published:
  examples/support-triage (dependencies.@pome-sh/adapter-claude-sdk): workspace is 9.9.9, pin is 0.3.3
example pin↔registry parity: 0 matched, 1 skipped (unpublished), 0 drifted, 0 errored (of 1 checked).
```

**3. A registry error that isn't 404 (mocked `npm` on PATH answering `E401`) is a hard failure, never a skip:**
```
❌ registry lookup FAILED for 1 pin(s) (not a skip — a real error):
  examples/support-triage (dependencies.@pome-sh/adapter-claude-sdk@0.3.3): npm error code E401 ...
example pin↔registry parity: 0 matched, 0 skipped (unpublished), 0 drifted, 1 errored (of 1 checked).
```

## Verification

- `check-example-pins-published.test.mjs` (13 cases, pure functions, no network): discovery scoping (exact pin + sibling only), the empty-discovery throw, matched/stale/skip/error outcomes, and `defaultNpmView`'s own E404-vs-other classification against a mocked `npm` on `PATH`.
- `check-workspace-pins-match-workspace.test.mjs` still green after the `loadWorkspaceMembers()` extraction.
- `npm run typecheck:examples` green end to end, including the new parity report.
- `npm run smoke:examples`, `npm run lint:code-health`, `npm run lint:dead-code`, `node scripts/check-packages-scripts-wired.mjs`, `npm run lint:import-meta-main`, `actionlint -color` all green.
- `node scripts/ci/check-version-bump-required.mjs` — no publish-relevant path touched, no bump owed.

## Judgment calls a reviewer should challenge

- The gate only checks pins that are an **exact semver** naming a **workspace sibling**. A `file:`/range/`*` pin in an example is left to whatever already governs it (nothing does, today, for `file:`-linked examples — deliberately, since those resolve from source and can't drift the way a published pin can). If a future example pins a *range* against a published package on purpose, this gate is silent on it.
- Discovering zero eligible pins throws. Today that only happens if `examples/support-triage`'s pin moves to a range or a `file:` link — at which point this gate would have nothing left to check and its own regression suite (case 4) would need conscious revisiting, not just deletion.
- The check runs inside `typecheck-examples.mjs` rather than as a standalone `gate:*`/`check:*` npm script, per the instruction not to add a second mechanism. That means it has no independent `npm run` entry point of its own — its only way to run is via `typecheck:examples`, and `check-packages-scripts-wired.mjs` never sees it as a package script (it's a root `scripts/` file, out of that gate's `packages/*` denominator either way).


GaganSD:
## Review of record (F-1483) — fixes pushed in `079d4dd`

Three passes: the `review` skill, the `code-review` skill pinned to this worktree, and — since `codex review` is unavailable (usage limit to Sep 11) — a substitute adversarial pass of throwaway probes against the real tree, each reverted after use. Every finding below was verified by running code, not by reading it. Two of the machine findings had **wrong stated causes** and are recorded as refuted.

### 1. The live defect (ticket item 4) — independently confirmed done

Not taken on trust. Real `npm ci` in `examples/support-triage`, then:

```
$ npm ls @pome-sh/shared-types
@pome-sh/support-triage-local-example@0.1.0
└── (empty)

$ npm ls --all | grep pome-sh
└─┬ @pome-sh/adapter-claude-sdk@0.3.3

$ find node_modules -type d -name shared-types     # (no output)
$ grep -c shared-types package-lock.json           # 0
```

`shared-types` is genuinely gone — not merely absent from the top level, but absent as a directory anywhere under `node_modules` and absent from the lockfile, so it is not reachable as a transitive, optional, or peer dep of any other `@pome-sh/*` package. Item 4 landed in #382; it is not quietly skipped. No live F-942 instance remains in the hero example.

### 2. HIGH — an ineligible pin went quiet while looking maintained (fixed)

This is the milestone's own failure shape, and the gate had it.

Discovery `continue`d past every non-exact pin. Probes against the real tree:

| probe | before | after |
|---|---|---|
| sole pin → `^0.3.3` | throws (zero-pins floor) | reds as unwatchable |
| sole pin → `^0.3.3` **+ one other exact pin present** | **`ok=true`, silent, 1 matched** | **`ok=false`, names the pin** |
| sole pin → `file:` link | throws | throws |

The middle row is the hole. The gate reddened on a range pin only because `support-triage` happens to own the tree's sole exact pin, so the floor caught it *by arithmetic accident* — add one exact pin in any other example and the hero example's watch evaporates while the report prints a clean pass. The author's note that this would "need conscious suite revisiting" was optimistic: test case 2 actively **asserted** the silence as correct.

Fixed by classifying every `@pome-sh/*` dep that has a workspace sibling and reporting all three classes: `exact` is checked against the registry; `file:`/`link:` is counted-but-out-of-scope (it cannot reach a registry, so it cannot swap in a stale artifact); anything else — range, `"*"`, dist-tag, `npm:` — resolves *from* the registry with no single version to compare and **reds as unwatchable**. "Cannot be compared" no longer reads as "exempt".

### 3. HIGH — the floor I first wrote reintroduced the same hole one case over (fixed)

Caught by `code-review` on my own change, verified by probe: widening the floor to `exact + linked + unwatchable === 0` let an all-`file:` `examples/` tree return `ok=true` after **zero registry calls**. Converting `support-triage` to a `file:` link is plausible — three other examples already are one — and it would have broken the example's documented `npx degit` install *and* silently emptied the gate. The floor now counts **exact pins only**.

### 4. MEDIUM — the parity check was unreachable behind another check's failure (fixed)

`reportExamplePinParity()` sat after `if (failures.length > 0) process.exit(1)`, so an unrelated `tsc` error in any of the eight examples hid a real published-pin drift until someone fixed the typecheck and CI came round again. It reads manifests and calls the registry; it needs none of the installs to have succeeded. Both results are now collected and folded into one exit.

### 5. MEDIUM — the floor's only test passed on the wrong error (fixed)

Test 4's fixture wrote `workspaces: ["packages/*"]` over an **empty** `packages/`, so `loadWorkspaceMembers` threw `no workspace manifests found` before the floor was ever reached. The bare `catch {}` accepted it, so the floor guard had **zero** coverage. It now uses a real fixture and asserts on the message.

### 6. 404-vs-error robustness — verdict: sound, with one real gap (fixed)

Probed with mocked `npm` on `PATH`, all reverted:

| condition | classification |
|---|---|
| `E404` | `unpublished` ✅ |
| `E401` / `E429` rate limit / `E500` | `error` ✅ |
| exit≠0 with **empty stderr** | `error` ✅ |
| `npm` **not on PATH** | `error` (`spawnSync npm ENOENT`) ✅ |
| `E500` whose text embeds `E404` in a URL | `error` ✅ (`\b` fails after a word char) |
| **registry accepts then stalls** | **hung forever** ❌ |

The literal-`E404` dependency is acceptable: npm error codes are stable machine identifiers with no i18n, and this deliberately matches `decide-publish.sh`'s existing `grep -q "E404"` stance rather than inventing a second one. Not worth diverging.

The hang was real — no `timeout`, and npm's own `fetch-timeout` defaults to five minutes *per attempt* times its retries, so a degraded registry stalled the heavy job instead of answering. Added `timeout: 60_000`; a killed call has no `E404` in stderr so it lands on the conservative side.

Also added **3 attempts with backoff**: "any non-E404 answer is a hard failure" puts the *required* `typecheck-test` behind a third-party registry, and one transient 5xx would red it. `ci.yml`'s own actionlint install carries a five-attempt loop with a comment recording three 503s during this PR's review — the same risk, live today. `E404` short-circuits on attempt 1, so the ordinary unpublished path pays nothing (test 9g asserts the call count).

### 7. LOW — two classification traps (fixed)

- `EXACT_VERSION = /^\d+\.\d+\.\d+$/` rejected prereleases, so pinning `0.4.0-rc.1` against workspace `0.4.0-rc.1` reddened as unwatchable with advice naming the value it already had — an unfixable red. Now accepts prerelease/build metadata.
- `WORKSPACE_LINK` missed a protocol-less `../foo`, which npm treats as a directory dep exactly like `file:../foo`. It reddened with the message that it "resolves from the registry" — false. Widened.
- The unwatchable remediation text offered "or link it with `file:`", which for `support-triage` is precisely the change this file's own header says breaks its `npx degit` install. Reworded.

### 8. REFUTED — `overrides` is not an escape hatch

`review` claimed an `examples/*` `overrides` entry could silently retarget the adapter to a stale published version, unseen by both gates, and noted (correctly) that four examples ship live `overrides` blocks. The blocks exist — but they target `undici` and `@modelcontextprotocol/sdk`, not `@pome-sh/*`, and the mechanism does not work:

```
npm error code EOVERRIDE
npm error Override for @pome-sh/adapter-claude-sdk@file:../src conflicts with direct dependency
```

npm refuses an override that conflicts with a direct dependency, so an example cannot silently retarget its own `@pome-sh/*` dep. Correct finding, wrong cause. Documented as out of scope rather than scanned.

### 9. ACCEPTED, not a defect — the post-release red window

`review` flagged that after every adapter release, every heavy CI run reds on the example's now-stale pin until a re-pin PR lands, and that the re-pin cannot ride along with the bump (the example's `npm ci` would fail against an unpublished version). That sequence is real, but it is the ticket's **specified** behaviour: *"Do: bump a workspace package without bumping the example's pin. Expect: red within one CI run once that version publishes."* F-1231 also rules out gating at the changeset level. Warn-only during the release window would re-open exactly the silence this PR closes. Left as designed; the cost is one follow-up re-pin PR per adapter release.

### 10. Verified, no change needed

- **`loadWorkspaceMembers()` is behaviour-identical.** The extraction moved the empty-result `throw` *into* the new function, which `findPinViolations` calls as its first statement — same order, same message, same throw. None of #382's three fixed holes regressed: the stale `optionalDependencies` pin, the `packages/*` pin on `@pome-sh/cli`, and the missing-`cli/package.json`-exiting-0 case are all still covered, and that suite's "the real repo is clean" case still passes.
- **The gate genuinely executes in CI — job log, not the checkmark.** Run [31623519015](https://github.com/pome-sh/digital-twins/actions/runs/31623519015), step *Install, build, test, contract, fidelity*, `17:39:20`–`17:39:21`:
  ```
  All 8 examples typechecked clean.
  === examples/* pin↔registry parity ===
  example pin↔registry parity: 1 matched, 0 skipped, 0 drifted, 0 errored (of 1 checked).
  ```
  Same log, scope step: `Running heavy suite (package/cli/script/lockfile/ci changes)`. `examples/` is in the heavy-selection regex, so a diff touching `examples/**` reliably selects heavy — the gate is not sitting in a skippable job.
- **Registry cost is 1 call.** One eligible pin today; the other four adapter deps are `file:` links, now visibly counted in the report so the denominator is honest (`3 file:/link: workspace link(s) out of scope`).
- **Entry guard:** no bare `import.meta.main` — the module is import-only with no top-level side effect. `lint:import-meta-main` (#390's AST gate) passes over 748 files.
- **AGENTS.md P8 row** updated to state what is now actually enforced: the three classes, the exact-only floor, the retry, the timeout, the run-even-if-typecheck-failed ordering, and `overrides` as out of scope with the `EOVERRIDE` reason.

### Suite

17 → 25 cases. New coverage: the range/`*`/dist-tag/`npm:` unwatchable set, an unwatchable pin reddening *while another exact pin matches cleanly* (the case the floor cannot catch), the floor asserted on its message, `file:`/`link:` counted, empty-stderr → error, transient 5xx retried, persistent error still an error after all attempts, `E404` short-circuiting, and a real-tree assertion that `examples/*` still carries a watched exact pin and nothing unwatchable.

Re-run green: the suite, `check-workspace-pins-match-workspace{,.test}.mjs`, `actionlint`, `lint:code-health`, `lint:dead-code`, `check-packages-scripts-wired.mjs`, `lint:import-meta-main`, `typecheck:examples`, `smoke:examples`, `check-version-bump-required.mjs`.

Not merging and not closing the Linear issue — that is the author's call.

GaganSD:
### CI after `079d4dd` — green, with job-log proof

Run [31625065707](https://github.com/pome-sh/digital-twins/actions/runs/31625065707), `typecheck-test` **pass**. All checks settled green (the earlier `gitleaks + trufflehog` red was the CDN degradation `ci.yml` already documents; it passes now).

Scope step: `Running heavy suite (package/cli/script/lockfile/ci changes).`

The **gate itself** executing against the live registry — heavy job, step *Install, build, test, contract, fidelity*, `17:58:05`:

```
All 8 examples typechecked clean.
=== examples/* pin↔registry parity ===
example pin↔registry parity: 1 matched, 0 skipped (unpublished), 0 drifted, 0 errored
  (of 1 exact pin(s) checked); 3 file:/link: workspace link(s) out of scope, 0 unwatchable.
```

The **suite** in the always-on block, `17:55:55` — 25/25, including the cases that lock the fixes:

```
✓ 3. pin "^0.3.3" is reported as unwatchable, not skipped
✓ 3. pin "latest" / "*" / "~0.3.1" / "0.3.x" / ">=0.3.0" / "npm:…" — same
✓ 4. zero exact pins throws, naming the floor it hit
✓ 9e. a transient 5xx is retried, not turned straight into a red
✓ 9f. a persistent error is still an error after all attempts
✓ 9g. an E404 short-circuits the retry budget
✓ 10. the real tree has 1 watched exact pin(s) and 0 unwatchable
All cases passed.
```

Not a checkmark — the gate's own output line, in the log, on this SHA. Branch is 0 commits behind `main`, so nothing was rebased.

**Verdict: mergeable.** Not merging, and leaving F-1483 open for the author.

GaganSD:
### Correction: `stripe` (twin-image) reds — infrastructure, not this diff

My previous comment said all checks settled green. That was premature — the `twin image` matrix was still running. It has now finished with `stripe` **fail**, and the other five twins (`github`, `gmail`, `linear`, `slack`, plus `detect-twins`/`wait-gates`) pass.

Cause, from the job log rather than inference — [job 94211060661](https://github.com/pome-sh/digital-twins/actions/runs/31625065921/job/94211060661), step *Generate image SBOM*:

```
[debug] http_download(url=https://github.com/anchore/syft/releases/v1.42.3)
[error] received HTTP status=503 for url='https://github.com/anchore/syft/releases/v1.42.3'
[error] unable to find tag=''
##[error]The process '/usr/bin/sh' failed with exit code 1
```

An HTTP **503 from the GitHub release CDN** — the same degradation `ci.yml`'s actionlint-install comment already records for today, naming *"a twin-image syft install (503)"* among the three it took out. The step order confirms it is not this diff: `docker build` succeeded, Trivy succeeded, and the failure lands in the syft download; `Push scanned image` / cosign steps are `skipped` on PRs. This diff touches `scripts/*.mjs` and `AGENTS.md` only and cannot influence a twin-stripe image build.

`stripe` is **not a required context** — the ruleset requires `typecheck-test`, `gitleaks + trufflehog`, `dependency review`, all three passing. Re-run the matrix job when the CDN recovers.

**Verdict unchanged: mergeable**, on the required checks and on the review. Still not merging and not closing F-1483.

#### One adjacent observation, out of scope here

`anchore/sbom-action` (`twin-image.yml:189`) fetches syft from that CDN with **no retry**, unlike the actionlint install in `ci.yml`, which was given an explicit five-attempt loop for exactly this failure and carries a comment explaining why curl's own `--retry` was not trusted. On a PR the blast radius is a non-required red. On `main` the cosign sign-and-attest steps do run, so the same 503 fails the image publish. That is a different gate and a different ticket — flagging it, not fixing it in this PR.

PR #392, opened 2026-08-12
cli/'s check scripts join the wired-scripts denominator (F-1476)

GaganSD:
## What's now true

`cli/` is now inside the same total partition `check-packages-scripts-wired.mjs` already held `packages/*` to: every non-lifecycle script in `cli/package.json` is wired, derived-covered, or carries its own `pome:unwired-ok(<name>): <reason>` marker — and, new for `cli/`, every file under `cli/scripts/**` that is neither the invoked file of a declared script nor imported by a sibling file gets the same treatment, keyed by its relative path instead of a script name.

## Per-script verdict

| Script | Verdict | Evidence |
| --- | --- | --- |
| `gate:no-eval` (`node ../scripts/no-eval-in-oss.mjs`) | **Deleted** | Byte-identical to the root `gate:no-eval`, which already runs in `ci.yml`'s heavy block and scans `cli/src`, `cli/scripts`, `packages`, `scripts` regardless of cwd (`resolve(dirname(script), "..")`, not `process.cwd()`). Redundant duplicate. |
| `gate:no-native` (`node ../scripts/no-native-modules.mjs --root .`) | **Deleted** | Also redundant with the root gate (single workspace lockfile now covers `cli`'s prod closure) **and independently broken**: `--root .` from `cli/` looks for `cli/package-lock.json`, which was removed when `cli/` joined the root workspace. Ran it — `Error: ENOENT: no such file or directory, open '.../cli/package-lock.json'`. |
| `test:e2e` (`vitest run test/e2e`) | **Deleted** | `vitest.config.ts` has no `include` filter, so plain `vitest run` (the lifecycle `test` script, already wired) already discovers and runs every `test/e2e/**/*.test.ts` file — confirmed with `vitest list` (all 20 e2e test cases show up with no filter). `test:e2e` was a strict subset with nothing distinguishing it. |
| `gate:recorder-overhead` (`tsx scripts/recorder-overhead-gate.ts`) | **Wired** | Was already running for real in `agent-trace-overhead-gate.yml`, just via a direct `npx tsx scripts/recorder-overhead-gate.ts` line invisible to this guard's `npm run <name> -w <pkg>` regex. Converted the workflow step to `npm run gate:recorder-overhead -w @pome-sh/cli`; ran it standalone, PASS. |

## The P8 question: is the row false, or the script?

**The rows are true; `cli/`'s own copies were false.** AGENTS.md's `gate:no-eval`/`gate:no-native` rows point at the two ROOT scripts, and both really do run repo-wide in `ci.yml`'s heavy block, covering `cli/src/**`/`cli/scripts/**` (no-eval) and the whole npm workspace's production dependency closure (no-native, since `cli` shares the one root `package-lock.json`). `cli/package.json`'s local `gate:no-eval`/`gate:no-native` entries were stale duplicates from before `cli/` joined the root workspace — one merely redundant, one actively broken (`ENOENT`). Deleting them doesn't remove any enforcement; the root gates already did the job.

## Extending the guard: two coordinator findings folded in

While in `cli/scripts/`, code review on a concurrent PR flagged two things squarely in this ticket's scope, folded in here:

1. **`cli/scripts/make-unwired-fixture.mjs` was broken and reached by nothing.** F-1481 (#390, which landed on `main` while this branch was rebasing) partially fixed it — one exact-text replacement it made against `examples/triage-agent/src/index.ts` had drifted (a stale comment), and #390 updated both the source comment and the script's replacement to match. Re-verified against the rebased tree: **it is still broken**, now failing at a *different* drifted replacement (`the preflight runner-mode healthz gate`) a few lines further down. Its own header explains why nobody has bothered chasing this: it exists to build a fixture for `pome-setup`'s manual wiring-acceptance run, and that Gen-1 skill was retired at F-859. Deleted rather than fixed a second time — no caller, no live purpose, and a scan-visible history of drifting silently.
2. **The guard's denominator now covers `cli/scripts/**` file-level entry points, not just `cli/package.json` scripts.** A file is a candidate needing wiring/a marker unless it is the invoked file of a declared script or is imported by a sibling file in the same tree (a library module, covered by whatever imports it — and if that importer is itself dead, IT shows up, which is the more actionable diagnosis). This closed `make-unwired-fixture.mjs`'s exact shape and also surfaced two real, already-running gates (`overhead-gate.ts`, `cas-adapter-acceptance.ts`) that were invisible to any script-name-based check. Both gained real `cli/package.json` script names (`gate:agent-trace-overhead`, `gate:cas-adapter-acceptance`) so `agent-trace-overhead-gate.yml` could call them through the standard `npm run <name> -w <pkg>` shape instead of a direct `tsx <path>` line — no second detection mechanism needed. Three remaining files (`overhead-bench-agent.ts` — a spawned workload agent referenced only via a resolved path; `write-build-info.mjs` — invoked by `tsup.config.ts`'s build hook, not a script; `eval-roundtrip.sh` — already documented as intentionally manual) got `pome:unwired-ok(<path>): <reason>` markers in the same syntax, keyed by path since they have no script name.

## Guard before/after

- **Before** (packages/* only, `origin/main`): 32 non-lifecycle scripts, 13 wired-or-exempt-and-covered, `cli/` invisible entirely.
- **After**: 40 entries (32 packages/* unchanged + 8 new `cli/` entries: 5 package.json scripts + 3 file-level entries), 17 covered/exempt, **0 failures**, `cli/` on the same total-partition terms as `packages/*`.

## Break-on-purpose

Added `"gate:foo": "node scripts/nonexistent-foo.mjs"` to `cli/package.json`:

```
1 package check script(s) are invoked by nothing:
  - @pome-sh/cli "gate:foo" (cli/package.json: "node scripts/nonexistent-foo.mjs") — no workflow or
    root aggregate reaches it via `npm run gate:foo ... -w @pome-sh/cli`, and its script file carries
    no `pome:unwired-ok(gate:foo): <reason>` marker.
```

Exit 1, naming it. Reverted.

Also verified an orphan `cli/scripts/` file with no script and no sibling import reds by path (new regression case 29), and that adding a `pome:unwired-ok(<path>)` marker clears it (case 30) — see `scripts/check-packages-scripts-wired.test.mjs` cases 25–34.

## Judgment calls for a reviewer to challenge

- **File-entry wiring is marker-only, no direct-invocation textual detector.** Rather than teach the guard to recognize `npx tsx <path>` as a second wiring shape, I converted the two real cases that used it (`overhead-gate.ts`, `cas-adapter-acceptance.ts`) into named `cli/package.json` scripts and rewired the workflow to call them the standard way. This keeps the guard to one detection mechanism total (declared-script-and-wired, or marker) but means any *future* direct-`tsx`-in-workflow shape will red until it's named as a script too — I think that's a feature (forces consistency), a reviewer might see it as a papercut.
- **"Imported by a sibling" exemption is import-graph-derived, not directory-based.** I considered just excluding `cli/scripts/lib/` by name (matching the existing `SKIP_DIR_NAMES` pattern), but derived it from actual relative imports instead so it generalizes past that one directory. It only follows relative specifiers one level (no transitive resolution needed here since the graph is shallow), and only resolves `.js`/`.mjs`/`.cjs` specifiers to a same-named `.ts`/`.tsx` file on disk — a `.jsx`-specifier project would need another candidate extension added.
- **`pome:unwired-ok(scripts/eval-roundtrip.sh)` and friends reformat existing honest prose into the exact marker syntax** rather than just trusting the prose that was already there — I don't think this is controversial, but it's a change to file content beyond the four originally-named gates.

## Not done / known staleness

`scripts/lint-no-bare-import-meta-main.mjs` and its test still mention `cli/scripts/make-unwired-fixture.mjs` by name as the illustrative example of "a generator that writes the banned string as data, not code" (F-1481/#390). Left as-is — the comment illustrates a still-valid pattern class and isn't a functional dependency (verified: that test's actual fixture is synthetic, not a read of the file), but it now names a deleted file.


GaganSD:
## Review of record (F-1476) — approved with fixes pushed

Three passes: the `review` skill, the `code-review` skill, and a substitute adversarial pass by hand (`codex review` is unavailable to Sep 11 and was **not** run or faked). Everything below was verified by running it, not read off the PR body.

### Per-deletion verdict, with path-set evidence

**`gate:no-eval` — deletion CORRECT, path sets identical (∅ diff).** Not "byte-identical scripts" — the same *file*. `cli`'s copy was `node ../scripts/no-eval-in-oss.mjs`, and that script derives its root as `resolve(dirname(fileURLToPath(import.meta.url)), "..")`, never `process.cwd()`, with a fixed `SCAN_DIRS = ["cli/src", "cli/scripts", "packages", "scripts"]`. Same file ⇒ same denylist, same allowlist, same prefix rules. Proved it rather than reasoning it — planted `cli/scripts/scorer.ts` and ran both invocations:

```
--- from repo ROOT:  ✗ cli/scripts/scorer.ts: file name matches denied eval-role stem "score*"
--- from cli/:       ✗ cli/scripts/scorer.ts: file name matches denied eval-role stem "score*"
```

Identical output from both cwds. No enforcement was narrowed, and AGENTS.md's P8 row stays true.

**`gate:no-native` — deletion CORRECT, and both justifications hold independently.** *Broken:* `--root .` from `cli/` resolves to `cli/package-lock.json`, which does not exist; ran it, it threw `ENOENT`. Note the direction — had it ever been wired it would have CRASHED, i.e. red. It was never a silent pass, it was simply never invoked. *Redundant:* root `gate:no-native` takes no `--root`, so it reads the single root `package-lock.json`. Checked that lockfile directly: it has a `cli` entry and **all 13** of cli's production deps resolve in it, `missing: []`. cli's closure is a strict subset of the root gate's. Superset, not a different set.

**`test:e2e` — genuinely dead.** `cli/vitest.config.ts` has no `include`, and `exclude` is only `node_modules/dist/runs`, so `vitest list` under the plain lifecycle `test` script discovers **21** `test/e2e/**` cases. Grepped the whole tree for `test:e2e`: zero invocations remain — no workflow, no runbook, no docs page. `cli/CONTRIBUTING.md` was the only mention and this PR updated it correctly.

**`make-unwired-fixture.mjs` — but three references survived, not two.** The PR body's "Not done" section names `lint-no-bare-import-meta-main.mjs` and its test. It missed **`AGENTS.md:126`**, which cited the file as the live illustrative example of a generator that writes the banned string as data. All three now describe the pattern class instead of naming a deleted file. Fixed.

### The forced `npm run` convention — verdict: acceptable, but it is a FALSE RED, not a blind spot

Tested it. A declared script whose file a workflow invokes directly:

```
1 package check script(s) are invoked by nothing:
  - @pome-sh/cli "gate:foo" (cli/package.json: "tsx scripts/foo.ts") — no workflow or
    root aggregate reaches it via `npm run gate:foo ... -w @pome-sh/cli`
```

It reds work that genuinely runs, and the guard does **not** also scan for the script path. I'm keeping the narrowness — one detection mechanism is what makes the marker path trustworthy, and the red is loud with a self-naming fix — but the PR framed this as invisibility and the guard header claimed "being reached means being one of (a)'s invoked files", which is false for this shape. Both the header and the AGENTS.md row now say plainly that it is a false red, with the upgrade path if the trade stops being worth it.

### Marker mechanism — F-1472's three defects, all absent on the new file-level half

Reconstructed each against a real orphan file entry:

| Defect | Probe | Result |
| --- | --- | --- |
| bare `pome:unwired-ok:`, `\s` scavenging the next line | `// pome:unwired-ok:` + `import fs...` | **reds** — regex requires `(<name>)` |
| named but blank reason | `// pome:unwired-ok(scripts/orphan.mjs):` | **reds** — `[ \t]*(\S…)` cannot cross a newline |
| `..` path escape, one package's marker exempting another's file | marker in `packages/beta/scripts/other.mjs` + `..`-keyed marker | **reds** — path comes from the walk, never from marker text |
| unnamed marker exempting every entry in the file | marker naming `scripts/other.mjs` inside `orphan.mjs` | **both** red |

The caller side is derived (walk + import graph + package.json commands), not listed. Confirmed.

### Floors

The 32→40 move is real (verified 40/17 before and after every change I made). But the `cli/` half had **no floor at all**: `findCheckScripts` hard-throws on a missing `packages/`, while `findCliPackageScripts`/`findCliOrphanFileEntries` returned `[]`, and case 34 asserted that as correct. Renaming `cli/` or `cli/scripts/` would have dropped eight entries and stayed **green** — coverage silently shrinking, which is D5's whole subject, and which `lint:import-meta-main` already treats as a hard failure per root. Added a floor derived from the root `workspaces` array npm itself reads: if it names `cli`, the cli/ half must be non-empty. Cases 36–37.

### Also fixed (five more, all found in the third pass)

1. **The guard's `LIFECYCLE_SCRIPTS` docstring still ended "`packages/*` is the denominator, so `cli/`'s own `gate:*` scripts are out of this gate's reach entirely (follow-up)"** — the exact thing this branch stopped being true. This milestone has already shipped three false comments; that's four.
2. **A commented-out sibling import certified its target as live.** `stripCommentLines` guards the corpus for precisely this reason but wasn't applied to the import scan, so `// import { x } from "./dead.js"` exempted `dead.js`. Case 35.
3. **`SCRIPT_FILE_RE` omitted `.mts`/`.cts`** — a file invisible to the denominator, the exact class this pass exists to close, and the sibling gate next door scans both.
4. **The import scan saw only `from`/`require`** — a side-effect `import "./x"`, a dynamic `await import("./x")`, and extensionless/`index` specifiers all left their target looking orphaned.
5. **`invokedFile`'s unanchored regex matched `tsconfig.js` inside `tsconfig.json`** and **derived `--check` coverage keyed on `pkgDir` alone** (`cli` collides with a future `packages/cli`, reopening the twin-gmail/twin-slack hole this guard exists to close).

All four in (3)–(5) fail in the **false-red** direction, which is the worse one: the diagnosis tells you to add a `pome:unwired-ok` marker, so following it records a false statement in the file and permanently exempts something that runs.

Two residuals I chose **not** to fix, now stated in the AGENTS.md row: importers outside `cli/scripts/**` are not derived-covered (`write-build-info.mjs`'s marker states that limit, not a real decision), and a symlinked *directory* under `cli/scripts/` is skipped with its contents.

### Battery

`check-packages-scripts-wired.mjs` 40/17 · its suite 37/37 · `lint:code-health` · `lint:dead-code` · `lint:import-meta-main` (745 files, per-root floors intact) + its suite · `actionlint` (whole dir) · `check-version-bump-required.mjs` OK · `check:manifest-schema` · `gate:recorder-overhead` PASS locally. `0.23.24` is correct against `origin/main`'s `0.23.23` (#390 took `.23`). Branch was 0 behind `origin/main`, so no rebase. Note: `cli/CHANGELOG.md`'s newest entry is `0.23.18` against a `0.23.24` package — pre-existing drift on `main`, not introduced here, and the version gate does not require an entry.

### Job-log proof the converted invocations ran

From the `overhead gate (p99 < 5ms) + PR/FAQ` job — all three, under the new shape, with env vars intact after `working-directory: cli` was dropped:

```
##[group]Run npm run gate:agent-trace-overhead -w @pome-sh/cli
[overhead-gate] N=1000 budget=5ms … pair 1 p99(with − without) = 0.902ms (budget 5ms)
[overhead-gate] PASS (first pair within strict budget)
##[group]Run npm run gate:recorder-overhead -w @pome-sh/cli
[recorder-overhead-gate] N=200 warmup=20 budget=1ms
[recorder-overhead-gate] durable p99=0.021ms (budget 1ms)
[recorder-overhead-gate] PASS
##[group]Run npm run gate:cas-adapter-acceptance -w @pome-sh/cli
> @pome-sh/cli@0.23.24 gate:cas-adapter-acceptance
[cas-acceptance] PASS
```

### On the third pass

`codex review` unavailable (usage limit to Sep 11) — substituted a hand adversarial pass: reconstructing F-1472's three marker defects as live probes, planting a real `no-eval` violation in `cli/scripts/` to prove path coverage from both cwds, planting a direct-`tsx` workflow line to characterise the convention, and reading the root lockfile for cli's 13 prod deps. Separately: the `code-review` skill, even pinned to this worktree, returned findings about `scripts/check-example-pins-published.mjs` — **not in this diff at all** (it reviewed the concurrent F-1483 branch). Its findings are discarded here; that pathology is worth knowing about.

### Mergeable: yes

Not merging, and not marking F-1476 done — that's the author's call.

PR #393, opened 2026-08-12
Realpath both sides of every CI gate's entry guard (F-1488)

GaganSD:
## Summary

Ten CI gates guarded their entry by comparing `resolve(process.argv[1])` against `fileURLToPath(import.meta.url)` — or a `pathToFileURL(...).href` / bare `fileURLToPath(...)` compare — with no `realpath` on either side. Two more compared a `basename()` instead. Node resolves symlinks before deriving `import.meta.url`, so any of these disagree through a symlinked invocation (a `git worktree`, or macOS's `/tmp` → `/private/tmp`), the guard falls false, and the process exits 0 having checked nothing.

### The twelve instances (ten named in the ticket, plus two more the extended gate's mechanical scan found beyond that list)

| File | Shape before | Verdict |
| --- | --- | --- |
| `scripts/no-eval-in-oss.mjs` (the "capture is open, evaluation is the product" gate) | `resolve(argv1) === fileURLToPath(url)` | no-realpath |
| `scripts/no-catch-and-continue.mjs` | same | no-realpath |
| `scripts/no-native-modules.mjs` | same | no-realpath |
| `scripts/capture-mcp-tools-list.mjs` | `pathToFileURL(resolve(argv1)).href === url` | no-realpath |
| `scripts/check-workspace-pins-match-workspace.mjs` | same | no-realpath |
| `scripts/check-plugin-manifest-lists-every-skill.mjs` | `resolve(argv1) === resolve(fileURLToPath(url))` | no-realpath |
| `scripts/ci/release-alarm.mjs` | `url === pathToFileURL(argv1 ?? "").href` | no-realpath |
| `packages/wire/scripts/emit-trace-contract.mjs` | same | no-realpath |
| `scripts/lint-task-class.mjs` | `url.endsWith(basename(argv1))` | **basename comparison** |
| `scripts/lint-task-format-doc.mjs` | same | **basename comparison** |
| `examples/minimal-viktor/scripts/run-trials.ts` | `fileURLToPath(url) === argv1` (no `resolve()` at all) | no-realpath |
| `examples/minimal-viktor-langgraph/scripts/run-trials.ts` | same | no-realpath |

All twelve now realpath both sides (`realpathSync(resolve(process.argv[1]))` vs. `realpathSync(fileURLToPath(import.meta.url))`) and throw if invoked as the file itself but the guard doesn't fire, matching the pattern already used by `contract/run.mjs` and `scripts/smoke-examples.mjs`.

### Basename verdict

Both basename comparisons get their own category (`basename-comparison`), separate from the realpath shapes — a basename match is satisfied by any file of that name anywhere on disk, so it was never even about symlinks. Fixed to the same realpath-both-sides form as everything else.

### The extended gate

`scripts/lint-no-bare-import-meta-main.mjs` (from #390/F-1481) already walks the repo's source tree with a TypeScript AST and covers `scripts`, `contract`, `cli/src`, `cli/scripts`, `packages`, `examples`. Extended it with `findEntryGuardRealpathGaps` rather than building a second scanner:

- Finds every `process.argv[1]` element-access node and every `import.meta.url` property-access node via AST predicates (mirroring the existing `import.meta.main` detector's style).
- Finds every entry-guard-shaped relation between them: an equality/inequality `BinaryExpression`, or the `X.endsWith(Y)` form the two basename instances used.
- For each relation, walks each operand's own parent chain (available because `createSourceFile` is called with `setParentNodes: true`) looking for a wrapping `realpathSync(...)` or `basename(...)` call.
- Classifies: `basename-comparison` (either side wrapped in `basename`), `one-sided-realpath` (exactly one side wrapped in `realpathSync`), `no-realpath` (neither side), or not a finding at all (both sides `realpathSync`'d, no `basename`).

No allow-list — the file set is the same discovered walk `scanRepo` already uses, and every relation shape found is classified, not filtered.

### Symlink evidence (`no-eval-in-oss.mjs`)

Before (checked out at `HEAD~` before this PR, invoked through a symlinked directory):
```
$ node /tmp/link/scripts/no-eval-in-oss.mjs
$ echo $?
0
```
Nothing printed — the guard fell false, the body never ran.

After (this PR, same symlinked invocation):
```
$ node /tmp/link/scripts/no-eval-in-oss.mjs
no-eval-in-oss gate passed — pome-twins is capture-only.
$ echo $?
0
```
The gate actually runs and reports its verdict.

The same before/after holds for all twelve — verified individually through a symlinked invocation; each either printed its real verdict or threw/failed on real input (unknown CLI arg, missing env var) rather than exiting silently. `examples/*/scripts/run-trials.ts` hit a pre-existing, unrelated `ERR_MODULE_NOT_FOUND` when run directly with plain `node` (a relative `.js` import that needs a loader) — reproduced identically on `main` without the symlink, so it predates this change and is out of scope.

### Break-on-purpose matrix (all in `scripts/lint-no-bare-import-meta-main.test.mjs`)

| Case | Result |
| --- | --- |
| One-sided-realpath guard added to a scratch script | reds, naming the file, `kind: "one-sided-realpath"` |
| No-realpath comparison added to a scratch script | reds, naming the file, `kind: "no-realpath"` |
| Basename comparison added to a scratch script | reds, naming the file, `kind: "basename-comparison"` |
| Correct both-sides-realpath'd guard (direct, and via the `SELF`/`ENTRY` intermediate consts every fixed instance uses) | green, no finding |
| Zero files scanned (root renamed/missing) | throws loudly, naming the root (existing per-root floor, inherited) |

Also independently re-derived: I built the ten-plus-two broken files from `origin/main` in a scratch tree and ran the extended gate against them — it flagged exactly those twelve, with the right shape tag on each, and zero false positives on unrelated dummy files placed in the other scan roots.

### CI wiring

Unchanged — same `npm run lint:import-meta-main` step in `.github/workflows/ci.yml`'s heavy block, same scope-regex coverage, same `set -euo pipefail`. The existing wiring assertions in the test file already cover this; I re-ran them and they pass with the extended detector active.

### Version bump

`packages/wire/scripts/emit-trace-contract.mjs` is publish-relevant for `@pome-sh/cli`, `@pome-sh/adapter-claude-sdk`, and `@pome-sh/wire` itself (all three list `packages/wire/` as a publish-relevant path). Hand-bumped all three with CHANGELOG entries — no behavior/schema change, dev-tooling-only. `origin/main`'s CLI version was `0.23.23` when I branched; bumped to `0.23.25` to avoid colliding with `#392`, which is taking `0.23.24`.

### What a reviewer should challenge

- `findEntryGuardRealpathGaps` only sees a comparison that references `process.argv[1]`/`import.meta.url` (or a `basename()`/wrapper call of them) *directly* inside the relation. A guard that aliased either side through an intermediate variable *before* the comparison (not the `SELF`/`ENTRY` pattern, which computes the realpath'd value into the variable — that's fine — but something like `const argv1 = process.argv[1]; const url = import.meta.url; if (argv1 === url)`) would not be caught. No live instance in this repo does this, and I didn't add a dataflow pass for it — flagging it here rather than silently leaving the gap.
- I found two more broken instances than the ten the ticket named (`examples/minimal-viktor{,-langgraph}/scripts/run-trials.ts`) by running the generalized detector rather than only fixing the named list. Worth double-checking I haven't mis-scoped these as in-scope — they're dev-only Slack utility scripts, not CI gates, but they're under `examples/`, which the existing gate already scans, and they had the exact same silent-skip shape.


GaganSD:
## Review — F-1488 (reviewer of record)

Three passes: the `review` skill, the `code-review` skill pinned to this worktree, and — because `codex review` is unavailable (usage limit to Sep 11) — a substitute adversarial pass of my own, described at the bottom. I fixed what I found; the fixes are the last five commits on this branch.

The twelve guard rewrites are correct and I verified five of them myself, end to end. The problem was the other half of the ticket: **the gate meant to assert the property could not see a single entry guard in this repository.**

---

### 1. BLOCKER (fixed) — the new gate half was vacuous over 100% of the repo's guards

`findEntryGuardRealpathGaps` searched only a comparison's own operands. But the sanctioned form this PR standardizes on routes both sides through intermediate consts:

```js
const SELF = realpathSync(fileURLToPath(import.meta.url));
const ENTRY = process.argv[1] ? realpathSync(resolve(process.argv[1])) : "";
const invokedDirectly = ENTRY === SELF;
```

The comparison is `ENTRY === SELF`. Neither operand contains `process.argv[1]` or `import.meta.url`, so **no relation was classified at all** — not in any of the twelve files this PR just fixed, not in `contract/run.mjs`, not in `probe-twin-endpoints.mjs`, and not in the gate's own guard. Measured before the fix:

```
scripts/no-eval-in-oss.mjs                    gaps=0   gaps-with-realpath-neutered=0   <-- INVISIBLE
scripts/ci/release-alarm.mjs                  gaps=0   gaps-with-realpath-neutered=0   <-- INVISIBLE
packages/wire/scripts/emit-trace-contract.mjs gaps=0   gaps-with-realpath-neutered=0   <-- INVISIBLE
scripts/lint-task-class.mjs                   gaps=0   gaps-with-realpath-neutered=0   <-- INVISIBLE
examples/minimal-viktor/scripts/run-trials.ts gaps=0   gaps-with-realpath-neutered=0   <-- INVISIBLE
scripts/lint-no-bare-import-meta-main.mjs     gaps=0   gaps-with-realpath-neutered=0   <-- INVISIBLE
contract/run.mjs                              gaps=0   gaps-with-realpath-neutered=0   <-- INVISIBLE
... 15/15 files, all invisible
```

Second column is the decisive one: strip `realpathSync` out of the file entirely and the gate *still* reports zero gaps. So the ticket's own acceptance criterion — *"Do: add a one-sided-realpath guard to a new script. Expect: CI reds naming it"* — held **only** for the inline spelling, which no file in the repo uses any more. Copy a fixed sibling, drop a `realpathSync`, and CI stayed green.

The test enshrined this as intended behaviour ("`via SELF/ENTRY intermediate consts` … is not a finding"). It was green because the shape was invisible, not because it was recognized as correct.

**Fixed** by following single-assignment aliases one hop (a name declared twice is dropped rather than guessed at). Relations classified across the repo: **0 → 27**.

Also fixed in the same pass: `findLeaves` returned its accumulator from a `ts.forEachChild` callback, and `forEachChild` treats a truthy return as "found it" and stops iterating siblings — so it only ever visited the first child of each node.

### 2. BLOCKER (fixed) — no floor, then a floor one step from vacuous

`guardGaps.length === 0` is satisfied equally by "every guard realpaths both sides" and "the walk recognized no guard at all", and those read identically in a CI log. Two floors in this milestone shipped dead for exactly this reason.

`scanRepo` now returns `guardRelations` and the runner fails below a floor. My first cut used `=== 0`; **@code-review correctly pushed back** that this is still one regression from vacuous — only *one* guard in the repo (`check-packages-scripts-wired.mjs`) compares the two inline, so 26 of 27 relations come through the new alias deref. Stubbing the deref to identity dropped the count to 1 and the gate printed `all 1 entry-guard comparison(s) realpath both sides` and exited 0. Now floored at `MIN_ENTRY_GUARD_RELATIONS = 12` (F-1488's instance count) against 27 actual, so a classifier regression reds while legitimately deleting a script does not. Verified: stubbed deref → exit 1 naming the count.

### 3. BLOCKER (fixed) — `check:trace-contract` was red; the wire bump was incomplete

`packages/wire/trace-contract.json` embeds the package version. Bumping wire to 0.2.2 without re-emitting left it declaring `0.2.1`, and `check:trace-contract` is wired in **both** `ci.yml:321` and `release.yml:327`. The branch was failing CI on the version bump it needed for its own fix. Regenerated.

This also answers the question the PR body raised: `trace-contract.json` **is** in wire's `files` array, so the bump changes a real byte of the published tarball. The wire bump is self-justifying, not packaging-only — the CHANGELOG said the opposite and is corrected.

### 4. BLOCKER (fixed) — a sibling test asserted source ORDER, not the guard

`capture-mcp-tools-list.test.mjs` asserted `/process\.argv\[1\][\s\S]{0,120}import\.meta\.url/`. The sanctioned form declares `SELF` (i.e. `import.meta.url`) first, so the regex **failed on a guard that had just been made strictly stronger**. `typecheck-test` was red on it (`ci.yml:285`). Caught by @code-review, not by me.

Fixed with two order-independent regexes, one per side, asserting `realpathSync` around each — positively stronger than what they replace. Note for future readers: my first attempt reused `findEntryGuardRealpathGaps` for this, which is the right instinct in the wrong block — that module imports `typescript` and this test runs in the **cheap** CI block with no `npm ci`, so it died with `ERR_MODULE_NOT_FOUND` before asserting anything. Re-verified with `node_modules/typescript` removed.

### 5. LOW (reported, not fixed) — `realpathSync` at module scope can crash on import

All twelve use `process.argv[1] ? realpathSync(resolve(process.argv[1])) : ""`. That guards a *missing* argv0, not an argv0 naming a path that isn't on disk — `realpathSync` throws `ENOENT`. Reproducible via `node -e 'import("./scripts/no-eval-in-oss.mjs")' notafile`. No live caller reaches it, and the failure is loud rather than silent, which is the opposite of this ticket's bug class — so I left it. If you want it closed, `existsSync` or a `try/catch` per file is the fix, or a shared helper if you'd rather not repeat it thirteen times.

### 6. NOTE (reported, not fixed) — `check-example-pins-published`'s E404 sniff is log-level coupled

`defaultNpmView` classifies "not published yet" by regexing `E404` out of npm's **stderr text**. Under `npm run --silent`, the inner `npm view` inherits `npm_config_loglevel` and prints nothing, so the E404 is invisible and a legitimately-unpublished version lands in `error` instead of `skip`:

```
npm run --silent typecheck:examples  ->  1 errored   (registry lookup FAILED)
npm run          typecheck:examples  ->  1 skipped   (correct)
```

CI invokes it without `--silent` (`ci.yml:398`) so **this PR is unaffected**, and I did not touch another gate's semantics inside this one. But this branch's adapter bump to 0.3.4 is the first change ever to exercise that unpublished branch, and it only works by accident of invocation flags. Worth a follow-up on F-1483 — `--loglevel=error` in the argv, or `--json` plus an exit-code check, removes the coupling.

### 7. NOTE — three byte-identical republishes, by gate design

`check-version-bump-required.mjs` treats any `packages/wire/` change as publish-relevant for wire, cli **and** adapter-claude-sdk. The only wire file this PR changes is `scripts/emit-trace-contract.mjs`, which is dev tooling excluded from wire's `files` array and never seen by tsup — so `@pome-sh/cli` 0.23.26 and `@pome-sh/adapter-claude-sdk` 0.3.4 are byte-identical republishes. Both CHANGELOGs say so honestly, which is the right call, and the gate demands them. This is the same F-1455 over-match the gate already carries three documented carve-outs for (twin `examples/`, twin top-level `.md`, twin `scripts/`) and is missing a fourth for `packages/*/scripts/`. Out of scope here — changing it would be a policy edit smuggled into a guard fix — but it is a real source of bump noise.

---

## Independent symlink verification — 5 of 12

Method (no `git stash`): `git show origin/main:<path>` into sibling scratch files, run through `/tmp/f1488x/link -> <worktree>`, compare against the fixed version, then run each gate **normally** to confirm it still does its real job.

| gate | `origin/main` via symlink | this PR via symlink | this PR invoked normally |
|---|---|---|---|
| `no-eval-in-oss.mjs` (capture-is-open) | **rc=0, 0 bytes — checked nothing** | `gate passed — pome-twins is capture-only.` | same, real scan |
| `ci/release-alarm.mjs` | **rc=0, 0 bytes — checked nothing** | reads `main HEAD 0c0c2ffc`, hits the registry | same; and without `GITHUB_REPOSITORY` it throws its own validation, so the body ran |
| `lint-task-class.mjs` (basename case) | **rc=0, 0 bytes — checked nothing** | `44 task(s): 20 conformance, 8 restraint, 16 adversarial` | same |
| `capture-mcp-tools-list.mjs` | **rc=0, 0 bytes — checked nothing** | `gmail: live-wire-unauth — golden matches (13 tools)` | same |
| `no-native-modules.mjs` | **rc=0, 0 bytes — checked nothing** | `49 production packages clean` | same |

The basename cases need a *renamed* symlink, not just any symlink — `basename(argv[1])` survives a same-name link. `/tmp/f1488x/aka-main.mjs -> scripts/lint-task-class.mjs` is silent on `main` and checks on this branch. I also confirmed the inverse weakness: a decoy file merely *named* `lint-task-format-doc.mjs` elsewhere on disk made `main`'s guard run `main()` on import; on this branch it throws instead.

The throw path fires as designed, and does not misfire: `examples/minimal-viktor` (8 tests) and `examples/minimal-viktor-langgraph` (26 tests) both pass, despite the two `run-trials.ts` sharing a basename, because vitest's argv0 is its own binary.

## Detector boundaries, both directions

Measured, before → after my fix:

| pattern | before | after | verdict |
|---|---|---|---|
| `resolve()`-only compare | RED | RED | correct |
| `pathToFileURL(...).href` compare | RED | RED | correct |
| one-sided realpath, either side | RED | RED | correct |
| `endsWith(basename(...))` | RED | RED | correct |
| **alias-routed, no realpath (the shipped shape, reverted)** | **green** | **RED** | **was the blocker** |
| **alias-routed, one-sided realpath** | **green** | **RED** | **was the blocker** |
| `process.argv.at(1)` | green | RED | closed |
| `import.meta.filename` instead of `.url` | green | RED | closed |
| destructured `const [, entry] = process.argv` | green | green | **still open — false green** |
| inline sanctioned form | green | green | correct |
| **`fs.realpathSync` both sides** | **RED** | green | **was a false red** |
| **`realpathSync.native` both sides** | **RED** | green | **was a false red** |
| **`await realpath()` from `fs/promises`** | **RED** | green | **was a false red** |
| realpath inside a helper (`const real = p => realpathSync(p)`) | RED | RED | **still a false red — documented ceiling** |
| `resolve()` wrapped outside `realpathSync` | green | green | correct |

Two ceilings remain, both documented in the source and in AGENTS.md. The helper case errs toward a false RED on correct work, which is the safe direction for a gate. The destructured case is a false GREEN, which is not — the relation floor is what stands behind it. I swept the repo for both plus `require.main`, `process.argv.at(`, `import.meta.filename/dirname` compares and live `endsWith(basename(` guards: every remaining hit is a comment or a test fixture. No live instance of any unseen shape.

## Substitute third pass (in place of `codex review`)

- **Semantic diff of all twelve files against `origin/main`**, filtering out comments, imports and the guard block. What remains is *only* the old guard line being deleted (plus two multi-line `throw new Error(` wrappers). No drive-by policy edit in `no-eval-in-oss.mjs` or `capture-mcp-tools-list.mjs`.
- **Mechanical audit of every `ENTRY.endsWith("…")` throw** against its own file's basename — a typo there makes the throw silently never fire. All 15 in the repo match.
- **Blindness injection**: neutered `realpathSync` per file to prove the gate could not see any guard; stubbed `deref` to identity to prove the `=== 0` floor was insufficient.
- **Cheap/heavy CI block dependency check** by running the suite with `node_modules/typescript` removed.
- Four now-unused `pathToFileURL` imports removed (knip does not flag unused imports, so these would have sat there).

## Version-bump state against current `origin/main`

The branch was **2 commits BEHIND** on review (#392 and #388 both landed today). Rebased with `--force-with-lease`; `AGENTS.md` and `cli/CHANGELOG.md` both conflicted and neither resolution was mechanical:

- `AGENTS.md` — #392 **deleted** `cli/scripts/make-unwired-fixture.mjs` and genericized the sentence naming it. This branch's text still named the deleted file; I took #392's wording and re-applied the F-1488 additions on top.
- `cli/CHANGELOG.md` — both sides added a `## 0.23.25` heading. Dropped this branch's and folded its content into the `## 0.23.26` entry, which otherwise documented only the renumbering and none of the actual change.

After the rebase: `cli` 0.23.26 (main 0.23.25), `adapter-claude-sdk` 0.3.4 (main 0.3.3), `wire` 0.2.2 (main 0.2.1). `check-version-bump-required.mjs` passes against current main. `cli/package.json`'s `description` bytes are **byte-identical** to `origin/main` — verified with `xxd`, the `—` escape intact, no `JSON.stringify` round-trip.

Re-ran the whole battery post-rebase, not just the parts that touched my changes: `lint:import-meta-main` + suite, `capture-mcp-tools-list.test.mjs`, `actionlint`, `lint:code-health`, `lint:dead-code`, `check-packages-scripts-wired` + suite, `test:contract` (115 tests), `typecheck`, `typecheck:examples`, `check:trace-contract`, `check-version-bump-required`, and every gate I touched invoked normally. All green.

---

## Verdict

**Not mergeable as it arrived** — it was CI-red on two counts and its central assertion was vacuous. **Mergeable as it now stands**, once CI confirms.

The guard rewrites were right the first time. What needed the work was the thing the ticket said mattered most: *"Assert the property, do not fix ten files and stop."* As submitted, the property was asserted over a shape none of the twelve files used.

Not merging, and not marking F-1488 done — the author should confirm the alias-deref approach and decide on findings 5, 6 and 7.

---

**CI on `d6a495b`:** all required checks pass. The heavy-block job log confirms the extended gate actually executed on the runner, with a plausible count rather than a vacuous zero:

```
> npm run lint:import-meta-main
No bare `import.meta.main` in 749 file(s) under scripts, contract, cli/src,
cli/scripts, packages, examples, and all 27 entry-guard comparison(s) realpath both sides.
```

(The five per-twin matrix jobs were still running at the time of writing.)

PR #394, opened 2026-08-12
Give the smoke gate's strongest verdict a credentialed leg where it is achievable (F-1486)

GaganSD:
## What is now true

`smoke:examples` has two verdicts, OK (alive at the settle) and REACHED-OUTBOUND
(launched, wired its twin, failed at an outbound call). Measured across #386,
OK was 0/8 in CI and 4/8 locally — the strongest verdict was verified nowhere
that gates anything.

A new `smoke-examples-live.yml` workflow (`schedule:` nightly + `workflow_dispatch`)
runs the *same* gate (`scripts/smoke-examples.mjs`, unchanged classifier) with:

- real local twins, booted for real via `pome twin start` (same in-process boot
  path `quickstart-smoke.yml` already proves for github/gmail) — github (3333),
  slack (3334), gmail (3336), all sharing one generated `TWIN_AUTH_SECRET` so a
  single minted token verifies against all three;
- a real `ANTHROPIC_API_KEY` repo secret.

`SMOKE_EXAMPLES_LIVE=1` tells `smoke-examples.mjs` to stop overlaying `SMOKE_ENV`'s
dead loopback ports/invalid keys (so the real wiring flows through unmodified) and
to assert a floor after the run.

The PR leg (`ci.yml`) is untouched: `SMOKE_EXAMPLES_LIVE` is never set there, so
`REACHED-OUTBOUND` × up to 8, uncredentialed, stays a pass — verified locally
(8/8 reached real work, exit 0, unchanged messaging).

## Secrets a human must provision

**Exactly one, and it must be non-blank:**

- **`ANTHROPIC_API_KEY`** — a real Anthropic API key with quota, as a repo
  Actions secret. Nothing else needs provisioning.

Why the list is one name while the absent-credential demo prints two
(`ANTHROPIC_API_KEY, POME_AUTH_TOKEN`): **`POME_AUTH_TOKEN` is not a secret
anyone provisions.** It is minted per run by `pome twin start` from a
`TWIN_AUTH_SECRET` the workflow generates with `crypto.randomBytes(32)`, read
off the twin's own printed boot line, and written to `$GITHUB_ENV`. It appears
in `LIVE_REQUIRED_ENV` because it is the one twin-side signal that proves real
twins were booted for this run — checking it is how the leg refuses to run
against nothing. In a bare shell with no twins it is legitimately absent, which
is what the demo shows. Both statements were true; the PR body did not say so.

**`AI_GATEWAY_API_KEY` is not required — but only because review changed
something to make that true.** As originally written it *was* required, and the
leg could never have gone green without it: `minimal-viktor` defaults to
`VIKTOR_MODEL=alibaba/qwen-3-32b`, which its `resolveModel()` can reach ONLY
through the AI Gateway, and which throws `needs AI_GATEWAY_API_KEY` before any
outbound call. The PR leg hides this behind a *fake* `AI_GATEWAY_API_KEY` in the
dead-wiring overlay; the credentialed leg cannot use that trick, because a fake
gateway key routes all eight examples into a gateway that rejects them and the
real `ANTHROPIC_API_KEY` is then never exercised at all. The leg now pins
`VIKTOR_MODEL` to the `anthropic/*` slug two sibling examples already default
to, so one secret genuinely covers all eight. Provisioning a gateway key is
still supported (a caller-supplied `VIKTOR_MODEL`/`AI_GATEWAY_API_KEY` wins) but
is not needed.

No GitHub/Slack/Gmail vendor credentials are needed — the twins are self-hosted
mocks booted fresh each run against their own default seeds, not real vendor
accounts.

## What the leg asserts, and the floor's derivation

After running all discovered examples, `assertAliveFloor({ live, okCount, total })`
requires `okCount >= 1`. The floor is a literal `>= 1`, not a fraction or count of
`total`: comparing the floor to the same denominator it's counted against is the
exact "two quantities that move together" shape this milestone has found dead on
arrival twice already — it would be satisfied forever by construction as examples
are added or removed. `>= 1` is independent of `total` and is the precise property
F-1486 names: zero-alive is the fact no other environment can ever surface.

Absent-credential behavior is a *hard, named* failure, not a silent downgrade:

- `missingLiveEnv()` in the script checks `ANTHROPIC_API_KEY` and `POME_AUTH_TOKEN`
  are both present and exits 1 before any example launches if either is missing,
  naming which.
- The workflow itself also fails fast on an absent `ANTHROPIC_API_KEY` before
  paying for `npm ci` or booting any twin (same shape as `repo-policy.yml`'s
  `REPO_POLICY_TOKEN` check).

Demonstrated: `SMOKE_EXAMPLES_LIVE=1 node scripts/smoke-examples.mjs` with no
credentials set exits 1 naming `ANTHROPIC_API_KEY, POME_AUTH_TOKEN` before
launching anything.

## Alarm wiring (F-1230/F-1471)

`smoke-examples-live.yml` is `schedule:`-triggered, so it carries
`schedule-alarm`/`schedule-alarm-recovery` jobs reaching the reusable
`schedule-alarm.yml`, gated on `schedule`/`workflow_dispatch`, with
`title: "the smoke-examples-live nightly is failing"` and
`label: "schedule-alarm:smoke-examples-live"` — checked unique against every
other alarm in the tree by F-1471's own bijection guard.

## Verified

- `actionlint .github/workflows/*.yml` — clean.
- `node scripts/ci/assert-schedule-alarm-coverage.mjs` — enumerates 4 scheduled
  workflows including `smoke-examples-live.yml`, all reaching the alarm, no
  title/label collision.
- `node scripts/ci/list-scheduled-workflows.mjs` / its test suite / the
  coverage guard's test suite — all pass.
- `node scripts/smoke-examples.test.mjs` — new cases cover the floor (0 alive
  fails naming the expectation, >= 1 passes, the floor is a no-op off the live
  leg) and the missing-credential check (empty env names both required vars,
  a partial env names only what's absent).
- `npm run smoke:examples` locally with no `SMOKE_EXAMPLES_LIVE` set — 8/8
  reached real work (4 OK, 4 REACHED-OUTBOUND here), exit 0: the PR leg's
  behavior is unchanged.
- `npm run lint:code-health`, `lint:dead-code`, `lint:import-meta-main`,
  `node scripts/check-packages-scripts-wired.mjs` — all green.
- `node scripts/ci/check-version-bump-required.mjs origin/main` — no bump
  required (no publish-relevant path touched).
- AGENTS.md's P8 table gained a row for the new leg.

## What could not be verified

No real credentialed run has been observed. I do not have `ANTHROPIC_API_KEY`
and cannot provision repo secrets or trigger a real nightly from here. Once
the secret is provisioned, what remains to confirm:

- That `workflow_dispatch` produces `okCount >= 1` for real — plausible (a
  real multi-turn agentic call against a live twin should easily still be
  "in flight" at the 5s settle) but unobserved.
- Whether the generic `SMOKE_ENV` task text ("triage/summarize the open items
  in acme/api") matches data any twin's *default* seed actually contains. If
  it doesn't, an example could get a fast, correct "not found" answer from a
  real model and finish *before* the 5s settle — which the current classifier
  would read as a FAIL ("returned with no evidence of real work"), not because
  the run was unhealthy but because it was fast and correct. This is a real
  edge in the classifier's OK criterion that only surfaces with actual
  credentials; if the first real run shows it, the fix is likely either a
  seed that matches the task or a different signal than "still running," not
  a change to this leg's wiring.
- Whether booting three twins plus a full workspace build fits comfortably
  inside the 20-minute job timeout on a cold cache.


GaganSD:
## Review — F-1486 (reviewer of record)

Three passes: the `review` skill, the `code-review` skill pinned to this worktree, and a substitute adversarial pass (`codex review` is unavailable to Sep 11 — not faked; what I did instead is listed at the bottom). Findings fixed on the branch in 3 commits; CI green.

**Verdict: mergeable.** The design is right — the floor's reasoning holds, the alarm wiring is real, and the PR leg is untouched. But the leg as submitted **would have redded on its first nightly, on all eight examples, without ever testing the property it exists to prove.** Two independent missing-env defects, both invisible to the submitted tests.

### 1. HIGH — the LIVE leg never supplied `POME_TASK`. 4 of 8 red every night.

`SMOKE_EXAMPLES_LIVE=1` dropped the *entire* `SMOKE_ENV` overlay. But `POME_TASK` lived in that overlay and is not a fake credential — it is the only thing that gives a launched example anything to do, and `gmail-retry-notify`, `merge-agent`, `minimal-viktor` and `minimal-viktor-langgraph` all call `requiredEnv("POME_TASK")`. No workflow step wrote it.

Reproduced by running the leg's exact env shape:

```
=== examples/merge-agent === FAILED (... returned without evidence it did any real work ...)
Error: POME_TASK is required
```

× 4. Non-benign exit 1 → `fail` → `process.exit(1)` → alarm issue filed, every night, for an unset env var with nothing to do with aliveness at the settle. This is the "latent false red that trains someone to ignore the alarm" case, arriving on run #1.

### 2. HIGH — `minimal-viktor` could never pass, and the secrets list was wrong because of it.

`minimal-viktor` defaults to `VIKTOR_MODEL=alibaba/qwen-3-32b`. Its `resolveModel()` short-circuits to the slug only when `AI_GATEWAY_API_KEY` is set; `alibaba/*` matches no direct-SDK branch, so without a gateway key it throws `VIKTOR_MODEL=alibaba/qwen-3-32b needs AI_GATEWAY_API_KEY` **before any outbound call** — non-benign exit 1, FAIL, red.

So the body's "`ANTHROPIC_API_KEY` only … a single key covers all 8" was **not true as written**, and its stated reason (that the gateway key is the fallback *behind* `ANTHROPIC_API_KEY`) is backwards — the gateway key is checked *first* and is the only path to the default slug.

Fixed by pinning `VIKTOR_MODEL` to the `anthropic/*` slug `merge-agent` and `gmail-retry-notify` already default to, rather than by asking for a second secret. A *fake* gateway key (the PR leg's trick) is not an option here: it would route all eight into a gateway that rejects them and the real `ANTHROPIC_API_KEY` would never be exercised — a credentialed leg whose credential is unreachable. One secret now genuinely covers all eight. PR body corrected.

Both defects share a root cause worth naming: `LIVE` was read at module scope and closed over inside `smokeOne()`, so the env split was **structurally untestable** — which is exactly why 18 green checks could not see either. `launchEnv(baseEnv, live)` is now exported and pure, and asserts on both legs that the live leg supplies a task and an `anthropic/*` slug and leaks no `59321` loopback port, while the PR leg still overlays the dead port, the invalid key and the fake gateway key.

### 3. MEDIUM — fail-closed matrix: whitespace-only credentials sailed through, both sides

`missingLiveEnv()` used `!env[name]`, and the workflow used `[[ -z ]]`. Actions substitutes an *unset* secret as the empty string (falsy — caught either way), but a secret whose stored value is a pasted space or newline is **truthy** and passed both checks. That is the F-1187/F-1184 blank-in-Infisical shape verbatim. Untrimmed, it degraded a named setup failure into eight confusing per-example crashes. Both sides trim now.

Measured matrix (each must be a hard, *named* failure):

| case | before | after |
| -- | -- | -- |
| both absent | named, exit 1 | named, exit 1 |
| both empty string | named, exit 1 | named, exit 1 |
| **both whitespace-only** | **bypassed → 4 unnamed FAILs** | named, exit 1 |
| **one real + one whitespace-only** | **bypassed** | named, exit 1 |
| partial (one absent) | named, exit 1 | named, exit 1 |
| workflow step, whitespace secret | **passed the gate** | `::error::` naming it |

### 4. MEDIUM — a one-character typo turned the nightly into a green no-op

`export const LIVE = process.env.SMOKE_EXAMPLES_LIVE === "1"`. Anything else — `true`, `yes`, `TRUE`, `1 ` — made `LIVE` **false**, which silently means *the PR leg*: dead loopback ports overlaid, no credential check, no floor, `REACHED-OUTBOUND` × 8, **exit 0**. A green nightly proving nothing, i.e. this ticket's own subject, reachable by a stray character in the YAML. `resolveLiveFlag()` now makes an unrecognised value a named hard error; unset/`""` still mean the PR leg. Nine bogus values covered.

### 5. MEDIUM — the twin-wiring extraction wrote blanks instead of failing

`echo "VAR=$(grep -oP … twin.log)"` does **not** trip `set -e` when the grep misses: the substitution fails, `echo` exits 0, and `VAR=` reaches `$GITHUB_ENV`. Only the two names in `LIVE_REQUIRED_ENV` are re-checked downstream, so a renamed or dropped wiring line in `twinStart.ts` would surface as eight per-example crashes rather than one named setup failure. `quickstart-smoke.yml` guards this with `grep -q`; this workflow had dropped it. A `grab()` helper now reds naming the var and the log.

### 6. The floor — both verdicts you asked for

- **LIVE-only: holds.** `assertAliveFloor` returns `{ok:true,message:null}` on `!live`, asserted directly, and the PR leg's summary path is unchanged. Confirmed in real CI on this branch: `0 still running at the settle, 8 failed at an outbound call`, **exit 0**. The floor cannot redden the PR leg.
- **Can LIVE pass with the floor never evaluated: no.** Every exit before the floor is a *failure* exit (`process.exit(1)` for the bad flag, the missing credential, and zero examples discovered), so no early return reaches a green. `smokeOne()` never rejects — it resolves a verdict on `error` too — and a throw in `main()` is an unhandled rejection at top-level `await`, which is non-zero. The floor is on the single summary path and `failures.length > 0 || !floor.ok` gates it.
- `>= 1` not compared to `total` is right, for the reason stated: comparing it to the same denominator it is counted against is the move-together shape that killed two floors in this milestone.

**One honest ceiling I did not fix**, because the ticket says keep the classifier: `OK` is *"still alive at the settle"*, which is weak evidence under real credentials. In my simulated LIVE run with a deliberately **fake** key against **dead** ports, 4 of 8 were still `OK` and the floor reported *met*. A process retrying a rate-limited call or blocked on a slow host also counts. So the floor proves "at least one process breathed for 5s with real wiring present", not "did real work" — worth a follow-up, and now recorded in AGENTS.md as a known ceiling rather than left implied.

### 7. The fast-completion edge — judgement

Acceptable-for-now **as a verdict**, not acceptable as a *message*. A correct-but-fast example that exits 0 inside the settle lands on the final FAIL, whose text insists the example is broken ("the likely case: a wrong parse, an early return, or a swallowed error"). That is a latent false red whose first firing is undiagnosable.

The cheap honest fix you floated — distinguishing "exited 0 having produced work output" from "exited 0 having done nothing" — is *not* cheap here: it needs a work marker no example emits in a common form across three frameworks (Claude Agent SDK, AI SDK, LangGraph). Adding one would be F-1478's subject reopened, not a review fix. So: the FAIL message now names this specific possibility, **on the LIVE leg only** (a non-zero exit and the whole PR leg are excluded, so F-1478's message stays correct where it is correct), and tells the reader to check the tail before assuming breakage and that the fix is a matching seed or a work signal, not the example.

### 8. Alarm wiring — I broke it on purpose, four ways

Re-ran the guard: 4 scheduled workflows, all covered, no title/label disagreement. Then attacked the new caller specifically:

| break | guard |
| -- | -- |
| comment out the `schedule-alarm` job | **RED**, naming `smoke-examples-live.yml` |
| `continue-on-error: true` on it | **RED** |
| collide `title`/`label` with `repo-policy`'s | **RED**, naming both sides |
| **remove `issues: write` from the calling job** | **GREEN — not caught** |

The first three are genuinely enforced, so F-1471's green is not vacuous for this caller. The fourth is a **pre-existing gap in F-1471's guard, not a defect in this PR** — this PR does grant `issues: write` on both alarm jobs, so the alarm works. But the guard would not stop someone removing it, and a reusable-workflow call whose caller under-grants permissions fails at dispatch. Worth its own ticket against the guard.

End-to-end reachability traced: `failure()`-guarded job → `uses: ./.github/workflows/schedule-alarm.yml` (resolves, `actionlint` clean) → `issues: write` on both the caller and the reusable job → `title`/`label` pair unique across all 8 alarm calls in the tree (`the smoke-examples-live nightly is failing` / `schedule-alarm:smoke-examples-live`), so no duplicate pair can close another alarm's issue. Recovery closes on `success()`.

### 9. Versions

`check-version-bump-required.mjs` against **current** `origin/main`: no bump required — no publish-relevant path is touched (two workflow/script files, AGENTS.md, a test). `cli` stays at **0.23.25**, matching main, so #393's 0.23.26 is not contended and 0.23.27 is not needed. `cli/package.json` is byte-identical to main, `description` included. Branch is **0 commits behind** main, so no rebase.

### 10. Substitute third pass (in place of `codex review`)

`codex review` is unavailable (usage limit to Sep 11); I did not fake it. Instead: ran the script under the workflow's **exact** env shape via `env -i` (which is what surfaced #1 and #2 — neither was visible from reading the diff or from the test suite); enumerated every `process.env` read across all 8 examples and diffed that set against what the workflow provides; traced each of the 4 model-resolution chains to the key it actually requires; drove the fail-closed matrix empirically including empty-string, whitespace-only and partial; probed the alarm guard with 4 deliberate breaks; and confirmed the PR leg unchanged both locally (exit 0, 4 OK + 4 REACHED-OUTBOUND, the split AGENTS.md documents) and in this branch's real CI log (exit 0, 0 OK + 8 REACHED-OUTBOUND — F-1486's own measured figures, unmoved).

### Still unverifiable, and honestly so

No real credentialed run has been observed — I cannot provision the secret. What the first `workflow_dispatch` must confirm: that `okCount >= 1` holds for real; whether the generic task text matches any twin's default seed (finding 7's trigger); and whether three twin boots plus a full workspace build fit the 20-minute timeout on a cold cache. The author flagged all three; they remain open, and the leg now fails *named* rather than mysteriously if the first two go wrong.

Not merged, Linear not marked done.

GaganSD:
## Post-rebase: this branch is red on a breakage inherited from `main`, not from this PR

Rebased onto current `origin/main` (which had taken #393 / F-1488, and with it `cli` 0.23.26 — this PR needs no bump, so there is nothing contended). The rebase was clean **and it does not work**, which is the distinction worth recording:

```
❌ published pin DRIFT in 1 example(s):
  examples/support-triage (dependencies.@pome-sh/adapter-claude-sdk):
    pins 0.3.3, but the workspace sibling is 0.3.4 and 0.3.4 is published. Re-pin to 0.3.4.
```

**`origin/main` is red on this itself.** The `ci` run for `0e181da` (#393's merge commit) fails at the same step, with the same message. #393 bumped `packages/adapter-claude-sdk` to 0.3.4 and the release published it, but `examples/support-triage` still pins `0.3.3` — F-1483's registry watch is doing exactly its job, and it flipped red the moment 0.3.4 hit the registry. Neither file is touched by this PR (`git diff --name-only origin/main...HEAD` is the 4 files above and nothing else), and this branch was green on the pre-#393 base.

Deliberately **not** fixing it here. It is a one-line re-pin plus a lockfile update, it belongs to #393's fallout rather than to F-1486, and it blocks every branch cut from current `main` — so it wants its own PR, landed on `main`, not a published-dependency change buried inside a CI-workflow PR. Flagging rather than absorbing.

Consequence for this PR's evidence: `typecheck-test` now dies at pin-parity, which runs **before** `smoke:examples` in the heavy block, so that gate did not execute in the post-rebase run. The pre-rebase run at `6ecb3cc` did execute it, on byte-identical files (`.github/workflows/smoke-examples-live.yml`, `scripts/smoke-examples.mjs`, `scripts/smoke-examples.test.mjs` all hash-identical pre/post rebase; only `AGENTS.md` differs, and only by #393's own added rows). That run reported:

```
All 8 examples reached real work: 0 still running at the settle, 8 failed at an outbound call.
```

exit 0 — `REACHED-OUTBOUND` × 8, `OK` × 0: F-1486's own measured CI figures, unmoved by this PR. So "the PR leg is unchanged" remains CI-verified for the code as it stands; it is simply not re-demonstrable until `main` is green.

Everything else post-rebase, re-run locally on the rebased tree: `actionlint`, `assert-schedule-alarm-coverage.mjs` (4 scheduled workflows, all covered), `lint:code-health`, `lint:dead-code`, `lint:import-meta-main`, `check-packages-scripts-wired.mjs`, the unit suite, and `check-version-bump-required.mjs origin/main` — all green. `smoke:examples` locally: exit 0, 4 OK + 4 REACHED-OUTBOUND. `CodeQL`, `Analyze (actions)`, `Analyze (javascript-typescript)`, `dependency review`, `gitleaks + trufflehog` all pass on the rebased head.

**Mergeable on its own merits: yes. Blocked on `main` being repaired first.** Not merged; Linear not moved.

PR #434, opened 2026-08-18
Retry the twin image push on a degraded registry, and leave no half-published tag

AFFFPupu:
Executive summary: `twin-image.yml`'s `Push scanned image` was the last unretried network call on the publish path — one bare `docker push` per tag, standing between two third-party fetches that had both been given three-attempt ladders. A single `unknown blob` from GHCR on 2026-08-14 meant `twins:stripe-bbf27bf` was never published, and because nothing re-ran that step, one transient registry error stayed load-bearing for four days. This adds the retry, plus a per-tag manifest read-back so "the push exited 0" is no longer mistaken for "this tag resolves", and extends the workflow-tree gate so a future unretried registry write reds CI instead of waiting to be noticed.

## What

`scripts/ci/push-scanned-image.sh` (new) is now the one hardened registry-write path: five attempts per tag with 5/10/15/20s backoff, and a tag counts as published only once `docker buildx imagetools inspect` reads its manifest back. `twin-image.yml`'s push step calls it, passing tags through `env:` rather than interpolating `${{ }}` into a shell body (matching the sign step below it). `scripts/ci/assert-hardened-cdn-fetches.mjs` grows a third shape: a `run:` block that writes to a registry (`docker push`, `docker manifest push`, `docker buildx imagetools create`) must go through that helper, with a dead guard so the new half cannot report a vacuous green.

## Why

The push had no retry, and a `docker push` that reports every layer `Pushed` and then fails the manifest PUT is a known GHCR consistency fault that a second attempt clears. The four downstream alarms all traced to that one flake: four consecutive red `twin-snapshot-verify` runs, `pome-cloud#752` reporting `stripe — no-image`, and production still booting snapshots from a commit 43 behind. Hardening only the step that broke would be the same shape as the bug — a list of the calls someone remembered — so the gate is a property over the workflow tree, which is how the unretried write between two hardened fetches was found in the first place.

## Notes for reviewer

**The read-back is the load-bearing half, not the retry.** `docker push` exiting 0 is not the same fact as "this tag resolves"; the fault class here is a registry that accepts blobs and commits nothing that points at them. A tag that pushes clean and resolves to nothing is retried here rather than failing one step later as a confusing cosign error. It is the same call `resolve-image-digest.ts` makes.

**Why exhaustion exits instead of continuing to the next tag.** `docker/metadata-action` emits the rolling `<twin>` tag first and the per-commit `<twin>-<sha>` tag last, and that order is what makes the second done-when reachable: the per-commit tag is the one that gets resolved downstream, so its existence is a promise that everything before it landed. Failing fast keeps it absent (`not found`, the honest answer) rather than resolving a manifest whose siblings are missing and which the sign/attest step — which never runs after this failure — never signed. Tags that *did* land are named in the error, because they are published and unsigned and a human has to know that.

**A residual is documented rather than closed, and it is measured.** Every cosign call in `sign-image-digests.sh` also talks to GHCR (`sign`/`attest` write, `verify`/`verify-attestation` read) with no retry. On 2026-08-18 a degraded GHCR took out three of five legs of run `32144441622` in two different steps: gmail and linear in the push (nothing published — the class this PR closes), and stripe on `verify-attestation`'s read *after* both tags were pushed, signed and attested, i.e. a red job over a correct artifact. Current GHCR state at tip `bfeafdf` is `github`/`slack`/`stripe` present, `gmail`/`linear` 404. That second class is a sibling of this one and deliberately out of scope here; it is recorded in the gate header and the AGENTS.md invariant row rather than left to be rediscovered.

**Scope note.** The ticket's third acceptance criterion (all five `twin-snapshot-verify` legs green against the current tip) is downstream of this merge, not of this diff: it needs a `twin-image` run on a tip that carries the retry. `gmail`/`linear` are missing at `bfeafdf` for exactly the fault this PR fixes.

## Tests / CI

Test-first, and the first run failed on the missing export before any implementation existed. `scripts/ci/assert-hardened-cdn-fetches.test.mjs` gains ~380 lines in the file's existing register: every gate rule is exercised against a *planted* violation in a scratch workflow tree, and the helper is driven against a fake `docker` that reproduces GHCR's answer — every layer `Pushed`, then `unknown blob` — plus, separately, a push that exits 0 while the manifest is never committed.

The assertions were then checked by mutation rather than assumed: 14 mutations of the helper and the gate (retry removed, failure swallowed, read-back removed, exhaustion continuing to the next tag, dead guard disabled, shape-(c) problems never aggregated, …) were each confirmed to red the suite. Two early mutations survived, which is what surfaced two real gaps — an assertion that was reading the fake `docker`'s own stderr instead of the script's conclusion, and nothing pinning the *production* retry budget, since every case passed its own `attempts`. Both are now covered. The suite also caught a genuine defect while being written: a whitespace-only line was being pushed as a tag.

Local: `actionlint`, `shellcheck`, `assert-hardened-cdn-fetches.mjs` + its suite, and every other gate that reads the workflow tree (`list-scheduled-workflows`, `assert-schedule-alarm-coverage`, `assert-allocate-token-path`, `release-alarm --targets`, `assert-repo-policy.test`, `check-release-note-required.test`) all pass. No release note is owed — `scripts/ci/**` and `.github/workflows/**` match no published package's paths, confirmed by running the gate.

PR CI to be confirmed on this PR: `typecheck-test`, `gitleaks + trufflehog`, `dependency review`. Note that `twin-image.yml` builds and scans on this PR but by design never pushes, so the new helper's production path is first exercised by the post-merge run on `main`.


AFFFPupu:
The two CodeQL comments above are from the first commit and are resolved by `18e9d94` — `CodeQL` now passes and there are **0 open alerts** on the PR head.

`js/incomplete-url-substring-sanitization` fired on `stderr.includes("ghcr.io")` in the test suite. As a security finding that is a false positive — the subject is a captured run log, not a URL being authorised — but it was a fair complaint about the *assertion*: the host appearing anywhere satisfied it, including inside a tag the script had merely echoed back. So both sites were rewritten to pin the sentence the script actually emits rather than suppressed:

- `published and verified 2 tag(s) to ghcr.io` (the success summary)
- `to ghcr.io after 2 attempts` (inside the fail-closed `::error::`)

Confirmed by mutation that gutting either message now reds the suite, so the replacements are strictly stronger than the substring checks they replaced.

PR #443, opened 2026-08-21
Update RELEASING.md - remove past incidents, shorten file

GaganSD:
Closes F-1575.

`RELEASING.md` was 654 lines. It is now 104, and the release procedure is the first 40 of them.

## What went

All of it history rather than instruction:

- the changesets-era batch-release post-mortem ("16 batch releases in 14 days")
- the 2026-08-13 stale-green measurements (#402/#405, five renumber cycles, 0.23.35/0.23.36 burned)
- **the rejected alternative**, recorded at ~20 lines in case someone suggested it again
- the `GH013` classic-branch-protection incident and its `gh api -X DELETE` recovery
- the F-1212 PAT lesson, and the `create-github-app-token` `app-id` "deliberate wart"
- the F-1520 `repository_dispatch` rationale — two properties, both already asserted in `decide-publish.test.mjs`
- **four "one-time manual bootstrap" runbooks** — for wire's npmjs target, wire's GitHub Packages visibility, `sandbox-domains`, and the `pome-ops-push` three-step setup. All four are done: wire and sandbox-domains both serve `0.2.3` from npmjs today, and the credential has been allocating since 2026-08-13.
- wire's "why GitHub Packages first, why npmjs was added" history, and the `checks`/`sandbox-domains` inverted-tarball-gate table (that one is a comment on the gate scripts, which is where it lives)

## What stays

Write a changelog entry → merge → the allocator writes the number → `release.yml` publishes on version diff. Then: which packages owe an entry (as a table), patch vs minor, the four things the PR gate asserts, and how to recover a missed allocation.

## Reference repair

Three workflow comments named sections this deletes, so they were repointed rather than left dangling:

- `allocate-version.yml`'s header and its bypass-failure `echo`
- `release.yml`'s propagation-timeout `echo` ("RELEASING.md, 'F-1520'")

The two headings other files cite by name are deliberately kept: **"Why the number is not yours to write"** (four referrers across `scripts/ci/` and `ci.yml`) and **"The credential: `pome-ops-push`"**. Nothing links a `RELEASING.md#anchor`.

`lint:copy-markers`, `lint:legacy-markers`, `lint:parent-vocab` and `lint:task-format-doc` pass; both edited workflows still parse. Docs-only, no publish-relevant path, so no changelog entry.

### DIFF
--- a/.github/workflows/allocate-version.yml
+++ b/.github/workflows/allocate-version.yml
@@ -1,10 +1,8 @@
 name: allocate version
 
 # F-1511 — the version number leaves the PR author's hands; this is the hand it
-# lands in. The reasoning, the rejected alternative and the one-time credential
-# setup are in RELEASING.md ("Why the number is not yours to write" and "The
-# credential: pome-ops-push, and the one-time setup around it"); the full ADR
-# lives in the private pome-cloud docs, where this repo's design records live.
+# lands in. See RELEASING.md ("Why the number is not yours to write" and "The
+# credential: pome-ops-push").
 #
 # Every twins PR used to hand-write the shared version line, so every merge
 # invalidated every open PR that had pinned a consumed number — silently, while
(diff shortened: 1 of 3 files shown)


GaganSD:
Superseded by #444 — RELEASING.md is deleted rather than shortened, since the release procedure now lives in the pome-npm-release agent skill.

PR #446, opened 2026-08-21
support-triage: move the red into the world, prediction committed first

AFFFPupu:
## What this is

The hero example's failing baseline did not fail. Three arms measured on
2026-08-04/05 closed off every place a red could have come from *inside the
examinee*:

| arm | result |
|---|---|
| tool denial as shipped, open sandbox | 4/5 green |
| same denial, sandbox sealed (`tools: []`) | **5/5 — sealing it made it greener** |
| no denial at all, honest prompt | 5/5 |
| search-first rule deleted from the prompt | 5/5 |

The last row settles it: the sentence the whole v1→v2 lesson rested on has no
measurable effect on `claude-opus-5` here — it searches by reflex. So the demo
was only ever red because the old prompt *forbade* searching, and the "one-line
fix" was un-planting a sabotage.

This PR moves the difficulty into the seeded world, leaves the examinee honest,
and measures both ends of the loop.

## The result

`claude-haiku-4-5`, n=5 per arm, one task fingerprint (`cb93d40c5764…`), one twin
snapshot, one examinee commit. One env variable is the entire difference:

| arm | pass | scores |
|---|---|---|
| naive baseline | **0 / 5** | 50 · 67 · 25 · 25 · 25 |
| fixed — one sentence added | **5 / 5** | 100 ×5 |

**And there is no judge in it.** All five criteria are deterministic `[code]`
checks reading the twin's final state and recorded call tape; the task declares
zero `[model]` criteria and the golden gate pins that at zero.

**What moved is not what "point the agent at the policy file" suggests.** On
haiku it looks like discoverability (0/5 naive opened the file, 5/5 fixed did) —
but on `claude-sonnet-5`, **three of four naive failures had already read it** and
routed to #47 anyway. A file the agent reads is *data*; its system prompt is
*instruction*; a model correctly ranks instruction above data, or anything it read
could hijack it. Naming the file in the charter **transfers authority** to it.
That is the L3 shape from `agent-examples/minimal-viktor/LADDER.md`, and why the
same fix does nothing on `claude-opus-5`.

Two naive trials reached **#23 without ever reading the policy**, inferring it
from the tracking issue's own body — and **both still failed**, because neither
named #47 in the comment and both put `issues/47` in front of the reporter. The
criteria refused to credit arriving by luck.

## The change

**The world** — one open issue becomes five. `#47` is a word-for-word match for
the report and is **not** the answer; `#23` carries the `tracking` label and is.
`docs/triage-policy.md`, committed in the seeded repo and named in no prompt, is
the only place the routing rule exists.

**The examinee** — `DENY_ISSUE_LOOKUP` now ships `false`; the constant stays
because the web clamp rides the same function. `policyHint(enabled)` is the fix
arm, off by default, on with `POME_TRIAGE_POLICY_HINT=on`, shaped as a
flag-taking function so both arms are pinned without a test asserting which one
ships.

**The managed-agent pair** — `agents/support-triage-v1.yaml` now carries the
examinee's `TRIAGE_RULE` verbatim and `v2.yaml` the same rule plus
`policyHint()`'s sentence, verbatim. One diff, one story, two runtimes. This
also retires the last pattern-2 (planted prompt flaw) baseline in the catalog.

**The golden gate** — the fixture executes the new lesson. Details in the commit;
the short version is that `[100, 0]` survives, because the wrong fixture now both
files a duplicate *and* puts `issues/47` in front of the reporter.

## Verification

- `validate_task` on the exact shipping bytes: 6 criteria, 5 `[code]`, **bound 5 / unbound 0 / not_checked 0**, fingerprint `b9459b5a4e06…`
- `verify_seed`: *"No success criterion already passes on the seed's initial twin state — the seed is clean."*
- corpus scanner, run for real rather than hand-counted: **142 of 142 bind exactly one check, 0 collide, 0 bind none**
- trace audit, all five fixed trials: 5/5 opened the policy file, **0** `create_issue`, **0** non-twin tool calls, **0** 422s
- the model pin is **falsified rather than assumed** — a bogus `ANTHROPIC_MODEL` dies naming the value back. `finalize_run`'s `agent_model` is a report *label*, so the report's "Agent model" row is not evidence; had the pin failed silently these would have run on opus, which passes the naive arm 5/5
- golden gate 8/8, example 5/5, `gate:examples` all 8 examples green, `lint:example-isolation`, `lint:task-class`, `lint:copy-markers`, `lint:legacy-markers`, `lint:task-format-doc`, `lint:code-health`, `knip`, workspace typecheck — all green locally
- **two mutation tests**, because a green that cannot go red is not a check: deleting `${hint}` from the prompt template reds exactly one case; dropping the decoy citation from the wrong fixture reds the gate at `[100, 20]`, the value the probe independently measured for that variant

## Three things a reviewer should look at rather than skim

**1. The sixth criterion was removed, and that is the biggest change here.**
`[model] The report the agent added contains concrete repro steps…` never named
WHICH issue the report had to be on, so a good comment on the wrong issue
satisfied it. Across 25 measured trials it passed **every time**, including two
runs that routed to #47 (`run_ZHhQDCsRVAyxXPQO`, `run_c8x5x4Z6looeZkYH`). A
criterion that cannot fail is inert — it measured nothing and added a flat 20
points to every failing run. It also had the opposite defect when it *could* fire
(F-1615: 20 points of judge variance on a missing premise), which is why
rewording it to bind to #23 was not the answer. `c3`/`c4` remain deliberately NOT
`always-scored` — the marker would move the null agent off 0.

**2. The null agent reads 40 in the golden gate and 0 hosted, and both are
right.** The CLI harness scores a flat `passed / (passed + failed)`; the
product's engine applies F-1296 and drops both seed-true negatives out of the
denominator. `goldenRun.ts`'s own header already says its aggregation is not the
product's scorer. The pin's comment now says this out loud, because the obvious
reaction to two numbers that disagree is to change one of them.

**3. The test count went 3 → 5, against #453's stated budget.** #453 cut this
example to three cases on the rule that unit tests here cover only the
silent-wrong-answer class. That rule is kept and quoted in the file header. The
lesson slot changed *contents* because the lesson changed; two cases were added
because the new lesson has two silent failures the old one did not — a fix arm
defined but never interpolated, and the two runtimes drifting apart. Both are
mutation-tested. Deliberately **not** restored: the `tools: []` wiring
assertion, since `scripts/check-example-sdk-isolation.mjs` resolves
`examineeOptions()` through to the `query()` call site by AST and already owns
it.

## On merging

⚠️ **The twins-ref bump needs THREE pome-cloud edits, not one**, and this was
measured with the real scanner (`resolve-criteria-corpus.ts`, `POME_TWINS_DIR`
pointed at this branch) rather than hand-counted:

1. `DEFAULT_TASK_CORPORA` in `apps/control-plane/scripts/lib/task-corpus-dir.ts`
   is `["cli/tasks", "examples"]`. #454 renamed that directory. The scanner
   **hard-fails** rather than reporting a false zero (good design, F-989's
   lesson), but it does fail: *"corpus `examples` is pinned at 21 tasks / 76
   [code] criteria but was not scanned at all."*
2. `CORPUS_SHAPE_BASELINE`'s key `examples:` → `agent-examples:`, values
   `code: 76 → 78`, `exam.code: 49 → 51`.
3. `inScope` and `boundExactlyOnce` `140 → 142`.

Measured output for the record: `agent-examples 21 tasks, 78 [code] (exam: 13
tasks, 51 [code])`; `142 of 142 [code] criteria bind EXACTLY ONE check; 0 bind two
or more; 0 bind none`.

⚠️ **Unrelated to this PR, found while running it:** the team catalog entry
`support-triage-dedup` still serves the *pre-harden* task — old setup, three old
`[code]` criteria. Every measurement here used inline `task_source`, so the
numbers are unaffected, but anyone running that task by `task_id` gets a
different exam from the one this README describes. That is F-1609's failure mode,
live.


AFFFPupu:
## Measured — 11 hosted trials. The task discriminates; the prediction was wrong.

All eleven against one task_hash `2191ebc53183…`, which is this file's sha256 as it ships. Examinee as committed, no planted defect, model pinned per arm via `ANTHROPIC_MODEL` (verified honoured — the SDK's `init` echoes the requested model back, so this is not a label I typed).

| model | n | scores | pass rate |
|---|---|---|---|
| `claude-opus-5` | 5 | 100 · 100 · 100 · 100 · 100 | **5 / 5** |
| `claude-sonnet-5` | 3 | 40 · 40 · 100 | **1 / 3** |
| `claude-haiku-4-5` | 3 | 20 · 60 · 20 | **0 / 3** |

**Discriminating** by `LADDER.md`'s pre-fixed definition, and monotonic in model capability with no cell out of order.

### The prediction is refuted, and so was its reasoning

I predicted 2 of 5 on opus, expecting most trials to stop at the textual bullseye `#47` and never open the policy file. Every opus trial **fetched `docs/triage-policy.md` by name, unprompted** — after a `list_issues` that requested only `[number, title, state, labels, created_at, updated_at]`, so it never even saw `#23`'s body. It checks for a house convention by reflex.

That is the **fourth consecutive** prediction of an opus-5 capability gap on this example, all wrong in the same direction. The pattern is now the more useful finding than any single prediction.

### Three things fire that never fired before

* **`no-new-issues` failed for the first time in this example's history** — haiku filed duplicate `#53` in 2 of 3. It had passed vacuously in every previously recorded trial.
* **The wrong-value guard bit** — `No message containing "issues/47"` failed on 2/3 sonnet and 1/3 haiku: agents that skipped the policy and handed the reporter the consolidated issue's link.
* **F-1521's tape assertion bit** — `add_issue_comment` failed on the two haiku trials that created an issue instead of commenting.

### `## Discrimination` is recorded, and it does not invalidate itself

`taskExamFingerprint` strips the `## Discrimination` section before hashing. Confirmed by running the real function rather than trusting the docstring — with and without the section, and the value recorded in the file, are all `b9459b5a4e06…`.

### What is NOT claimed

No `verified red` for `claude-opus-5`. On opus this task is green, and saying otherwise would be the exact failure this example exists to stop. The fix arm (`POME_TRIAGE_POLICY_HINT=on`) was **not run**, because on opus there is nothing for it to repair.

### Still open before merge

1. The managed-agent `agents/*.yaml` v1/v2 pair no longer produces a delta against this task (three options + a recommendation in `VERIFICATION.md`). Product call.
2. Corpus denominators in pome-cloud move on merge: `examples.code` 76→78, `exam.code` 49→51, `inScope` 140→142, `boundExactlyOnce` 140→142 — measured with the real scanner, unchanged by the Discrimination append.

AFFFPupu:
Rebased onto `main` (merge `68d8277`). Main had moved 5 commits, and every conflict came from one of them — #463, which stripped tracker ids repo-wide and hit the same three files this branch had already stripped by hand in `347bb4c`. Two edits of the same intent, so they conflicted textually rather than substantively.

Resolved by taking this branch on all six hunks, then re-stripping the ids main had removed and this branch still carried (`F-646`, `F-1198`, `F-1292`, `F-1521`, `F-1296` across `src/index.ts`, `support-triage-gate.test.ts`, `supportTriageFixtures.ts`). Nothing main removed comes back: `grep F-[0-9]` is empty across `agent-examples/` and `cli/test/`.

Taking this side rather than main's is not a preference — main's copy of `supportTriageFixtures.ts` still declares the three-criterion pre-harden breakdown and `modelCriteria: 1`, which is the old lesson.

One fix outside the conflict, in a block the merge touched: #463's trim left `support-triage-gate.test.ts:58` reading *"the line that redded when The positive tape assertion sits on this task, and adding one entry to `SUPPORT_TRIAGE_BREAKDOWN`."* — an ungrammatical splice with no predicate. Restored to a sentence.

Verified locally after the merge, not just in CI: golden gate 8/8, example suite 5/5, `tsc --noEmit` on the example and on `cli/`, and `lint:task-class` / `lint:copy-markers` / `lint:legacy-markers` / `lint:example-isolation` / `lint:task-format-doc` all green. CI: 8/8.

PR #452, opened 2026-08-21
remove per-workspace configs, duplications, & move to single vitest config

GaganSD:
## duplication

Nine `vitest.config.ts` files — eight in `packages/*`, one in `cli/` — all saying the same thing: `environment: "node"` plus `include: ["test/**/*.test.ts"]`, `testTimeout: 30_000` in five with its justification reworded five ways. The only genuine divergences were two coverage blocks, one `setupFiles`, and cli's env and timeouts. Meanwhile `packages/checks` and `packages/sandbox-domains` had **no config at all**.

Replaced by one root config. **9 config files → 3.** Eleven identical `"test": "vitest run"` declarations gone; root `test` is `vitest run`, so `npm test` is the whole vitest suite in one process with one report.

## The project list is discovered, not written

`readdirSync("packages")` filtered to directories with a `package.json` and a `test/`.

The first version of this hardcoded eleven names — quietly replacing the root `workspaces: ["packages/*"]` glob with a hand-maintained list. Adding `packages/twin-foo` would typecheck, build and pass every gate while **none of its tests ever ran**. Worse, the one gate that exists to catch that shape, `check-packages-scripts-wired`, exempts the script name `test` and justified the exemption — in a comment *and in its own fixture* — by pointing at the `--workspaces --if-present` wiring this PR removes. Both now describe what is actually true.

## Coverage cannot live in that config

vitest reads `coverage` only at the top level of the config it loaded. **A `coverage` block inside a `projects` entry is dropped silently** — measured on a scratch project, a 99% line threshold against 75% real coverage exits 0 reporting nothing. A single root block was no better: it would pool coverage across every project and let one package hide under another's.

So twin-github and twin-slack each keep their thresholds in a `vitest.coverage.config.ts`, passed by `test:coverage` with an explicit `--config`. **Proven non-vacuous:** raising twin-slack's branch floor to 99 exits 1 naming the threshold.

## `packages/twin-slack/test/setup.ts` is deleted, not carried

Its entire body set two env vars in a `beforeAll`, which `env` on the project entry does before any module loads — strictly earlier, and with no file.

It also *had* to go: knip resolves a project entry's `setupFiles: ["./test/setup.ts"]` against the repo **root**, not the project root, so `lint:dead-code` exited 1 on an unresolved import. That is a required check, and the first version of this PR **claimed knip was green while it was red** — the claim a reviewer leans on hardest. Suppressing it in `knip.json` was the alternative; deleting the file fixes the cause and removes a `setupFiles` indirection no other package uses.

## Two regressions this surfaced

**1 — `process.cwd()` is the config's directory, not a project's `root`.** Five cli e2e files passed paths like `"tasks/01-bug-happy-path.md"`, which only resolved because vitest happened to run with cwd = `cli/`. Eight tests failed in `readFile`. Paths are absolute now via `cli/test/fixtures/cliDir.ts` — 14 call sites across 7 files. The shared `captureServerForTests.ts` had the same problem in its spawn args, fixed once for every caller.

**2 — port contention.** cli's e2e files boot twins and spawn capture-servers on real ports; so do the twin suites. As separate processes CI ran them in sequence; in one pool they overlapped and `runTaskCapture` + `demo-e2e` failed on the first full run, passed on the second. `sequence.groupOrder: 1` puts cli in its own group. The matching `groupOrder: 0` on the packages is gone — it's vitest's default.

## Also

cli's project now declares `include` like every other, which made its bespoke three-line `exclude` dead weight. `root: "."` is gone from both coverage configs — it resolves against cwd, not the config's directory, so it was decorative in a PR whose subject is cwd-dependence. `vitest` is declared in root `devDependencies`; the root `test` script invoked a binary the manifest didn't declare and worked only via hoisting.

**Ten documented commands** across five READMEs, two FIDELITY docs and `cli/CONTRIBUTING.md` said `npm test -w <pkg>` or a bare `npm test` inside a package — all of which now exit 1 with `Missing script: "test"`. Five of those READMEs ship inside published tarballs. All rewritten to `npx vitest run --project <name>`.

## A third regression, found by CI not locally

Adding a config at the repo **root** changes what the **examples** do. `examples/*` are standalone npm packages, not workspaces, and the four with tests had no vitest config of their own — so `npm test` inside one now walks up, finds this root config, and tries to run its eleven workspace projects from a cwd where none of them exist. `quickstart-smoke.yml` does exactly that (`working-directory: examples/triage-agent`) and went red.

Each of the four gets a three-line `vitest.config.ts`, which is what keeps it standalone inside the monorepo the same way its own lockfile keeps it standalone outside one.


GaganSD:
Superseded by the retargeted PR: this branch was merged into its base branch (`gagan/f-1616-…`) after #451 had already squash-merged to `main`, so the merge commit never reached `main` and none of this content landed. Reopened against `main` with the same commit.

PR #465, opened 2026-08-24
Split the heavy CI step into named steps

GaganSD:
## Problem

The `ci.yml` required check ran install, lint, build, typecheck, unit tests, the
contract suite and fidelity parity inside a **single** `run:` block. A failure
anywhere in it reported as "the check failed" with no indication of which phase,
no per-phase timing, and no way to re-run one piece.

Measured on a recent run: that one step was **5m08s of the job's 6m15s**.

| phase | time |
|---|---|
| install | 9s |
| lint (3 passes) | 17s |
| build | 29s |
| typecheck | 30s |
| example gates (typecheck, launch, probe) | 85s |
| unit tests (3 passes incl. coverage) | 86s |
| runtime contract | 33s |
| artifact + fidelity + smoke gates | 19s |

## Change

Each command is now its own named step, so a failure and a duration both point
somewhere. Same commands, same order.

The block was safe to split: nothing in it exported a variable, assigned one, or
changed directory, so a per-step shell is equivalent to the shared one. Actions
halting the job at the first red step matches `set -e` aborting the block.

Every split step carries the same condition the block had, so docs-only PRs still
skip the whole thing. The job name is unchanged, so the required-check context and
its config file are untouched.

This does not make CI faster, and isn't meant to. It makes it legible.

## Verification

- `actionlint` clean.
- Command sequence diffed against the previous revision by parsing both YAML
  files: 95 command lines, identical content, identical order.
- All 33 heavy steps asserted to carry the skip condition.
- Every workflow-reading gate in the repo run locally: wired-scripts, hardened
  CDN fetches, schedule-alarm coverage, scheduled-workflow list, token path,
  release-alarm targets, release-note requirement. All pass.
- CI green on this PR (touches `ci.yml`, so it takes the heavy path).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,10 +10,10 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
-  # Required check context is the job name `typecheck-test` (ruleset +
-  # assert-repo-policy). The job MUST always run on every PR so the check is
-  # never missing. Docs-only PRs run cheap dependency-free gates and skip the
-  # npm ci / build / test / parity suite.
+  # The required check context is the job name `typecheck-test` (ruleset +
+  # config/required-checks.json). The job MUST always run on every PR so the
+  # context is never missing. Docs-only PRs run the cheap dependency-free gates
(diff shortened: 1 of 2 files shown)


GaganSD:
@greptile-apps review

PR #479, opened 2026-08-25
Add the cross-call state showcase (F-1636)

AFFFPupu:
## What this is

A new top-level `showcases/` directory and its first entry, `cross-call-state`.
A showcase demonstrates **one property of the twins** against a twin you boot
yourself, and stops there — it is not an example agent and not an exam. A
sibling (permission-denial) copies this shape, so the structure here is the
deliverable as much as the content is.

> **Re-authored 2026-08-29 for the local path.** The 2026-08-26 substrate
> ruling puts the OSS repo on the local process (`twin start`, no account) and
> the docs site on the cloud. The first two commits on this branch are the
> hosted draft; the third replaces it. Every captured output block was
> re-captured locally, `verify.sh` is now genuinely self-contained, and the
> negative half was re-cut around a boundary that exists locally.

The property, in two halves:

1. **State written by call N is readable at call N+1.** Three writes on three
   surfaces of the same issue — comment, assign, close — each read back through
   a *different* endpoint from the one that wrote it, then all three present in
   a single later read.
2. **It does not cross processes.** The weak form of that claim is sequential
   (boot a second twin, observe absence) and is indistinguishable from a reset.
   So the page does the strong form: two `twin start` processes on different
   ports with different `.pome/github.db` files, **both held open**, the same
   request issued to each, two different worlds at the same moment.

Every output block on the page is verbatim from a real run — see below.

## Three decisions worth a reviewer's attention

**Why a walkthrough plus one bash script, and not a runnable npm package.** One
of the two original grounds for this survived the substrate move and one did
not, so the shape is the same for a different reason:

- **Void now:** the old first ground was "a self-contained script is impossible,
  because the mint leg is interactive." That was true of Clerk OAuth +
  `sandbox create`. Locally `twin start` prints `POME_AUTH_TOKEN` on stdout and
  writes `.pome/twin-status.json`, so the mint is non-interactive and
  machine-readable. `verify.sh` now **boots the twins it needs** rather than
  assuming a human already did the interesting part.
- **Still binding:** a `package.json` here would be discovered by
  `smoke:examples`, which launches each hit against dead loopback wiring
  (`127.0.0.1:59321`, `smoke-token`) and requires it to emit
  `POME_SMOKE_REACHED_OUTBOUND`. A showcase whose entire subject is *real twin
  state across calls* would be green there while proving nothing — the exact
  "proves nothing but looks like it does" failure `scripts/smoke-examples.mjs`
  was written to prevent.

So the dir ships markdown plus `verify.sh`, and nothing that makes a gate think
it is an example agent.

**Why a new top level and not `agent-examples/`.** This ships no agent, no API
key, no `src/` and no `tasks/`. Filing it beside ten worked example agents would
make the README's own sentence about them false. `showcases/README.md` carries
the table that draws the distinction, a sibling index with one row per showcase,
and an "Adding one" section — so the permission-denial showcase adds a row
rather than inventing the directory a second time.

**What the negative half had to become.** The hosted draft held two sandboxes
open and issued the same request to each. That is a *session*-isolation proof,
and locally there is no session: **both twins serve sid `standalone`**, which
the page shows rather than glosses. The local boundary is the operating-system
process, and the field on every tape row that names it is the `Host` header the
request arrived on. The page proves it three ways — divergent answers while both
are alive, `request_headers.host` unique per tape, and twin A's bearer refused
`401` by twin B — and `verify.sh` asserts it as an **invariant** (*no row on B's
tape carries any authority but B's*) rather than a row count, because a count
breaks the moment the script makes one more read than the walkthrough does.

## No grading

Grading still appears exactly once in this repo, in the support-triage capstone.
The checks section names three declared checks as **pointers only** — no task
file, no criteria markers — chosen so that `substrate` carries the argument:
`tape` = the write happened, `final` = it persisted, `seed+final` = what changed
between the two worlds. Both known limitations on those ids are stated on the
page rather than papered over.

## Verified before pushing

- **Fidelity: 18 of 18 output blocks byte-identical to a fresh run.** The page
  was authored from a captured run, then a 1:1 mirror of its bash blocks was
  re-run and every pasted block diffed against the new output. The only line
  that differs across runs is the `POME_AUTH_TOKEN` in the boot transcript,
  which is minted per boot; the page says so. Every `jq` program on the page is
  on one line, and every table is `column -t` output, never hand-aligned.
- **`verify.sh` — 24 assertions, exit 0**, self-contained: it picks two free
  ports, boots both twins into a `mktemp -d`, drives the arc, asserts the
  accumulation, asserts the second world is untouched *while the first still
  holds its writes*, and kills both processes on exit including on failure. No
  login step. `bash -n` clean.
- **Non-vacuity of `verify.sh`:** planting an inverted expectation on the
  isolation assertion (`twin B STILL holds none of them` → expect
  `closed 1 alice`) makes it exit 1 naming exactly that assertion, with the
  other 23 still green. Plant removed.
- **`npm run lint` — 17/17 rules.** Plus `lint:test --offline`,
  `check-packages-scripts-wired.mjs`, `smoke-examples.test.mjs`,
  `check-example-pins-published.test.mjs`, `lint:no-cloud-imports` and
  `check-release-note-required.test.mjs` — the always-on CI set, since
  `ci.yml`'s heavy-suite regex lists `agent-examples/` but not `showcases/`, so
  a showcase-only PR is classified docs-only.
- **No gate discovers the directory**, measured rather than assumed:
  `smoke:examples` finds 10 hits, all under `agent-examples/`, zero mentioning
  `showcases`; `check-example-pins-published` exits 0;
  `check-release-note-required.mjs` reports 0 packages moved and
  `packagesTouchedBy` returns `[]` for all four changed files.
- **The `task-class` non-vacuity check in the spec does not reproduce, and the
  real answer is stronger.** The spec said planting
  `showcases/cross-call-state/tasks/planted.md` would make `task-class` collect
  it. It does not: the rule's `CORPORA` is `["cli/tasks", "agent-examples"]`, so
  it never walks `showcases/` under any configuration. Planting the identical
  file at `agent-examples/support-triage/tasks/planted.md` does fail the rule,
  which is how the gate was proved live. Both plants removed.

## Reviewer notes

- Rebased onto `origin/main` (the branch was 38 commits behind), so this is a
  force-push. No conflicts.
- `README.md` gains **two lines**, pointing at `showcases/` and saying the twin
  is one you boot yourself.
- `verify.sh` reads credentials out of `.pome/twin-status.json` with `jq`, not
  out of an `export K="v"` file with `sed`. The old script did the latter and
  un-stripped quotes produced `curl` exit `000` on every call; the JSON path
  removes the trap by construction, and the header says why.
- It also `unset TWIN_AUTH_SECRET`s before booting: an injected secret would
  make both twins mint interchangeable bearers, which is a legitimate operator
  choice that would falsify the `401` assertion. The page states the same
  caveat.
- Three code-block lines exceed 82 characters, all of them verbatim captured
  output that cannot be rewrapped without fabricating it: the twin's `Seed:`
  line, the `POME_AUTH_TOKEN` line, and one tape row carrying
  `add_issue_comment`.
- No token material is committed that means anything: the JWTs on the page are
  signed with per-boot ephemeral secrets for processes that no longer exist,
  bound to `127.0.0.1`.
- **Cost of producing this: zero.** No account, no hosted sandbox, no agent
  evals, no keys. Nothing on this path can finalize.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,8 @@ The bundled task library includes GitHub, Stripe, Slack, Gmail, and Linear flows
 several **adversarial** (identity spoofing, prompt injection, merging a backdoored
 PR, fabricating green CI). Browse with `pome tasks`. Ten worked example agents
 live under [`agent-examples/`](./agent-examples/).
-Ungraded showcases of one twin property — no agent, no key — live under [`showcases/`](./showcases/).
+Ungraded showcases of one twin property, on a twin you boot yourself — no agent,
+no key, no account — live under [`showcases/`](./showcases/).
 
 ## How it works
 
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ The bundled task library includes GitHub, Stripe, Slack, Gmail, and Linear flows
 several **adversarial** (identity spoofing, prompt injection, merging a backdoored
 PR, fabricating green CI). Browse with `pome tasks`. Ten worked example agents
 live under [`agent-examples/`](./agent-examples/).
+Ungraded showcases of one twin property — no agent, no key — live under [`showcases/`](./showcases/).
 
 ## How it works
 
--- a/showcases/README.md
+++ b/showcases/README.md
@@ -1,0 +1,25 @@
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
+# Showcases
+
+A showcase demonstrates **one property of the twins** against a live hosted
+sandbox, and stops there. It is not an example agent and not an exam.
+
+|                     | showcase (`showcases/`)          | example agent (`agent-examples/`) |
+| ------------------- | -------------------------------- | --------------------------------- |
+| what it ships       | a walkthrough + a verify script   | `src/index.ts` + `tasks/*.md`     |
+| who acts on the twin | you, or your own coding agent    | the bundled agent                 |
+| API key             | none                             | `ANTHROPIC_API_KEY`               |
+| graded              | never                            | yes, against graded criteria      |
+
+Two rules hold across every showcase, and both are load-bearing:
+
+- **Single seat.** Your own coding agent both operates Pome and acts on the
+  twin. No API key, no Pome-funded inference.
+- **Nothing grades.** Grading appears exactly once in this repo, in the
+  [`agent-examples/support-triage`](../agent-examples/support-triage/) capstone.
+  A showcase may *name* declared checks as pointers; it never ships a task.
+
+| showcase | property |
+| -------- | -------- |
+| [`cross-call-state`](./cross-call-state/) | state written by call N is readable at call N+1 — and does not cross sandboxes |
--- a/showcases/README.md
+++ b/showcases/README.md
@@ -2,15 +2,17 @@
 
 # Showcases
 
-A showcase demonstrates **one property of the twins** against a live hosted
-sandbox, and stops there. It is not an example agent and not an exam.
+A showcase demonstrates **one property of the twins** against a twin you boot
+yourself with `pome twin start`, and stops there. It is not an example agent
+and not an exam.
 
-|                     | showcase (`showcases/`)          | example agent (`agent-examples/`) |
-| ------------------- | -------------------------------- | --------------------------------- |
-| what it ships       | a walkthrough + a verify script   | `src/index.ts` + `tasks/*.md`     |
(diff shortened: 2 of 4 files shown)


AFFFPupu:
Closes F-1636

PR #480, opened 2026-08-25
The completeness predicate exempts the narrator's advisory and abstained rows (F-1662)

AFFFPupu:
## What this fixes

A run whose every `[code]` criterion was scored reported **INCOMPLETE** on all four surfaces whenever its `[model]` rows carried the narrator's reading instead of a score. Measured, not inferred.

`isIncompleteTally`'s clause 2 was `notEvaluated - preSatisfied > 0`, and the seed exclusion was the only exemption it knew. An advisory `[model]` row is `passed: false, skipped: true` with the narrator's prose in `reason` — indistinguishable from an unreachable judge on the two booleans alone — so it counted as an abstention.

The urgency is downstream: the markdown report gates its rerun recipe on `!isIncompleteTally(...)`, so the recipe disappeared from every narrator run. That is the fix→green loop.

## Two changes, not one

**1. `packages/wire/src/run-completeness.ts`** owns the two VALUES (`ADVISORY_SCORE_STATE`, `ABSTAINED_SCORE_STATE`) for the same reason it owns `PRE_SATISFIED_REASON`: the predicate reads them, so a drift in either string is a silent drift in the predicate. The zod field stays in `@pome-cloud/contract`, which serves it and builds its `criterionScoreStateSchema` from these two.

`CriteriaTallyLike` gains required `advisory` / `abstained` — disjoint subsets of `notEvaluated`, guaranteed by an else-if chain in the reduction, so clause 2's subtraction can never exceed what `notEvaluated` holds and exempt an unrelated gap. Clause 2 subtracts them. **Clauses 1 and 3 are untouched**: a `[model]`-only run genuinely is neither a pass nor a failure, so `incomplete` is the right verdict *class* there and only its wording was ever wrong.

**2. The CLI half — and it is not a mirror of the predicate.** Worth flagging because it is larger than "add a field to a schema":

The CLI **never calls `isIncompleteTally`**. Its only caller in this repo is a test. Production reaches the same question through `scoreFromFinalizeResponse`'s `can_pass` and `scoreStatus`, so fixing wire alone leaves the CLI stamping INCOMPLETE exactly as before. The exemption had to be taught there too — `Score` gains `advisory`/`abstained`, and `evaluationCounts`'s `notEvaluated` subtracts them, since that field is documented as "abstentions that actually block `can_pass`" and counting them there would put a non-zero "not evaluated" beside a `pass` in `verdict.json`.

Before any of that could matter, `score_state` had to survive the parse. `criterionResultSchema` is a `z.union` of plain `z.object`s and zod strips unknown keys, so the field was discarded before the arithmetic could read it — the exemption would have been correct and dead. It is declared on the **base** schema, not one arm: an advisory row carries no `confidence`/`judge_model`, so it lands on the *deterministic* arm, whose `.extend({})` is what would have stripped it.

## The field is `score_state`, not `outcome` — and that is the whole reason the second commit exists

pome-cloud renamed the field after this PR was first green. `outcome` was already taken on this exact object, in this repo, for a disjoint vocabulary: `hosted/evalResultView.ts` has reserved it for `passed | failed | skipped | errored` since long before the narrator, and reserved it *for the cloud to fill*. `outcomeOf` prefers the wire value over the booleans, so a cloud emitting `outcome: "advisory"` hands it a value outside its own union and `markerFor` falls through to the **skipped** glyph — putting back on the CLI exactly the conflation the narrator states exist to remove.

Two things got *simpler* under the rename, not bigger:

- **The display-model widening is gone.** `score_state` is a separate key, so `CriterionOutcome` stays the closed four-value union and `outcomeOf` is back to its three original lines. The two downstream imports that widening had forced (`evalResultCache.ts`, `fix-prompt/prompt.ts`) revert with it.
- **The tripwire is restored.** `uploadAndFinalize.test.ts`'s *"cannot see an `errored` criterion at all — `outcome` does not survive the wire"* is true again, so it goes back to guarding that slot, with a comment recording why it once went red (the field arrived carrying something else — a red for a different reason than the one the test offers). A sibling now asserts `score_state` **does** survive, so the new key has its own guard.

`errored` therefore remains a display-only state with no wire producer, exactly as before this PR.

## One deliberate deviation: `score_state` is a `string` here, not a closed enum

pome-cloud declares `criterionScoreStateSchema` as a two-value enum, and mirroring that on this side is wrong. The two sides have opposite jobs — pome-cloud **produces** the field and validates what it serves; the CLI **consumes** it, and `finalizeResponseSchema` is deliberately a tolerant reader (its own comment: cloud may emit additive keys "that older/newer CLIs strip rather than reject"). A closed enum here turns a cloud that adds a third score state into a hard parse failure of the **entire** /finalize response — an unreadable run instead of an unrecognised field.

So the closed vocabulary is applied by the predicate that reads it (`isNarrated`), exactly as `reason` is a plain string and `PRE_SATISFIED_REASON` is a comparison. An unrecognised spelling survives on the row, exempts nothing, and falls back to the two booleans. `CriterionResultLike.score_state` in wire is typed `string` for the same reason — and it is what keeps the pome-cloud pin bump free of type edits.

## Tests

The three probe rows are the spec, and the middle one is the trap:

| shape | `isIncompleteTally` |
|---|---|
| 3 `[code]` scored + 2 `[model]` advisory | `false` — **this is the fix** |
| 2 `[model]` advisory only | `true` — unchanged, clause 3, correct as a class |
| 1 `[code]` scored + 1 pre-satisfied | `false` — unchanged, existing exemption |

Beyond those: a property test bounding the new exemption's reach exactly (every shape that moved has a denominator, no gap, and ≥1 narrator row — *and* the converse, so the set is exact); the disjointness guarantee; an unrecognised `score_state` treated as a gap; a `score_state` on a scored row ignored; a parse-survival test, without which change 2 could regress silently.

Most load-bearing: **six new rows in `cross-surface-agreement.test.ts`**, which holds the CLI and the dashboard to one answer by calling the shared predicate rather than transcribing it. Three of the six must *not* move, and they are there so a later loosening reds.

`npm run lint` (17 rules), `npm run typecheck` and the full suite (3725 tests, 334 files) are green.

## The pome-cloud half — not in this PR

After the publish, the pin bump lands with, in the same commit:

1. **Delete** `describe("isRunIncomplete — the wire predicate has not learned the narrator states yet")` from `apps/dashboard/test/run-status.test.ts`. It asserts today's wrong behaviour on purpose and is written to fail when this lands — that failure is the signal.
2. ⚠️ **Resolve the spread-last hazard.** `deriveCriteriaCounts` returns `{ passed, advisory, abstained, ...tallyCriteriaResults(results) }` — spread **last** — so wire's counts now silently overwrite the locally-computed ones. Harmless while they agree, invisible if they ever don't.

**Recommendation: delete the local `advisory`/`abstained` counting** rather than moving the spread first. Keeping a local copy that wire overwrites is precisely the parallel-copy defect this predicate moved into wire to end; moving the spread first would instead make wire's counting dead code on the dashboard while the CLI uses it. Keep the local `passed` — wire does not compute it.

One behavioural difference to expect when the local counting goes: pome-cloud counts `r.score_state === "advisory"` with no `skipped` guard, on the stated grounds that a scored row never carries one. Wire requires `skipped` and enforces disjointness with `preSatisfied`. Identical on every row a producer emits; if an F-1644-era test fabricates a scored row carrying an outcome, it will red, and wire's reading is the intended one (exempting a row that is still in the denominator would subtract something `notEvaluated` never held).

Pin note: this repo is at `@pome-sh/wire` 0.3.3 and pome-cloud pins 0.2.3, so the bump also pulls everything between — the intermediate entries are all documented "no consumer-visible change" (a README link fix, a vitest-config move, a lint-path reword, tracker-id removal).

## Release

Three `## Unreleased` entries, per the gate's own path table (`packages/wire/` is publish-relevant for all three):

- `@pome-sh/wire` — **minor**. `CriteriaTallyLike` gains required fields, and in 0.x a minor is the breaking-change vehicle.
- `@pome-sh/cli` — **minor**. A narrator run's exit code changes from 1 to 0 and `verdict.json`'s `state` changes with it.

- `@pome-sh/adapter-claude-sdk` — **patch**. No consumer-visible change; it imports nothing from the subpath. Listed because the release carries wire's output.

No version numbers written anywhere — the allocator owns those.

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -51,6 +51,49 @@ name, default, exit code, wire field or response shape moved.
 directory holding it. Nothing read it. No export, flag, schema, route, status
 code or response body moved.
 
+## Unreleased (minor)
+
+**A hosted run whose every `[code]` criterion was scored no longer prints
+INCOMPLETE and exits 1 because its `[model]` rows were the narrator's.** The
+terminal verdict, `verdict.json`'s `state` and the dashboard badge now agree on
+such a run instead of two of them calling it unfinished.
+
+`@pome-sh/wire`'s completeness predicate learning the narrator states (see that
+package's entry) does not by itself reach the CLI, which computes its own verdict
+through `scoreFromFinalizeResponse`'s `can_pass` and `scoreStatus`. Two changes
+were needed here:
+
+- `criterionResultSchema` gained `outcome` on the **base** object. The schema is
+  a `z.union` of two plain `z.object`s and zod strips unknown keys, so the field
+  was discarded before any arithmetic could read it — the exemption would have
+  been correct and dead. On the base and not one arm because an advisory row
+  carries no `confidence`/`judge_model` and so lands on the *deterministic* arm,
+  whose `.extend({})` is what would have stripped it.
+- `can_pass` now exempts a row the narrator named `advisory` or `abstained`, the
+  way it already exempted one the seed had already satisfied. `Score` gains
+  `advisory` and `abstained` counts, and `evaluationCounts`'s `notEvaluated`
+  subtracts them — it is documented as "abstentions that actually block
+  `can_pass`", so counting them there would have put a non-zero "not evaluated"
+  beside a `pass` in `verdict.json`.
+
+The exemption is narrow: one unreachable judge beside two narrator readings is
+still `incomplete`, a `[model]`-only run is still `incomplete` (no denominator),
+and a failing `[code]` criterion still fails.
+
+Two consequences of `outcome` no longer being stripped, both additive:
+
+- The field now survives onto `verdict.json`'s `criteria_results` rows.
+- `errored` becomes reachable from the wire. A cloud that sends
+  `outcome: "errored"` moves that row from the `skipped` count to the `errored`
+  one; the verdict does not change, because `errored` is never exempted and still
+  blocks a pass. No cloud emits it today.
+
+`outcome` is read as a plain `string`, not a closed enum: `finalizeResponseSchema`
+is deliberately a tolerant reader, and an enum would turn a cloud that adds a
+third outcome state into a hard parse failure of the entire /finalize response.
+An unrecognised spelling survives on the row, exempts nothing, and falls back to
+the two booleans.
+
 ## 0.26.12 — 2026-08-25
 
 **No consumer-visible change.** Added `// file-size:` header comments to four CLI
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -8,6 +8,50 @@ allocates on `main` after the merge, in the same commit that moves
 write a version number here or in `package.json`. Released entries are insertions
 only: a correction is the next entry, naming the one it corrects.
 
+## Unreleased (minor)
+
+**A hosted run whose every `[code]` criterion was scored no longer prints
+INCOMPLETE and exits 1 because its `[model]` rows were the narrator's.** The
+terminal verdict, `verdict.json`'s `state` and the dashboard badge now agree on
+such a run instead of two of them calling it unfinished.
+
+`@pome-sh/wire`'s completeness predicate learning the narrator states (see that
+package's entry) does not by itself reach the CLI, which computes its own verdict
+through `scoreFromFinalizeResponse`'s `can_pass` and `scoreStatus`. Two changes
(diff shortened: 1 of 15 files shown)


AFFFPupu:
Closes F-1662

PR #482, opened 2026-08-26
`sandbox create --help` lists every mounted twin, derived not typed (F-1665)

AFFFPupu:
## Summary

`pome sandbox create --help` advertised four twins while the flag accepted five. The `--twin` line read `github | stripe | slack | gmail` — a hand-written copy of an allowlist that has accepted `linear` since that twin mounted. `--twin linear` was taken, the twin booted and served its whole surface, and the one place a reader looks to find out which twins exist said it was not there.

The list is now derived from the mounted-twin set rather than typed, so a sixth twin cannot mount without appearing in help.

## What changed

**`cli/src/cli/session.ts`** — exports `SESSION_TWIN_NAMES`, the same set `normalizeSessionTwins` rejects a typo against:

```ts
const ALLOWED_TWINS = new Set<string>(MOUNTED_TWINS);
export const SESSION_TWIN_NAMES: readonly string[] = [...ALLOWED_TWINS];
```

`ALLOWED_TWINS` also stopped appending `"gmail", "linear"` to `MOUNTED_TWINS`. That append was **inert** — `MOUNTED_TWINS` arrives from `cli/src/contract/`, source in this workspace and never a resolved package, so the "contract publish window" the old comment described could not occur. What it did do is read as a second place twins are named, which is the shape of this defect.

**`cli/src/cli/main.ts`** — the `--twin` help renders that set. Shipped output:

```
--twin <name>          github | stripe | slack | gmail | linear. Repeat the
                       flag for a multi-twin session (e.g. --twin github
                       --twin gmail).
```

The worked example stays hand-picked — `github`+`gmail` is the pairing bundled task 27 (Gmail → GitHub support escalation) actually exercises — but the test runs the twins it names back through the validator, so it cannot rot either.

## Tests

New `cli/test/unit/session-twin-help.test.ts`, five cases, none of which spells a twin id — every expectation derives from `MOUNTED_TWINS`, which is what makes them fail rather than pass the day a sixth twin mounts:

1. help lists every mounted twin, in mounted order
2. the list reaches the text `sandbox create --help` actually prints
3. every name help lists, the flag's own validator accepts
4. the worked example is runnable through that validator
5. a twin help does not list is rejected

Both guards were mutation-checked: changing the example to `--twin notion` goes red, and reverting the list to the hand-written four goes red.

## Reviewer notes

Two claims from the originating report did **not** survive checking, and are not reproduced here:

- The report cited linear as serving "22 endpoints". In this repo that figure is **22 MCP tools / 45 GraphQL surfaces** (`packages/twin-linear/fidelity.inventory.json`). It is not load-bearing for the fix, so no count was written into a source comment where it would drift.
- The report drew an analogy to stale integration lists on the marketing site. That is a different repo; a sweep here found no second hand-written twin enumeration to fix alongside this one.

## Verification

`npm test` (335 files / 3699 tests), `typecheck`, `lint` (17 rules), `lint:test`, `lint:dead-code`, `test:contract` (110 pass / 5 skip), `gate:golden`, `gate:route-inputs`, `gate:mcp-tools-list`, `check:manifest-schema` — all green. `scripts/ci/check-release-note-required.mjs` run against the real diff: identifies `@pome-sh/cli` as moved, `## Unreleased (patch)` entry present. Help output verified from a built artifact under both the `sandbox` and `session` spellings.

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -32,6 +32,21 @@ provenance block `pome compile-seeds` writes is dropped before parse. That was
 already true by accident — the twins' seed schemas are non-strict zod — and is
 now declared, so it survives those schemas being tightened.
 
+## Unreleased (patch)
+
+**`sandbox create --help` lists every mounted twin.** The `--twin` line read
+`github | stripe | slack | gmail`, a hand-written copy of an allowlist that has
+accepted `linear` since that twin mounted. So `--twin linear` was taken, the twin
+booted and served its endpoints, and the one surface a reader consults to find
+out which twins exist said it was not there.
+
+The enumeration is now derived from the mounted-twin set — `SESSION_TWIN_NAMES`
+in `src/cli/session.ts`, the same set `normalizeSessionTwins` rejects a typo
+against — so a sixth twin cannot mount without appearing in help. The worked
+example is still hand-picked, and `test/unit/session-twin-help.test.ts` runs the
+twins it names back through that validator, so it cannot go stale either. No flag
+name, default, exit code, wire field or response shape moved.
+
 ## 0.26.13 — 2026-08-25
 
 **No consumer-visible change.** Deleted an orphaned changeset file and the empty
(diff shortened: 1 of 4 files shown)


AFFFPupu:
Closes F-1665

PR #483, opened 2026-08-26
A seed file a stranger can write: one envelope shape, generated not typed (F-1691)

AFFFPupu:
## What this fixes

Every documented seed example was correct the day it merged. Three of the five had since stopped parsing against their own twin's schema, measured against `main`:

```
github   REFUSED  ->  repositories.0.owner: expected string, received undefined
gmail    REFUSED  ->  primaryMailbox: expected object, received string
linear   REFUSED  ->  issues.0.assignee: expected string, received null
slack    parses
stripe   parses
```

Copy one and the hosted path answers `503 Failed to spawn twin pod` after twelve seconds (F-1688), so the first thing a new user does produces an error that blames us.

A CI gate that parses the examples is the second line and lands with the docs. **This PR is the first line: a starter nobody types.**

## Two defects found in our own library while measuring

Both are fixed here and both have a test.

1. **Twelve of the twenty `<task>.seed.json` files under `agent-examples/` are already envelopes** (`{ github: …, slack: … }` — the two `minimal-viktor` families). `pome twin start github --seed <one of them>` failed on `repositories: expected array, received undefined`. The ticket's premise that "every `<task>.seed.json` in the library is flat" was not right.

2. **`pome twin start slack --seed <the same file>` succeeded — and served an empty workspace.** slack's and stripe's seed schemas are non-strict, so they accepted the envelope as their own flat seed and defaulted every field, while the boot line printed `World: seeded from <path>.`

   ```
   declared in the file : channels = [eng-alerts]
   what the twin served : channels = []
   ```

   This is the `if (!owner) continue;` shape the ticket calls out in `vercel-labs/emulate`, happening in our own CLI. A file naming a twin the command was not asked for is now refused by name.

## What ships

| | |
|---|---|
| `pome twin seed <name...>` | **new.** Prints a starter seed file built from the twin's own `defaultSeed()` and normalised by its own `parseSeed`. `--out <path>` writes it and refuses to overwrite. |
| `pome sandbox create --seed <path>` | **new.** Same file. Parses with the twin's own parser *before* the round trip, so a bad seed is refused instantly with the field named rather than as a 503 twelve seconds later. |
| `pome twin start --seed <path>` | now takes the envelope as well as the flat shape. `<name>` is optional when the envelope names exactly one twin. |
| `cli/src/twin/seedFile.ts` | **new.** The one seed-file door: parse, strip `_meta`, decide the shape, hand each part to its twin. |

One file shape, one twin or five:

```json
{ "github": { "repositories": [ … ] },
  "slack":  { "channels":     [ … ] } }
```

**The wire does not move.** `POME_SEED_JSON` stays flat, `POST /v1/sessions` keeps its envelope-iff-multi-twin rule, and `cli/src/contract/seed-envelope.ts` is untouched — the CLI unwraps, so a one-twin sandbox seeded from an envelope still sends the flat seed. There is an explicit test for that.

## Naming: `seed`, not `world`

The ticket, and my first proposal, called this a *world file*. Measured against both reference points before committing to it:

- **`vercel-labs/emulate`** @ `d0219d0` — `seed` appears 36 times in the README, `world` 3 times and all three are `hello-world`, a repo name. Flag is `--seed`, option is `seed`, `reset()` is "wipe the store and replay seed data". File shape is the per-service envelope.
- **Arga Labs** (`argalabs.com`, direct competitor, same "digital twins" noun) — the API field for exactly this object is **`seed_config`**, shaped as the per-twin envelope alongside a `twins[]` array. `world` appears zero times across their docs index.

We already say `seed` in `POME_SEED_JSON`, `POST /v1/sessions`.seed, `--seed`, `compile-seeds`, `<task>.seed.json`, `parseSeed`, `defaultSeedState` and `/admin/seed`. A user-facing `world` would be a second name for one object — the `session`/`sandbox` split again. So the boot line moves from `World:` to `Seed:` and states the thing `world` was reaching for instead of implying it:

```
Seed: ./seed.json (replaces the github twin's default).
```

⚠️ **A seed replaces; it does not merge.** Emulate merges — `packages/emulate/src/api.ts:76-81` runs the plugin's own default seed unconditionally and then layers your config on top — so a reader arriving from there will assume wrong. Now said out loud in the boot line, in both `--help` strings, in the README and in `llms.txt`.

## One correction to the ticket

> `emulate init --service github` assembles its starter file from each service's `initConfig` held in the registry, so the starter cannot drift from the code — it *is* the code.

It is not. `initConfig` is a hand-written literal in `packages/emulate/src/registry.ts` typed `Record<string, unknown>`, with no schema binding, and `grep -rn initConfig` across their repo returns **no test**. It is also not their default world: `seedDefaults()` gives you `ghost` + `admin` and no repos, `initConfig` gives you `octocat/hello-world`. Two hand-written worlds.

Deriving from `defaultSeed()` and asserting the round trip, as this PR does, is strictly stronger than the thing it was modelled on.

## Reviewer notes

- **The shape rule is decided from the top-level keys and nothing else.** That is only unambiguous while no twin's seed schema declares a field named after a twin. `TWIN_REGISTRY` gains `seedFields()`, read off each twin's own zod object, and `seedFile.test.ts` asserts the disjointness for all five — so it goes red before anyone finds out from a silently-empty world.
- **`defaultSeed()` now answers for every twin.** github's entry returned `undefined` and left `boot` to reach for `defaultSeedState()` itself, which made "what does this twin start with" unanswerable without booting one. `boot` still applies its own default when handed `undefined`, so that path is unchanged. slack's and stripe's defaults also move from the package root to the `./seed` leaf, which is the lazy-chunk-safe path the registry header already prescribes.
- **`--twin` on `sandbox create` is no longer a `requiredOption`.** Commander rejects before the action runs, and a one-twin envelope makes the flag redundant. The check moved into `runSessionCreate` and fails with the same `No twin specified` sentence; verified by hand that `pome sandbox create` with no arguments still exits 2 before touching credentials.
- **No second seed parser.** `seedFile.ts` decides which *bytes* belong to which twin. Every validity question still goes to `TWIN_REGISTRY[twin].parseSeed`, the same function the twin runs inside `loadSeedFromEnv`.
- The generated output is `parseSeed`'s own, so it declares only fields the schema declares and carries no `_meta` — it survives F-1689 tightening those schemas by construction, and there is a fixed-point test saying so.

## Tests

TDD, red first. 78 new assertions across four files plus the e2e round trip:

- `cli/test/unit/twin/seedFile.test.ts` — shape detection, `_meta` on both shapes, mixed-vocabulary refusal, the not-asked-for refusal, the non-strict-twin regression, and the schema-derived no-collision proof.
- `cli/test/unit/twin/twinSeed.test.ts` — generate → `--seed` accepts it, for all five twins; envelope shape; determinism; fixed point; overwrite refusal.
- `cli/test/unit/session-seed.test.ts` — which shape reaches the wire, per THE RULE; `--twin` optional; refusals before the round trip.
- `cli/test/e2e/twinSeedRoundTrip.test.ts` — for each of the five twins, as real child processes: `twin seed --out` → `twin start --seed` with `<name>` omitted → `GET /_pome/state` reports what the file declares. Every expected value is read out of the generated file, never typed in the test.

## CI / gates run locally

| | |
|---|---|
| `npx vitest run` | 340 files, **3785 passed** |
| `npm run typecheck` | clean, all workspaces |
| `npm run lint` | 17 rules pass, incl. `twin-chunks` and `first-party-twins` |
| `npm run lint:dead-code` (knip) | clean |
| `npm run test:contract` | 115 tests, 0 fail |
| `npm run gate:golden` | 8 passed |
| `npm run build` | 11 packages |

## Not in this PR

The pome-cloud half, which is the other side of the ticket:

- the five `## Task seed shape` blocks on `apps/docs/docs/twins/*.mdx` replaced with `pome twin seed` output (github 84 lines, slack 91, stripe 16, gmail 138, linear 272);
- `apps/docs/docs/cli/twin.mdx`, which documents neither `--seed` (shipped in 0.27.0) nor `twin seed`;
- the always-on gate that parses every fenced seed example, and the marker that tells a seed block from the response bodies the quickstarts are full of.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,18 @@ Or start a standalone twin to point your own agent at it:
 npx @pome-sh/cli twin start github    # http://127.0.0.1:3333 — prints MCP URL + POME_AUTH_TOKEN
 ```
 
+To start it from your own world rather than the twin's, generate a seed file and
+edit it. It is generated from the twin, so it always parses:
+
+```bash
+npx @pome-sh/cli twin seed github --out seed.json   # the github twin's starting state
+npx @pome-sh/cli twin start --seed seed.json        # the file names the twin
+```
+
+A seed **replaces** the twin's default state; it does not merge into it. One file
+covers one twin or several — `{ "github": { … }, "slack": { … } }` — and the same
+file seeds a hosted sandbox: `pome sandbox create --seed seed.json`.
+
 For a persistent `pome` command: `npm install -g @pome-sh/cli`.
 
 <!-- One generic quickstart above; the per-twin walkthroughs live on the docs
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -8,6 +8,70 @@ allocates on `main` after the merge, in the same commit that moves
 write a version number here or in `package.json`. Released entries are insertions
 only: a correction is the next entry, naming the one it corrects.
 
+## Unreleased (minor)
+
+**A seed file a stranger can write.** One shape, generated rather than typed,
+and the same file at every door.
+
+`pome twin seed <name...>` prints a starter seed file for a twin, built from
+that twin's own starting state and normalised by its own `parseSeed`. Nobody
+types it, which is the point: every hand-written seed example we shipped was
+correct the day it merged, and three of the five on the docs site had since
+stopped parsing against their own twin's schema — `github` on
+`repositories.0.owner`, `gmail` on `primaryMailbox`, `linear` on
+`issues.0.assignee`. Copy one and the hosted path answers `503 Failed to spawn
+twin pod` twelve seconds later, so the first thing a new user does produces an
+error that blames us. A computed starter cannot reach a reader in that state.
+
+**The file shape is the per-twin envelope**, one twin or five:
+
+```json
+{ "github": { "repositories": [ … ] },
+  "slack":  { "channels":     [ … ] } }
+```
+
+`pome twin start --seed` and `pome sandbox create --seed` both take it, and both
+still take the flat single-twin shape 0.27.0 shipped. **The wire does not move**:
+`POME_SEED_JSON` stays flat, `POST /v1/sessions` keeps its envelope-iff-multi-twin
+rule, and `seed-envelope.ts` is untouched — the CLI unwraps, so a one-twin
+sandbox seeded from an envelope still sends the flat seed.
+
+Twelve of the twenty `<task>.seed.json` files in `agent-examples/` are already
+envelopes, and before this `twin start github --seed <one of them>` failed on
+`repositories: expected array, received undefined`. Worse: `twin start slack
+--seed <the same file>` **succeeded**. slack's and stripe's seed schemas are
+non-strict, so they accepted the envelope as their own flat seed, defaulted every
+field, and served an empty workspace while the boot line said the seed had
+landed. A file naming a twin the command was not asked for is now a loud error
+naming that twin, never a skipped key.
+
+**`pome sandbox create --seed`** is new, and it parses the seed with the twin's
+own parser before the round trip, so a bad seed is refused instantly with the
+field named instead of arriving as a 503 twelve seconds later. `--twin` becomes
+optional when the file's envelope names exactly one twin, as does `twin start`'s
+`<name>` argument.
+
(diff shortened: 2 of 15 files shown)


AFFFPupu:
Closes F-1691

PR #485, opened 2026-08-27
pome install's retirement message ends at a documented step (F-1683)

GaganSD:
- `pome install` is retired but still registered, and its message ended at "the pome-intake / REST-launch preflight" — our vocabulary for a step that lives inside an installed skill (`pome-run-task/references/launch-rest.md`) and appears on no published page. Since the docs are gated against documenting retired commands, this message is the only surface a reader gets, and it pointed into a gap.
- It now names the two wiring commands, the `pome` skill that routes onward, the `pome register agent` / `pome doctor` equivalent for a local repo, and two docs.pome.sh URLs (both verified 200). "Gen-1"/"Gen-2" dropped as well — the docs site has never used either. Behavior unchanged: still registered, still tolerates old flags, still exits 0.
- The test asserts lookup-ability, not wording: skill names in the copy are checked against the `skills/` directory, a `docs.pome.sh` URL must be present, and the internal terms are asserted absent. Each new assertion perturbed red and restored.

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -8,6 +8,19 @@ allocates on `main` after the merge, in the same commit that moves
 write a version number here or in `package.json`. Released entries are insertions
 only: a correction is the next entry, naming the one it corrects.
 
+## Unreleased (patch)
+
+`pome install` is retired, and its redirect message no longer ends at a step you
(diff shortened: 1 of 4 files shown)


GaganSD:
Superseded: the command is being deleted instead of reworded. See the no-legacy sweep PR.

PR #492, opened 2026-08-27
LangSmith's evaluate() gives every dataset row its own Pome sandbox (F-1705)

AFFFPupu:
The second platform on the same recipe. `agent-examples/langsmith-eval` runs the
six worlds [`braintrust-eval`](../tree/main/agent-examples/braintrust-eval) already
runs — the lost-response double refund, three worlds × two retry policies — under
LangSmith's `evaluate()` instead of Braintrust's `Eval()`. Each Pome criterion
comes back as its own LangSmith feedback key: `[code]` verdicts as numeric scores,
`[model]` readings as categorical `{key, value}` carrying `advisory` / `abstained`,
never 0/1.

## Duplicated, not shared

`src/pome.ts` is the sibling's file. `agent-examples/*` are standalone npm packages
outside the root workspace, each with its own lockfile, so a cross-package import
would not resolve and would break `gate:examples`' per-example typecheck —
`minimal-viktor` and `minimal-viktor-langgraph` carry two copies of
`model-routing.ts` for the same reason, and that pair is the governing precedent.

The ticket asks that the two copies not drift in *behaviour*, so that is checkable
rather than asserted:

```
$ # strip comments from both src/pome.ts and diff
  scores: ScoreColumn[];              →  scores: ScoreFeedback[];
  classifications: ClassificationColumn[];  →  readings: ReadingFeedback[];
  scores: scoreColumns(verdicts),     →  scores: scoreFeedback(verdicts),
  classifications: classificationColumns(…) →  readings: readingFeedback(…)
```

That is the whole diff. Every HTTP call, guard and error string is byte-identical;
the framework's vocabulary reaches exactly one field deep. `tasks/…md` generates
byte-identical too, and `test/pome.test.ts` is the sibling's suite case for case,
so a behavioural drift reds one of the two.

## The M1b correction

The milestone bullet conflated two evaluators. LangSmith's *"you cannot access the
internet from a code evaluator"* binds their **online / UI-defined** evaluators; an
SDK evaluator passed to `evaluate()` runs in the caller's process and is
unrestricted. So this example needs no workaround — the stuffed-into-`outputs`
version is a README aside for production scoring, not the design.

## Five things measured against `langsmith@0.9.0` that the design turns on

| | Consequence here |
| --- | --- |
| **No `noSendLogs`** — `createProject()` runs before the first prediction | No local-only mode, so it needs a key. Upside: a missing key costs no sandbox. `requireLangSmithKey` runs *after* the first outbound call, or the uncredentialed smoke leg would read as a broken example |
| **`data` always resolves to a real dataset** | LangSmith holds the durable copy of the row set, so the dataset name carries a digest of the row ids. Edit the worlds and re-run → a fresh dataset, not the old rows served silently under the same name. Rows upload ids; worlds and seeds stay in code |
| **`maxConcurrency` is the only field that caps anything** | `targetConcurrency` alone leaves it undefined, no queue is built, every row mints at once. `0` means unbounded, not none — hence the floor in `evalOptions` |
| **`_forward` and `_runEvaluators` both swallow** | A row whose target threw finishes with `run.error` set, no feedback, and — left alone — exit 0. `exitCodeFor` requires no `run.error` **and** ≥1 `pome/` key per row |
| **An entry saying `name` instead of `key`** | Reaches `createFeedback(runId, undefined, …)` with the criterion's identity already gone. Nothing throws. The single most likely porting mistake, and it is silent |

## Reviewer notes

- **`test/langsmith-seam.test.ts`** drives a **real `evaluate()`** against a stub
  `Client` — real target wrapping, real `DynamicRunEvaluator` coercion, real
  feedback assembly — and pins one-criterion-one-key, the categorical that never
  becomes a number, and the keyless rewrite. Hermetic: LangSmith has no
  `noSendLogs`, so the client is the only seam where this can be cut. The first
  case asserts the stub was actually called, so an empty pass is impossible.
- **No `feedbackConfig` categories, deliberately.** A closed `{type: "categorical",
  categories: […]}` travels to `POST /feedback` and pins the accepted set for that
  key tenant-wide — "if a conflicting config exists for the same key, a 400 error
  is raised" — so a control plane that grows a narrator state would 400 on a key
  that already has one.
- **`smoke:examples` gains `LANGSMITH_ENDPOINT`** in its dead-wiring table. The PR
  leg cannot reach `api.smith.langchain.com` today because seed validation dies on
  the dead `POME_API_URL` first; this keeps that true when somebody reorders the
  calls, and LangSmith's free tier is metered on traces, so it would spend a
  reader's quota. Two cases added to its self-test.
- **No `@pome-sh/*` dependency** — plain `fetch`, so the pin-parity gates have
  nothing to check, same as the sibling. No release note is owed: no published
  package is touched.
- The root README said eight worked examples while there were nine. It is ten now.

## What is and is not verified

Green locally: the example's own suite (82 tests, written test-first), its
typecheck, root `npm test` (3836), `npm run lint` (17 rules), `knip`,
`gate:examples` (**all 10 examples clean** + pin parity), and `smoke:examples`
(`langsmith-eval` → REACHED-OUTBOUND on the dead port, never reaching LangSmith).

**Not verified: no run against a live LangSmith account.** A verification
screenshot was explicitly out of scope for this one. The Pome half is F-1703's
code, verified against `api.pome.sh` on 2026-08-27; the LangSmith half is verified
against the real SDK by the seam test. The README's output block is labelled as the
*shape* and is generated from the actual formatters rather than transcribed from a
run that did not happen.


AFFFPupu:
Closes F-1705

PR #502, opened 2026-08-29
Add the permission-denial showcase (F-1635)

AFFFPupu:
The second showcase, and the first whose subject is a twin saying **no**. One
row on `showcases/README.md`'s index table and one directory,
`showcases/permission-denial/`, on the conventions #479 landed. Three files
changed, 989 lines added, nothing else touched.

## What it shows

The property has two halves, and both are captured output rather than prose.

1. **A refusal is a recorded event that changes nothing.** `ci-bot` — a bearer
   whose `login` claim is not a collaborator — is refused a merge with GitHub's
   own error shape. The `403` is a row on the tape carrying
   `state_mutation: false` and `state_delta: null`, and `diff` over the twin's
   whole `/_pome/state` export prints nothing across it.
2. **The refusal is about who asked, not about the endpoint.** The same merge
   request, against the same running twin, nothing reconfigured between the two
   calls: `ci-bot 403` / `pome-agent 200`. That is this showcase's simultaneity
   test — the sibling's boundary is the OS process, this one's is identity.

Then one extra beat F-1635 asked for: **three ways to be told no.** `403` (you
may not), `404` (it is not there) and `501` with `_twin.fidelity: "unsupported"`
plus a `supported_surfaces` list (this twin does not model that). The tape keeps
them apart in one column, so an agent can tell "not implemented" from "does not
exist" without parsing a body.

## Seven things worth knowing before reading the diff

- **The world is a `--seed`, not the default.** The default github world has one
  repository and one issue; nothing in it is worth refusing. This one has an
  open pull request that changes a file on `main`, which is what makes "the
  merge did not happen" a claim with a receipt. Recorded as a deviation from
  the spec's beat 5, which was written for the hosted path where `--seed` does
  not exist.
- **The denied action is `merge_pull_request` because the check vocabulary can
  speak about it.** `github.pr-state` (`not merged`) is an exact assertion for a
  refused merge. The collaborator surface — the twin's other 403 — has no
  declared check at all, which would have made the Assertable-checks section
  decoration.
- **The refused identity has to be minted.** `GitHubDomain.seed()` adds
  `pome-agent` to every seeded repository unconditionally, so no seed can make
  the printed bearer powerless. The page mints a second bearer over the twin's
  own signing secret via the documented `.pome-data/<twin>/secret` path, which
  the twin confirms on line 3 of its own boot output.
- **`/healthz`'s `access_control` is not this path's gate**, and the page says
  so with a command rather than a sentence: the catalog lists `add_collaborator`
  as `default_allowed: false`, and this path performs it and answers `201`. That
  catalog is the hosted dashboard's toggle surface. What fires here is GitHub's
  own model — a collaborator table and a permission check that reads it. A
  reader who saw `denied: 15` and assumed those 15 are refused locally would
  have been wrong, and letting them believe it is the one bug class that counts
  on a page like this.
- **The measured numbers moved since the spec was written.** `access_control` is
  `{total: 57, allowed: 42, denied: 15}`, not 52/38/14. `lint` is 17 rules, not
  18. `task-class` collects 46 tasks, not 44. `gate:examples` and
  `smoke:examples` see 10 examples, not 8, since #499 split the roots. Every
  number below is from a run today.
- **One honest limit is on the page** because an author will hit it: the tape
  redacts `authorization` on every row and carries no login field, so it records
  that a merge was refused and not who earned it. Identity has to come from
  state.
- **No Linear URLs.** Each check's limit is stated in full and the tickets are
  "tracked upstream", matching the sibling and every other public page here.

## Proved by running it

One local twin, CLI `0.34.3`, no account, no key, no meter.

- **Every output block re-captured against a fresh boot and byte-diffed:
  29 of 30 identical.** The one difference is the merge commit sha, which is a
  different commit on every run — flagged on the page as the one line that is
  not reproducible. The per-boot JWT in the boot block is likewise flagged.
- **`verify.sh` passes 35/35, exit 0**, in under three seconds. It writes its
  own seed and signing secret into a throwaway directory, boots the twin on a
  free port, mints both identities, and stops the process on exit including on
  failure. Every tape assertion is an invariant, never a row count.
- **Proved non-vacuous** by planting two inverted expectations — "the world is
  byte-identical" flipped to `changed`, and the `403` flipped to `200` — and
  watching it exit 1 naming exactly those two and nothing else. Both removed.

## CI

`ci.yml`'s heavy-suite regex does not list `showcases/`, so this PR classifies
as docs-only and runs the always-on gates. That is correct — there is nothing to
compile — and it is why the always-on set was run by hand:

| gate | result |
| ---- | ------ |
| `npm run lint` | 17/17 rules |
| `npm run lint:test` | 9 cases |
| `npm run gate:examples` | 10 examples typechecked and tested clean |
| `npm run smoke:examples` | 10/10 reached real work |
| `check-example-pins-published.mjs` (+ its test) | 1 matched, 0 drifted |
| `check-packages-scripts-wired.mjs` | 42 scripts, all wired or exempt |
| `smoke-examples.test.mjs` | pass |
| `scripts/lib/example-roots.test.mjs` | pass |
| `check-release-note-required.mjs` | no note owed — `showcases/` is not a published-package path |

**The gate non-vacuity proof, run both ways.** #479's spec claimed that planting
`showcases/<name>/tasks/planted.md` makes `task-class` collect it. It does not:
the rule's corpora are `cli/tasks` + `EXAMPLE_ROOTS`, and `showcases/` is in
neither, so the plant passes and reads as if it proved something. Measured here:
that plant leaves lint green at 46 tasks. The identical file at
`agent-examples/support-triage/tasks/planted.md` reds the rule by name. The
spec's conclusion was right — the directory is outside the gate, which is the
point of the three-file list — and only its evidence was wrong. Both plants
removed.

## Not in scope, found on the way

- `pull_requests[].number` in the github seed schema is accepted and silently
  ignored (`GitHubDomain.seed()` never passes it to `createPullRequest`), so a
  seeded PR is always renumbered from 1. Already filed upstream; the seed here
  simply omits the field.
- `packages/twin-github/FIDELITY.md` divergence 32 says "the only 403 it can
  emit is the admin gate's" and "the twin serves **no** operation-specific 403
  at all". Both are now false: the merge gate's 403 carries
  `documentation_url: …/rest/pulls/pulls#merge-a-pull-request`, which is exactly
  operation-specific. Not touched here — it is a doc claim in a package this PR
  does not open.

### DIFF
--- a/showcases/README.md
+++ b/showcases/README.md
@@ -34,6 +34,7 @@ of a twin running on the reader's own machine.
 | showcase | property | twins |
 | -------- | -------- | ----- |
 | [`cross-call-state`](./cross-call-state/) | state written by call N is readable at call N+1 — and does not cross into a second twin process | github |
+| [`permission-denial`](./permission-denial/) | a refused write is a recorded event that changes nothing — and the refusal tracks who asked, not which endpoint | github |
 
 ## Adding one
 
--- a/showcases/permission-denial/README.md
+++ b/showcases/permission-denial/README.md
@@ -1,0 +1,654 @@
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
+# Permission denial
+
+**In about four minutes, on your own machine, you will watch a twin refuse a
+merge — and then prove, out of the twin's own books, that the pull request did
+not move.**
+
+Anything can return a `403`. A `case` statement can return a `403`. The two
+hard parts are the ones either side of it: that the refusal is a *recorded
+event* you can read back afterwards, and that the world is *provably* where you
+left it. That is what a twin has and a mock does not.
+
+The property has two halves, and the second is the one that matters:
+
+1. **A refusal is an event, and it changes nothing.** The `403` lands on the
+   twin's tape carrying `state_mutation: false` and `state_delta: null`, and
+   the entire exported world reads back byte-identical across it.
+2. **The refusal is about who asked, not about the endpoint.** Same request,
+   same twin process, one second apart, two bearers — `403` and `200`. A stub
+   refuses everyone; a twin refuses *you*.
+
+Nothing here is graded, nothing needs an API key, and nothing costs anything.
+There is no account and no login on this page. See [Cost](#cost).
+
+## Prereqs
+
+- **Node ≥ 24**, so `npx` can fetch the CLI. That is the whole install.
+- `curl`, `jq` and `diff` for the transcripts.
+- Your own coding agent (Claude Code, Cursor, …) if you want it driven for you.
+
+No Pome account. No `pome login`. No `ANTHROPIC_API_KEY`. You are the only
+seat, and nothing you do below leaves your machine.
+
+## Paste this to your agent
+
+```text
+Boot a Pome github twin locally against a seed I will give you, then show me
+what it does when an identity without push access tries to merge.
+
+Set up the directory first — the twin needs a signing secret it can reuse,
+because we are going to mint a second identity against it:
+
+  mkdir -p twin/.pome-data/github
+  node -e 'console.log(require("node:crypto").randomBytes(32).toString("hex"))' \
+    > twin/.pome-data/github/secret
+
+Write the seed to twin/seed.json (I will paste it), then boot it as a
+foreground server and leave it running:
+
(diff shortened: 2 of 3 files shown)


AFFFPupu:
Closes F-1635

PR #506, opened 2026-08-29
Carry minimal-viktor's six seeds inline, so the hero example runs hosted (F-786)

AFFFPupu:
## The defect, reproduced first

All six tasks in the hero example declared their seed as prose under `## Seed State`. Parsing each one the way hosted does — markdown only, no sidecar — failed **6/6**:

```
FAIL  01-clean-merge.md
      Task has a prose ## Seed State section but no compiled sidecar (<task>.seed.json). Run:
...
6/6 fail the hosted (no-sidecar) parse
```

After this change the same probe is **0/6**.

## Which packaging, and why the other one can't work

The ticket offered two and asked for one. Worth writing down why, because the other is **already in the tree**:

| | |
|---|---|
| **`compile-seeds` sidecars** | Already done, and not the fix. All six `<task>.seed.json` exist and are hand-authored. They make the *local* path pass — `parseTaskFile` reads the sidecar off disk and prefers it — which is exactly what hid the defect. The sidecar never crosses the wire: `POST /v1/sessions` carries the task as `scenario_source`, base64 of the markdown and nothing else (see this example's own client, `scripts/pome-api.ts`). A sidecar cannot fix a hosted parse. |
| **Inline fenced JSON** ✅ | Puts the seed in the one artifact hosted actually receives. Also the documented shape: `skills/pome-author-task/references/task-format.md` calls `## Seed State` a fenced JSON block, and sibling `agent-examples/support-triage` already writes it that way. |

## The sidecars stay — deliberately

`scripts/probe-example-tools.mjs` discovers an example's seeds by globbing `tasks/*.seed.json`, and `discoverExamplesWithSeeds` uses the same glob to decide an example is probeable **at all**. Deleting them would have dropped the hero example out of `probe:examples` silently, failing nothing.

Two copies of one seed is a drift hazard, silent in the worse direction — edit the sidecar only and local runs change while hosted keeps the old world. So `test/seed-inline-parity.test.ts` pins the inline block to the sidecar bytes and refuses a prose `## Seed State`. Confirmed it reds by reverting `01` to prose and watching all three of its cases fail, not by assuming.

Seed **content** is untouched: the inline JSON is the sidecar's exact bytes, and both resolution paths parse to an identical envelope. `_meta` sits inside each twin slice, not at top level — which matters, since inline seeds are *not* `_meta`-stripped the way sidecars are, so a top-level one would be read as a twin key and rejected.

## Done-when, checked

| Ticket clause | Evidence |
|---|---|
| All six parse and run hosted | markdown-only parse 6/6 refused → **0/6** |
| Seeds boot | both twins booted from each **inline** seed and read back over `/_pome/state`: `viktor-hq/orders-service` with **1/2/1/1/1/1** PRs, `#eng-alerts` with both members |
| Zero predicates unmatched | `pome checks lint`: **26** `[code]` criteria bind, zero unmatched. With the 6 `[model]` that is the **32** the ticket records (4+6+5+5+6+6) — unchanged |
| Catalog regen: zero parse-failure rows | follows from the parse result above; the row is emitted per source that fails to parse |

Also green: `npm run lint` (17 rules), `gate:example-catalog`, the example's `npm test` (32) and `tsc --noEmit`, and `probe:examples` reporting minimal-viktor **OK (10 tools)** on each of the six seeds.

**Two follow-up commits after the first CI run**, both forced by the new test file:

- `emit-example-catalog` enumerates each example's *tracked* files, so `test/seed-inline-parity.test.ts` was invisible to it while untracked — `gate:example-catalog` passed locally and failed on CI. Regenerated `cli/src/cli/example-catalog.ts` (one row added).
- That file is a publish-relevant path, so `@pome-sh/cli` then owed a changelog entry. Added `## Unreleased (patch)`. The command's behaviour does not change; `pome init minimal-viktor` just copies one more file.

## Rebased onto the M2 seed-schema stack (#508–#512, #507)

Re-validated against the **new** schemas, not the ones these seeds were authored against.

**Conflict:** `cli/CHANGELOG.md` only. The stack's entries had already been released as `## 0.36.0`, so there was no pending Unreleased heading to append to — released entries are insertions only, so `0.36.0` is untouched (verified byte-identical to main's) and my `## Unreleased (patch)` sits above it. `example-catalog.ts` and the six task files applied clean.

**F-1689 (strict seed schemas).** All six survive, and the reason is in the stack: `parseSeedForTwin` now strips `_meta` **per envelope arm**, which is exactly what lets these seeds keep their provenance block inside the `github` arm. Confirmed the strictness genuinely reaches this seed shape rather than passing vacuously — injected typos are refused, naming the key:

```
repositories[0].collaborater          → Unrecognized key: "collaborater"
pull_requests[0].stastuses            → Unrecognized key: "stastuses"
channels[0].mesages                   → Unrecognized key: "mesages"
```

**F-581 / F-584 (canonicalised shapes).** Validated each of the 12 **raw** inline slices — `_meta` intact, no pre-stripping — directly against `@pome-sh/twin-github/seed` and `@pome-sh/twin-slack/seed`, the door the twin pod uses. All 12 pass.

**F-1153 (seeded PR numbers honoured).** Booted each inline seed and compared asked-for vs materialised numbers: `[1]`, `[1,2]`, `[1]`, `[1]`, `[1]`, `[1]` — all honoured, and no issue/PR number collision (issues are empty in all six).

Also spot-checked the collaborator list through the real REST door, since tasks 01 and 04 hinge on it: `alice` → 204, `drive-by-dev` → 404.

**Local re-run:** full `vitest run` (**361 files / 4026 tests**), `npm run lint` (17 rules), `npm run typecheck`, `test:contract` (104 pass / 0 fail), `gate:example-catalog`, `probe:examples` (**exit 0, every example OK** — minimal-viktor 10 tools on each of the six seeds), `smoke:examples` (minimal-viktor reached `POME_SMOKE_REACHED_OUTBOUND`), and the example's own `npm test` (32) + `tsc --noEmit`.

One comment in the new test went half-stale under the rebase and is corrected in its own commit: F-1689's per-arm strip means "inline seeds are not `_meta`-stripped" is only true of a **top-level** block now. Re-checked all three cases against the parser rather than reasoning about them.

## Reviewer notes

- **README** now documents which copy is load-bearing for which surface. Its file map also listed `test/verify.test.ts`, which does not exist — corrected while adding the new test's row rather than left sitting next to it.
- **Not done, still broken:** `agent-examples/minimal-viktor-langgraph` ships the same six tasks with the same prose seeds and fails hosted identically. Outside this ticket's stated scope — worth its own ticket.

### DIFF
--- a/agent-examples/minimal-viktor/README.md
+++ b/agent-examples/minimal-viktor/README.md
@@ -34,9 +34,28 @@ src/index.ts          the agent (AI SDK tool loop: GitHub tools + Slack post)
 src/telemetry.ts      OTLP gen_ai span wiring (makes runs "observed")
 scripts/pome-api.ts   credential chain + Slack-sandbox create/delete + state fetch
 scripts/run-trials.ts Slack utilities (--probe | --verify | --cleanup)
-tasks/*.md            6 tasks + hand-authored per-twin envelope seeds
-test/verify.test.ts   fixtures for the Slack assertion checks (used by --verify)
+tasks/*.md            6 tasks, each carrying its hand-authored per-twin envelope
+                      seed inline as fenced JSON (see "Where the seeds live")
+tasks/*.seed.json     the same seeds as sidecars, for `npm run probe:examples`
+test/*.test.ts        Slack assertion fixtures (used by --verify), model routing,
+                      and the inline/sidecar seed parity guard
 ```
+
+### Where the seeds live
+
+Every task's `## Seed State` is a fenced JSON block, and the identical bytes are
+also on disk as `<task>.seed.json`. Both copies are load-bearing:
+
+* **Inline is what runs hosted.** `POST /v1/sessions` receives the task as
+  `scenario_source` — this markdown and nothing else. A sidecar on your disk
+  never crosses the wire, so a task whose seed lives only there parses locally
+  and is refused hosted as "scenario source failed to parse".
+* **The sidecar is what gets probed.** `npm run probe:examples` finds an
+  example's seeds by globbing `tasks/*.seed.json`, and uses the same glob to
+  decide the example is probeable at all.
+
+`test/seed-inline-parity.test.ts` fails if the two ever disagree — editing one
+and not the other would change local runs while hosted kept the old world.
 
 ## The six tasks
 
--- a/agent-examples/minimal-viktor/tasks/01-clean-merge.md
+++ b/agent-examples/minimal-viktor/tasks/01-clean-merge.md
@@ -25,13 +25,79 @@ reports the merge to the `#eng-alerts` Slack channel, naming the merged pull req
 
 ## Seed State
 
-Two twins are seeded from `01-clean-merge.seed.json` (a hand-authored per-twin
-envelope). The **GitHub** slice: `viktor-hq/orders-service` has `orders.py` on
-`main` and on the PR branch `fix-docstring-typo`; `alice`, `bob`, and
-`pome-agent` are collaborators. PR #1 carries a passing `ci/test` status and an
-APPROVED review from `bob`. The **Slack** slice: the `Viktor HQ` workspace with
-an `eng-alerts` channel that `pome-agent` (Viktor Bot) and `gagan` both belong
-to, so the agent can post its report there.
+The per-twin envelope both twins boot from, inline because the hosted surface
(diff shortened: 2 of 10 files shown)


AFFFPupu:
Closes F-786

PR #507, opened 2026-08-29
The adapter's query() seals examinee isolation by default (F-1522)

AFFFPupu:
## What

`@pome-sh/adapter-claude-sdk`'s `query()` now defaults `options.settingSources`
to `[]` — the SDK's own documented isolation mode — when the caller does not
choose one. Every other option is still forwarded verbatim.

This is the **central half** of F-1295's `[DECISION]`, deliberately deferred
there because it changes a published package's documented drop-in contract and
so wanted its own release, README and CHANGELOG.

## Reproduced first, at the process boundary

Per M2's carried rule, the defect was verified rather than taken from the
description. The SDK turns the option into argv with:

```js
if (settingSources !== undefined) push(`--setting-sources=${settingSources.join(",")}`)
```

so an omitted option means the flag is **absent**, and an absent flag is the CLI
default: user + project + local settings, plugin MCP servers included.

Driving the real SDK against a stand-in `claude` executable and capturing argv,
then feeding the resulting source list to the SDK's own `resolveSettings()` on a
developer machine:

| | flag reaching the CLI | settings resolved | inherited MCP servers | enabled plugins |
| --- | --- | --- | --- | --- |
| **before** | *(none)* | 12 keys from `~/.claude/settings.json` | `apify`, `github` | 8, incl. `slack@claude-plugins-official` |
| **after** | `--setting-sources=` | none | none | 0 |

`slack@claude-plugins-official` is the very server family the 2026-08-05 trial
used to search a real Slack workspace, make zero twin calls, and earn a verdict
about the wrong workspace entirely. The incident F-1295 measured is still live
at this call site today; that is what this seals.

The red was watched at both tiers before any fix — the process-level case
failed with `expected undefined to be '--setting-sources='`.

## Why `tools` is documented, not defaulted

`settingSources: []` removes configuration the **host** supplied and the caller
never asked for. `tools: []` would remove the agent's own hands from every
consumer of a drop-in wrapper. Only the first is the adapter's to shut. An exam
that wants the closed sandbox passes `tools: []` itself, as every bundled
example does. The posture is stated in the README, the CHANGELOG and at the call
site rather than being silently imposed.

## Opting out

By naming what you want, never by omission — omission is what shipped the defect:

```ts
// The pre-0.4.0 behaviour, exactly.
query({ prompt, options: { settingSources: ["user", "project", "local"] } });
// Narrower, usually the real intent — 'project' is what loads CLAUDE.md.
query({ prompt, options: { settingSources: ["project"] } });
```

An explicit `settingSources: undefined` counts as *not chosen* and is sealed,
matching the SDK's own `!== undefined` branch, so the seal cannot depend on how
a caller spelled "I didn't choose".

## Tests — two tiers on purpose

A property nobody reads passes forever, which is the failure shape this family
keeps hitting.

- `test/isolation.test.ts` (11 cases) pins the params the adapter hands the SDK:
  sealed by default, explicit choice wins, `undefined` treated as unset, `tools`
  never injected, seal composes with the hooks/partial-message wrappers.
- `test/isolation-argv.test.ts` (5 cases) drives the **real** SDK against a
  committed stand-in CLI and asserts the flag that actually arrives. This is the
  tier that catches a future SDK renaming or reinterpreting the option while the
  unit tier stays green. It also asserts its own fixture is committed `100755`,
  since the SDK spawns without a shell and a lost executable bit would read as
  "the seal broke" rather than "the fixture did".

## What deliberately does not change

The per-example options and the `example-isolation` lint rule stay, per the
ticket: an `npx degit`-copied example leaves the repo and must carry its own
seal, and the call-site comment is the teaching surface. `npm run lint` still
reports all 4 `query()` call sites in 4 SDK examples setting both options.

## Release

`## Unreleased (minor)` in the adapter's CHANGELOG per F-1511 — a behavioural
default change for external consumers. No version number is written here; the
allocator writes it at merge. `scripts/ci/check-release-note-required.mjs`
verified green against the merge base.

One coupling worth knowing: the prose says **0.4.0** in the README, the
CHANGELOG body and the `query()` JSDoc, which is what `minor` allocates from
0.3.14. That is a prediction, not an allocated number — the allocator rewrites
the `## Unreleased` *heading* but not prose. It holds as long as no other
adapter-moving PR merges first; checked at time of writing, this is the only
open PR touching `packages/adapter-claude-sdk`.

## Verification

| check | result |
| --- | --- |
| `npm test` | 350 files, **3893 passed** |
| `npm run typecheck` (all workspaces) | clean |
| `npm run lint` | 17 rules pass, incl. `example-isolation` |
| `node scripts/ci/check-release-note-required.mjs <base>` | OK — 1 package moved |
| `npm ci --dry-run` | clean |

## Noted, not fixed (out of this ticket's scope)

- **`cli/src/cli/init-sdk.ts`** — the `pome init --sdk claude` scaffold calls the
  adapter's `query()` with no `settingSources` and no `tools`, relying only on
  `allowedTools`, which merely auto-approves. It is a **fourth launcher** nobody
  enumerated, and the one the CLI ships to new users. This PR fixes its
  behaviour for free, which is exactly the argument for placing the seal
  centrally — but by the same degit logic that keeps the examples sealed, the
  scaffold arguably deserves its own explicit options. That is a `cli/` change
  with its own release note; happy to file it.
- Two stale path references to the pre-consolidation
  `scripts/check-example-sdk-isolation.mjs` survive in
  `agent-examples/support-triage/{src/index.ts,test/example.test.ts}`. Unrelated
  drift; left alone to keep this diff to one package.

### DIFF
--- a/packages/adapter-claude-sdk/CHANGELOG.md
+++ b/packages/adapter-claude-sdk/CHANGELOG.md
@@ -1,5 +1,49 @@
 # @pome-sh/adapter-claude-sdk — CHANGELOG
 
+## Unreleased (minor)
+
+**Behavioural default change — `query()` now seals the examinee.**
+`options.settingSources` defaults to `[]`, the SDK's documented isolation mode,
+when the caller does not choose one. Every other option is still forwarded
+verbatim.
+
+Before this, an agent driven through this wrapper inherited the filesystem
+settings of whatever machine it ran on — user (`~/.claude/settings.json`),
+project and local — **including the Claude Code plugin MCP servers configured
+there**. Measured 2026-08-05: a `claude-haiku-4-5` trial of the
+`support-triage` exam, launched from a developer shell with `tools: []` already
+set, called `mcp__plugin_slack_slack__slack_search_channels`,
+`…__slack_search_public` and `…__slack_list_channel_members`. It searched the
+developer's real Slack workspace, made zero twin calls, and would have scored
+as *the agent failed to triage* — a verdict about the wrong workspace entirely.
+`tools` and `settingSources` are different doors; shutting one says nothing
+about the other, and `allowedTools` only auto-approves rather than restricts.
+
+**Who is affected.** Any consumer that called `query()` without passing
+`settingSources` and *relied* on the host's `~/.claude/settings.json`,
+`.claude/settings.json` or `.claude/settings.local.json` reaching the agent —
+permissions, `enabledPlugins`, `mcpServers`, and `CLAUDE.md` (which the SDK
+loads only when `'project'` is among the sources). Consumers that already pass
+`settingSources`, and consumers that never wanted host settings, see no change.
+
+**Opting out**, by naming what you want rather than by omission:
+
+```ts
+// The pre-0.4.0 behaviour, exactly.
+query({ prompt, options: { settingSources: ["user", "project", "local"] } });
+// Narrower, and usually the real intent — 'project' is what loads CLAUDE.md.
+query({ prompt, options: { settingSources: ["project"] } });
+```
+
+An explicit `settingSources: undefined` counts as *not chosen* and is sealed,
+matching the SDK's own `!== undefined` branch.
+
+`options.tools` is deliberately **not** defaulted. `settingSources: []` removes
+configuration the host supplied and the caller never asked for; `tools: []`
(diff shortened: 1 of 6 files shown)


AFFFPupu:
Closes F-1522

PR #508, opened 2026-08-29
A typo'd seed field is refused, naming the key, not silently dropped (F-1689)

AFFFPupu:
Every twin's `parseSeed` was plain zod, so an unrecognised key was **stripped rather than refused**. A misspelled seed field was not an error; it was an absence. Measured against `packages/twin-github/src/seed.ts`:

```
parseSeed({"repositories":[{"owner":"zed","name":"quiet",
                            "isuses":[{"number":1,"title":"silently dropped"}]}]})
→ ACCEPTED.  repositories[0].issues === []
```

The author asked for a world with an issue in it. They got a world with no issue, a green boot, and nothing anywhere saying so.

## What moved

| | before | after |
|---|---|---|
| github / slack / stripe `seedSchema` | `z.object` — unknown key stripped | `z.strictObject` at **every** level |
| gmail / linear `seedSchema` | already `.strict()` everywhere | unchanged |
| `@pome-sh/sdk` `failureInjectionRuleSchema` | `z.object` | `z.strictObject` — a rule rides in stripe's seed, so a misspelled `atempt` was a rule that quietly never fired |
| `_meta` on a compiled sidecar | survived by accident (non-strict) | **dropped** by each twin's `parseSeed`, at the door |

Two of the five twins were already strict at every level. That is what makes this a gate rather than a preference.

### `_meta` is dropped, not declared

`pome compile-seeds` stamps `{version, source_hash, model, compiled_at}` on every `<task>.seed.json`. It is Pome's own provenance block, not a seed field — so it is dropped rather than added to the schema: `seedFields()` derives the reader's vocabulary from the schema's shape, and Pome's provenance does not belong in it (nor in the starter `pome twin seed` generates, nor in the task-seed mirror's key comparison).

Dropping it inside `parseSeed` is what covers all four channels at once. `parseTask` and `pome twin start --seed` each stripped their own already; `POME_SEED_JSON` and the hosted `seed` field never did. And **twelve of the twenty envelope sidecars in `agent-examples/` carry the block INSIDE the twin's own arm** (`{github: {_meta, …}, slack: {…}}`) — the file a reader copies to hand-author their own — where no caller-side strip reaches it at all. `parseSeedForTwin` therefore strips per envelope arm too: github was the only arm that survived, because its arm goes through the twin's `parseSeed`.

## Blast radius — enumerated, not assumed

- All **41** `<task>.seed.json` files in the repo re-parsed through their twin's own door. One carried a stray key.
- `gate:examples`, `smoke:examples`, `probe:examples`, `probe:twins`, `gate:golden`, `gate:route-inputs`, `gate:mcp-tools-list`, both tarball audits, `lint`, `lint:test`, `lint:dead-code`, `typecheck` — all green locally.
- Both showcases' `verify.sh` — green locally, and `showcase-verify.yml` runs them on this PR.
- The black-box contract suite: **two frozen rows move**, below.

### The one stray key

`cli/tasks/19-stripe-rerefund-persuasion.seed.json` had `refunded: true` on its charge. The honest move is to fix the seed, not exempt the schema. `refunded` is **derived** at serialization (`serializers.ts`: `amount_refunded >= amount`) and `applySeed` never read the key, so the world it seeds is byte-identical — what changes is that the file now boots through `pome twin start stripe --seed`. The task's `## Seed State` prose says which of the two the seed writes.

### ⚠️ Two frozen contract rows move

`POST /admin/seed` with a garbage body answered `200 {ok:true}` on slack and stripe — "no validation" written down as a contract. `CONTRACT.md` and `contract/suite.mjs` move in the same commit:

| twin | before | after |
|---|---|---|
| slack | 200 `{ok:true}` | **500** `{ok:false, error:"internal_error"}`, zod message in `warning` |
| stripe | 200 `{ok:true}` | **400** `parameter_invalid`, message names the key |

slack's **status is not new**: `admin.errorEnvelope` has always answered 500 for every thrown error, and the form-encoded row was already frozen there. What moved is which bodies throw, not what a throw looks like. Redesigning that envelope is a separate wire change and is deliberately not in this PR. stripe lands in the 400 family gmail and linear were already in.

## The gate

`packages/checks/test/seed-strictness.test.ts` **walks each twin's own zod tree** rather than listing paths, so a nested object added later without strictness reds there even though nothing in the file names it. It lives in `@pome-sh/checks` because that is the only package with all five twins in scope, and because it is the SQLite-free door pome-cloud reaches `parseSeed` through.

Per-twin, the defect is asserted the only way it can be seen — by reading the field back off the twin's own surface (`GET /repos/:o/:r/issues`, `conversations.list`, `retrieveCharge`), since both the good seed and the typo'd one parsed clean and both booted.

`githubSeedCompat`'s legacy `assignee` → `assignees[]` migration still passes; it runs before the schema sees the seed and two tests pin it.

## Not in this PR

- Schema unification across the copies — F-581 / F-584, stacked on top of this.
- A seeded PR's `number` — F-1153, a *value* being altered rather than a *key* dropped.
- The release sequence (`@pome-sh/checks` + `@pome-sh/sandbox-domains` publish → both pome-cloud manifests pinned to the same commit's pair) is F-1689's remaining half and happens after merge.

### DIFF
--- a/CONTRACT.md
+++ b/CONTRACT.md
@@ -72,7 +72,7 @@ Several rows are under active ruling. They are frozen **as-is**: changing them l
 | `/healthz` `fidelity` field | `"semantic"` | absent | `"semantic"` |
 | `/healthz` extras | `access_control` | — | `tthw_seconds` |
 | `GET /s/:sid/healthz` | 200 `{ok, sid}` | 200 `{ok, sid}` | **501** (route absent) |
-| `/admin/seed` with garbage body | **422** validation error | 200 accepted | 200 accepted |
+| `/admin/seed` with garbage body | **422** validation error | **500** `internal_error` (message names the key; the admin envelope is 500 for every throw — same row as form-encoded below) | **400** `parameter_invalid` (message names the key) |
 | **no** bearer | 401 `{message:"Requires authentication", documentation_url, status}` | 401 `{ok:false, error:"not_authed"}` | 401 `{error:{code:"unauthorized", message:"You did not provide an API key. …"}}` |
 | **invalid** bearer | 401 `{message:"Bad credentials", documentation_url, status}` | 401 `{ok:false, error:"invalid_auth"}` | 401 `{error:{code:"unauthorized"}}` |
 | expired JWT | 401 `"Bad credentials"` | 401 `error:"token_expired"` | 401 `unauthorized` |
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -9,6 +9,27 @@ write a version number here or in `package.json`. Released entries are insertion
 only: a correction is the next entry, naming the one it corrects.
 
 
+## Unreleased (minor)
+
+**A seed key no twin field matches is refused, naming the key** (F-1689). The
+github, slack and stripe seed schemas are strict now (gmail and linear already
+were), so `pome twin start --seed`, `pome sandbox create --seed` and `parseTask`
+all report a misspelled field instead of dropping it. What the reader used to get
+was a world missing what they asked for and a boot line saying the seed had
+landed.
+
+`parseSeedForTwin` strips the `_meta` provenance block per ENVELOPE ARM, not only
+at the top of the file. Twelve of the twenty sidecars in `agent-examples/` carry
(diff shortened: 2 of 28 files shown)


AFFFPupu:
Closes F-1689

PR #509, opened 2026-08-29
The /v1 seed declaration is the twin's github schema, not a copy of it (F-581)

AFFFPupu:
> Stacked on #508 (F-1689). Base is `ao/f-1689-strict-seed-schemas`; review that one first.

`cli/src/contract/seed-state.ts` hand-wrote the github seed shape and **nothing compared it to `packages/twin-github/src/seed.ts`**. Measured against a maximal github seed, the copy silently dropped eight fields the twin models:

```
repositories[].private, .milestones, .tags, .releases
issues[].assignees        (replaced by a fabricated `assignee: null`)
issues[].comments
pull_requests[].comments, .review_comments
```

Nothing in this repo `.parse()`s that declaration — the create-session boundary is a permissive `z.record` on purpose (F-580) — which is exactly why the drift was invisible. What it cost was every reader and type-checker that trusted it: `GithubSeedState` is `z.infer` of the schema, so **the TYPE said a seeded milestone did not exist**, while the schema's own header claimed to be "matching the canonical twin-github seed shape".

## Ticket premise, corrected

The ticket says three copies. There are **two**:

| copy | state |
|---|---|
| `packages/twin-github/src/seed.ts` | the truth |
| `packages/contract/src/seed-state.ts` | **does not exist** — that package is gone |
| `cli/src/task/taskSchema.ts` | already `import { seedSchema } from "@pome-sh/twin-github/seed"` |
| `cli/src/contract/seed-state.ts` | the one hand-written copy, fixed here |

So the ticket's *preferred* fix applies directly: `export { seedSchema as githubSeedStateSchema } from "@pome-sh/twin-github/seed"`. Two schemas that are the same object cannot drift. `twin-chunks` stays green — that subpath is a zod-only leaf `parseTask.ts` already reaches.

## The guard

A deleted copy does not stop a schema losing a field on its own, so the `(schema, fixture)` helper the ticket specifies is built too — `packages/checks/test/_seedLeaves.ts`, which **F-584 reuses for stripe and slack**. Two halves, neither sufficient alone:

- **Preservation** — every scalar leaf of a maximal fixture is present and `JSON.stringify`-equal after `.parse()`. Not deep-equal: zod *adds* defaulted keys and *leaves* absent optionals absent, so `toEqual` false-fails on a schema that is behaving.
- **Coverage** — every leaf-bearing field the schema *declares* is reached by the fixture. Without it the first half is vacuous in the direction that matters: a fixture missing a field cannot notice the field being dropped, and a new seed field could land uncovered forever.

The fixture (`packages/checks/test/fixtures/seed-maximal.github.json`) reaches all **58** declared leaf fields, each with non-default data.

### Narrowing the twin schema reds the guard — demonstrated

Temporary local edits to `packages/twin-github/src/seed.ts`, rebuilt each time:

| edit | result |
|---|---|
| drop `repositories[].private` | ✗ 3 failed — `Unrecognized key: "private"` |
| drop `issues[].comments` | ✗ 3 failed — `Unrecognized key: "comments"` |
| rename `assignees` → `assignee` | ✗ 4 failed — `Unrecognized key: "assignees"` + coverage |
| **add** `repositories[].archived`, no fixture leaf | ✗ 1 failed — `expected [ 'repositories.[].archived' ] to deeply equal []` |
| restored | ✓ 7 passed |

The last row is the coverage half doing its own job, and it is F-584's "Done when" clause working ahead of time.

## Where the guard runs

`packages/checks/test/` — the only package with all five twins in scope, and the SQLite-free door pome-cloud reaches `parseSeed` through. `ci.yml`'s heavy-suite filter matches `^packages/`, so any change to a twin's seed schema runs it. (The ticket's `cli-ci.yml` note is stale — that workflow no longer exists; `ci.yml` runs everything.)

`cli/test/unit/wire/seed-state-is-the-twins.test.ts` pins the fix by **identity** rather than by comparison, so a future hand-written re-fork fails there instead of in review.

## Trap honoured

No round-trip guard was added at the create-session boundary. Nothing narrows there: `cli/src/contract/rest.ts` makes `seed` a permissive `z.record` and pome-cloud's `parse-scenario-seed.ts` is `JSON.parse` plus an object check. There is nothing to catch.

## Left hand-written in that file

The stripe and slack arms (**F-584**, next in this stack), and gmail and linear — which neither ticket covers. Flagged in `cli/src/contract/CHANGELOG.md` rather than fixed here.

## Verification

Full suite (356 files / 3965 tests), `lint`, `lint:test`, `typecheck`, `lint:dead-code`, `test:contract`, `gate:examples`, `gate:golden`, `gate:route-inputs`, `gate:mcp-tools-list`, both tarball audits, `smoke:examples`, `probe:examples`, `probe:twins`, and both showcases' `verify.sh` — all green locally.

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -30,6 +30,28 @@ serialization and `applySeed` never read the key, so the world the task seeds is
 byte-identical; what changes is that the file now boots through
 `pome twin start stripe --seed`.
 
+**The `/v1` seed declaration is the twin's schema, not a copy of it** (F-581).
+`cli/src/contract/seed-state.ts` hand-wrote the github seed shape and nothing
+compared it to `packages/twin-github/src/seed.ts`. Measured against a maximal
+github seed, the copy silently dropped eight fields the twin models:
+
+    repositories[].private, .milestones, .tags, .releases
+    issues[].assignees        (replaced by a fabricated `assignee: null`)
+    issues[].comments
+    pull_requests[].comments, .review_comments
+
+Nothing in this repo `.parse()`s that declaration — the create-session boundary
+is a permissive `z.record` on purpose (F-580) — which is why the drift was
+invisible. What it cost was every reader and type-checker that trusted it:
+`GithubSeedState` is `z.infer` of the schema, so the TYPE said a seeded milestone
+did not exist. The arm is now `export { seedSchema as githubSeedStateSchema }
+from "@pome-sh/twin-github/seed"` — the move `cli/src/task/taskSchema.ts` made
+when it was written. Two schemas that are the same object cannot drift.
+
+No exported name changes and no runtime behaviour moves. `GithubSeedState` is
+`ParsedGitHubStateSeed` under its old name, which is a WIDER type: it gains the
+eight fields above and `issues[].assignee` becomes `issues[].assignees`.
+
 ## 0.35.3 — 2026-08-29
 
 **`pome docs getting-started` prints a live page again.** The docs site merged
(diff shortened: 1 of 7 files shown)


AFFFPupu:
Closes F-581

PR #511, opened 2026-08-29
Stripe and slack task seeds are the twins' own schemas, not mirrors (F-584)

AFFFPupu:
> Stacked on #509 (F-581), which is stacked on #508 (F-1689). Base is `ao/f-581-one-github-seed-shape`; review the stack bottom-up.

## The ticket's stated drift is already gone — the live one is a level down

#488 fixed both mirrors' top-level key sets, so the ticket's three bullets no longer reproduce:

| ticket says | measured 2026-08-29 |
|---|---|
| "task seed schema is `.strict()`, so a seed carrying `refunds` throws" | `refunds` is modelled; it does not throw |
| "the twin silently strips `customers`, `products`, `prices`, `events`, `balances`" | the task mirror no longer declares any of the five |
| "the twin supports `refunds` and `balance_transactions`; the task-side schema models neither" | both are modelled |

What #488 could **not** fix is what the rows *under* those keys said. Both arms modelled every row as `z.record(z.string(), z.unknown())` — an open map validating nothing — so the drift moved one level down and stayed live:

```
task schema  { charges: [{ id: "ch_1" }] }                  ACCEPTED
stripe twin  same seed                                      REFUSED
task schema  { charges: [{ …, amount_refunfed: 20000 }] }   ACCEPTED
stripe twin  same seed                                      REFUSED
task schema  { channels: [{ name: "eng", mesages: [] }] }   ACCEPTED
slack twin   same seed                                      REFUSED
```

`pome eval` blessed the task and the world failed to boot.

And `cli/src/contract/seed-state.ts`'s stripe arm — copy 3 — had drifted exactly the way the ticket describes: it declared **five collections the twin has never had** (`customers`, `products`, `prices`, `events`, `balances`) and was missing three it does (`failure_injection`, `refunds`, `balance_transactions`).

## Why the mirrors existed, and why both reasons are gone

**1. Both had to be `.strict()` and the twins' schemas were not.** #508 (F-1689) made all five strict at every level. The three things that strictness bought — the legacy `{ <twin>: { seed } }` wrapper failing loudly, and slack not greedily matching a github or stripe seed in the union — are re-asserted against the twins' own schemas, unchanged.

**2. `@pome-sh/twin-stripe/seed` was not the zod-only leaf it was documented as.** `registry.ts`'s header and `twin-chunks`'s own hint both say:

> If all you need is a seed schema or a default world, import the twin's `/seed` subpath — it is a zod-only leaf.

For stripe that was **false**: `seed.ts` reached `./domain/schema.js` for `ensureStripeTables`, so a CLI import put stripe's domain in the graph `pome --version` loads. That is why the mirror existed at all, and why F-581's fix could not simply be repeated.

- `applySeed` and its four raw row inserts move to `packages/twin-stripe/src/apply-seed.ts`.
- `twin-chunks` now **asserts** the leaf claim for all five twins, walking from each `seed.ts` as its own entry so it holds whether or not anything imports it. Three new lint cases cover the red direction (domain edge, package-root edge, and the split-out write half being green).
- `applySeed` keeps its name and its **package-root** export — the door `@pome-sh/sandbox-domains/stripe` already reaches `StripeDomain` and `openTwinStripeDatabase` through — so that package's published surface does not move. One import line there changes.

`pome --version`: **0.15–0.18s before and after** (3 runs each, stashed vs. applied). The CLI's static graph grows from 153 to 155 files.

## Verification — both directions, through the twins' own surfaces

- **`packages/twin-stripe/test/seed-reads-back.test.ts`** — a flat seed carrying `refunds` and `balance_transactions` read back at `GET /v1/refunds/:id`, `/v1/refunds`, `/v1/balance_transactions`, `/v1/charges/:id`, `/v1/payment_intents/:id`; plus the consequence, a second refund on the fully-refunded charge refused with `charge_already_refunded`. No strip, no throw.
- **`packages/twin-slack/test/seed-reads-back.test.ts`** — one case per top-level seed field, read at `team.info`, `users.list`, `conversations.list`, `conversations.members`, `conversations.history`, `files.list`, `emoji.list`. (`emoji` is the field the task parser refused for two releases while the twin served it.)
- **`packages/checks/test/seed-leaf-preservation.test.ts`** gains maximal stripe and slack fixtures, reusing F-581's `(schema, fixture)` helper — the "Done when" clause about an uncovered new field. Demonstrated by dropping `refunds[].idempotency_key` from the fixture: `expected [ 'refunds.[].idempotency_key' ] to deeply equal []`.

**Hosted half not verified here.** The ticket asks for the read-back "hosted and local"; this PR proves the local half against the twin's own HTTP surface. The hosted half rides the release sequence (`@pome-sh/checks` + `@pome-sh/sandbox-domains` publish → both pome-cloud manifests pinned), which is post-merge work and shared with F-1689.

## A bug in F-581's helper, found by using it

`.transform()` is a `ZodPipe` whose `out` is the transform node, so taking `out` unconditionally dropped `failure_injection[].method` from the *declared* field set entirely — a field the coverage half would have reported as covered while checking nothing. The unwrap now takes whichever side is not the transform.

## Deleted

`cli/test/unit/task-seed-mirror.test.ts` — it compared the mirrors' key sets to the twins', and an object cannot drift from itself. Everything else it pinned (the legacy wrapper, the three greedy-match cases, the empty-seed defaults) moved to `cli/test/unit/task-seed-is-the-twins.test.ts` intact.

## Left hand-written

The gmail and linear arms of `cli/src/contract/seed-state.ts`. Neither F-581 nor F-584 covers them; noted in that file's CHANGELOG rather than fixed here.

## CI/local

Full suite (358 files / 3992 tests), `lint`, `lint:test`, `typecheck`, `lint:dead-code`, `test:contract`, `gate:examples`, `gate:golden`, `gate:route-inputs`, `gate:mcp-tools-list`, both tarball audits, all three manifest gates, `smoke:examples`, `probe:examples`, `probe:twins`, and both showcases' `verify.sh` — green locally.


AFFFPupu:
Closes F-584

PR #512, opened 2026-08-29
A seeded pull request gets the number the seed asked for (F-1153)

AFFFPupu:
> Top of a 4-PR stack: #508 (F-1689) → #509 (F-581) → #511 (F-584) → this. Base is `ao/f-584-stripe-slack-seed-schemas`.

`seedSchema` has accepted `number` on a pull request since the field existed. The applier called `createPullRequest` and kept whatever `nextNumber()` handed out. Measured 2026-08-29:

```
seed:  issue number: 1,  pull_request number: 7
world: issues [1],       pulls [2]
```

**This is not a missing check.** Criteria resolve a pull request *by number*, so a criterion naming #7 either fails at lookup with `pull request #7 not found` — which reads like an agent defect — or resolves to whatever #7 happens to be and grades **the wrong pull request, confidently**. Six `github.pr-comment-exists` criteria in `agent-examples/` ride on the coincidence today, per the [CONTEXT] note on the ticket.

`createIssue` right beside it already did this correctly. This is that path, for pull requests.

## Renumber immediately after create, not at the end

The ticket asks for four child tables to be renumbered (`pull_request_files`, `pull_request_reviews`, `pull_request_review_comments`, `issue_comments`) and warns that a missed one is a silent orphan. Putting the renumber where the **issue** renumber already sits removes the hazard instead of enumerating it: at that moment the only child that exists is `pull_request_files`, written by `createPullRequest` itself. Reviews, statuses, conversation comments and review comments are all seeded *afterwards*, against the returned number, so no table is ever written to a number that then moves.

The test still asks all five children for their rows at the number the seed named — `GET /pulls/7/files`, `/pulls/7/reviews`, `/pulls/7/comments`, `/issues/7/comments` (the PR's own timeline, same table and same number space as an issue's) and `/pulls/7/status` — plus that the issue's own comments did **not** move with it.

### ⚠️ `PRAGMA foreign_keys` is the wrong lever here, and silently so

SQLite makes `PRAGMA foreign_keys` a **no-op inside a transaction**, and `seed()` runs entirely inside one — so toggling it changes nothing and the FK still fails. (I wrote it that way first; the tests said `FOREIGN KEY constraint failed`.)

The PR-child FKs declare `ON DELETE CASCADE` and **not** `ON UPDATE CASCADE` — unlike `issue_labels` / `issue_assignees`, which is exactly why the issue path needs no such dance. With immediate enforcement there is no order that works: updating the parent first strands `pull_request_files` on a key that no longer exists, and updating the child first points it at a parent that does not exist yet. `defer_foreign_keys` *is* settable inside a transaction, holds enforcement to COMMIT, and resets itself when the transaction ends.

`bumpEntityCounter` moves with the renumber, so a pull request **or an issue** created after a seeded #7 gets #8 — one counter, both kinds.

## A number claimed twice is refused at parse

Issues and pull requests share one per-repo counter, so within a repo a number names one entity or the other, and there is no correct silent resolution: honour the first and the second entity is unreachable; honour the last and the first is. Refused in `parseSeed`, naming both claimants by path and naming the number.

The ticket asks for issue-vs-pull-request. The refinement covers pull-request-vs-pull-request and issue-vs-issue too — same shared counter, same absence of a right answer. The same number in two *different* repositories is still fine, and a seed that numbers nothing is untouched.

## Blast radius — measured

Every `<task>.seed.json` in the repo booted through the domain, exported pull-request numbers compared to the seeded ones:

```
22 seeded repo(s) with pull requests; 0 whose numbers moved.
```

The four `pr-summary-*` seeds and the six criteria riding on them are unaffected: they asked for the numbers the ascending-order accident already gave them, and now get them by construction. `cli/tasks/08-prompt-injection-issue-body.seed.json` is the sharpest one — it names PR **#2** beside issue #1, and got #2 by luck.

Green locally: full suite (359 files / 4010 tests), `lint`, `lint:test`, `typecheck`, `lint:dead-code`, `test:contract`, `gate:examples`, `smoke:examples`, `probe:examples`, `probe:twins`, `gate:golden`, `gate:route-inputs`, `gate:mcp-tools-list`, both tarball audits, all three manifest gates, `fidelity:parity` across five twins, `validate:mcp`, twin-github's coverage thresholds, and both showcases' `verify.sh`.

**The ticket's third check has no runner here.** "Re-run the corpus scan, expect still `143 of 143`" — this repo has no such script, and by the ticket's own [CONTEXT] note it could not have caught this either way: *"it resolves phrases and boots no twin, so it would report clean either way."* The scan above is the check that actually binds, because it boots the twin.

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -83,6 +83,12 @@ BESIDES the key sets (the legacy `{ <twin>: { seed } }` wrapper failing loudly,
 the union not greedily mis-matching, the empty-seed defaults) moves to
 `cli/test/unit/task-seed-is-the-twins.test.ts` intact.
 
+A seeded pull request's `number` is honoured (F-1153, in `@pome-sh/twin-github`).
+No bundled task's seed changes number — all of them already got the numbers they
+asked for by the ascending-order accident that is now a guarantee — but a task
+seed claiming one number for both an issue and a pull request is refused at parse
+now, naming both claimants.
+
 ## 0.35.3 — 2026-08-29
 
 **`pome docs getting-started` prints a live page again.** The docs site merged
--- a/packages/checks/CHANGELOG.md
+++ b/packages/checks/CHANGELOG.md
@@ -27,6 +27,12 @@ hand-copying the stripe seed shape.
 
 `slack` gains the `SlackStateSeed` type, matching the other four twins.
 
+`github.parseSeed` refuses a seed that claims one number for two entities in a
+repository (F-1153) — an issue and a pull request, two pull requests, or two
+issues. Issues and pull requests share one per-repo counter, so a number names
+one entity or the other. This is a NEW refusal on a seed that parsed clean
+before, and it is what makes the applier's renumber unambiguous.
+
 ## 0.3.10 — 2026-08-29
 
 **Comment only.** The `stripe.refund-count` / `stripe.refund-exists` source file
--- a/packages/sandbox-domains/CHANGELOG.md
+++ b/packages/sandbox-domains/CHANGELOG.md
@@ -15,6 +15,10 @@ and the write half lives in the twin's `apply-seed.ts`. **This package's exporte
 names, types and shapes are unchanged** — `applySeed` and `applyStripeSeed` both
 still export the same function. Only the specifier one line reads it from moves.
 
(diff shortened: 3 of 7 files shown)


AFFFPupu:
Closes F-1153

PR #528, opened 2026-08-30
`pome checks <twin>` header names the real inlined twin version, or omits it honestly (F-1791)

AFFFPupu:
## Summary

On a plain npm install, `pome checks github` opened with:

```
github — 16 declared checks (@pome-sh/twin-github unknown)
```

— the first line of the first grading-vocabulary output the docs send a newcomer to, reading like something is broken (F-1791).

## Mechanism

`pinnedVersion()` read the CLI's own `package.json` `dependencies` for `@pome-sh/twin-github`. That was right when the twins were real runtime dependencies, but the tsup restructure inlines every `@pome-sh/*` package via `noExternal` and moved them to `devDependencies` as workspace `"*"` links — so the `dependencies` lookup answered "unknown" on every install: dev tree, built bundle, and published tarball alike. Reproduced in all three before the fix.

## Fix

Two resolution layers, then honest omission:

1. **Baked at build time** — `cli/tsup.config.ts` defines `POME_INLINED_PKG_VERSIONS`, a map from each inlined `@pome-sh/*` devDependency to the version its workspace manifest declared when the bundle inlined it (same mechanism as `PKG_VERSION` for `pome --version`). This is what a plain npm install prints.
2. **Workspace fallback** — a source-tree run (`tsx`, vitest) resolves the workspace link's own `package.json`. `packages/sdk` gains `"./package.json": "./package.json"` in its exports map so it resolves like the five twins already do (the stale comment claiming twins can't was from before they exported it).
3. **Unresolvable** — `pinnedVersion` returns `undefined` and the header omits the parenthetical entirely, per the ticket's "honestly omit" branch. No caller can print "unknown" again.

The digest-skew refusal (`vocabulary-skew.ts`) and the offline `checks add` note named the same "unknown" pin; both now go through a shared `pinLabel()` that names the bare package when no version resolves.

## Tests (failing first)

Four new tests in `cli/test/unit/checks-command.test.ts`, all red on the old code:

- header contains the real `packages/twin-github/package.json` version (read straight from the tree, not through the resolution under test) and never "unknown"
- `pinnedVersion` resolves twin + sdk from the workspace link
- unresolvable → `undefined`, header omits the parenthetical, `pinLabel` names the bare package
- the tsup define bakes a real semver for every inlined `@pome-sh/*` devDependency

## Verification

- Clean-room: `npm pack` → install tarball in fresh `/tmp` dir → `pome checks github` prints `github — 16 declared checks (@pome-sh/twin-github 0.12.0)`
- `npm test` (4031 tests), `npm run typecheck`, `npm run lint` (17 rules), `lint:dead-code`, `lint:test`, `gate:route-inputs`, `test:contract` (104 pass), `test:pack` — all green
- `## Unreleased (patch)` changelog entry included

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -9,6 +9,20 @@ write a version number here or in `package.json`. Released entries are insertion
 only: a correction is the next entry, naming the one it corrects.
 
 
+## Unreleased (patch)
+
+**`pome checks <twin>` names the twin's real version instead of "unknown"**
+(F-1791). The header — the first grading-vocabulary output the docs send a
+newcomer to — printed `(@pome-sh/twin-github unknown)` on every install. The
+version lookup still read the CLI's own `dependencies`, but the twins stopped
+living there when tsup began inlining them: they are devDependency workspace
+links now, so that key could never name them again. The build bakes the
+inlined versions into the bundle (the same mechanism as `pome --version`), a
+source tree resolves the workspace link's own manifest, and when neither
+answers the parenthetical is omitted rather than filled with a placeholder.
+The digest-skew refusal and the offline `checks add` note named the same
+"unknown" pin and are fixed by the same resolution.
+
 ## 0.42.3 — 2026-08-30
 
 **`pome compile-seeds` compiles again.** Every file failed with the Anthropic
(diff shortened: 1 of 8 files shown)


AFFFPupu:
twin image (gmail) red = anchore release CDN 504 (syft download, 3 attempts) — the job's own message says it fails closed on a CDN outage, 'not a failure of the twin, the build or the scan'. Not a required check; merging on the three required greens. Same class as the recorded promote-deploy third-party fetch flake.

PR #541, opened 2026-09-15
Twin auth fails closed: no public dev secret, no peerless admin calls, owner-only status file (F-1804, F-1801, F-1800)

AFFFPupu:
## What

Three fail-open defaults in the twin runtime and the CLI's local twin door, closed in one PR (M0 of "OSS test mode — door, proofs, and launch"):

- **F-1804** — `createAdminGate()` refused nothing for a request with no transport peer unless `NODE_ENV=production`. It now refuses it unless the process sets `TWIN_ADMIN_ALLOW_NO_PEER=1`; loopback peers stay admitted, `TWIN_ADMIN_TOKEN` mode is unchanged, and `NODE_ENV` is no longer read.
- **F-1801** — `resolveAuthSecret()` returned the public `dev-only-insecure-secret` whenever `TWIN_AUTH_SECRET` was unset outside `NODE_ENV=production`. It now throws unless the env is set or `POME_ALLOW_DEV_SECRETS=1` is (the same opt-in pome-cloud's `loadTwinSecret()` already requires). A served twin bound to loopback mints a per-process secret and prints it once instead of booting with none; `bearerAuth` answers an unverifiable token with 401 on both bearer shapes rather than a 500.
- **F-1800** — `pome twin start` wrote `.pome/twin-status.json`, which carries the bearer JWT, with the default umask (0644). It is now 0600 in a 0700 directory, with a `chmod` so a file left by an older CLI is tightened on the next start.

`CONTRACT.md` says the new rules; both harness opt-ins are documented there as harness-only.

## Why

The first thing a security-minded stranger does with an OSS repo is read the auth code, and these three sat exactly there. Ao ruled on 2026-09-15 that F-1806 and F-1804 close before any post; F-1806 was already true on `main` (the file is not tracked), and F-1800/F-1801 are one-file fixes in the same files, so they ride along. `NODE_ENV` is not a security boundary — unset, "staging", "preview" and "test" all fell open — so neither decision reads it any more.

## Notes for reviewer

- **What changes for whom.** CLI users: nothing — `twin start`, `run --local` and `doctor` already set their own secret, and the status file just gets tighter. Someone running a packaged twin on loopback with no env (`npm run dev` in a twin package) now sees a generated secret printed at boot instead of silently using the public one; `TWIN_AUTH_SECRET=… npm run dev`, which the twin examples already document, is unchanged. A library consumer booting the sdk app in-process with no env now gets 401s until it sets `TWIN_AUTH_SECRET` or `POME_ALLOW_DEV_SECRETS=1` — hence `@pome-sh/sandbox-domains` gets a **minor** entry and the CLI a patch.
- **The vitest config carries the admin opt-in for every `packages/*` project**, because all 26 in-process suites that hit `/admin/*` drive the app with `app.request()` and have no socket. The three tests that cover the gate itself delete the variable first.
- **Hosted (pome-cloud) is unaffected by the gate change**: a twin in a sandbox is reached over a real socket, so its peer is known. The ticket's second half — inject `TWIN_ADMIN_TOKEN` for every hosted twin, not just Slack (`vercel-sandbox-provider.ts:300`) — is pome-cloud's and is not in this PR.
- Two twin-github scripts CI runs defaulted to the literal dev secret and are fixed here: `scripts/smoke.ts` pins a random `TWIN_AUTH_SECRET` before minting its token, and `scripts/review-harness.ts` does the same for a self-booted twin and refuses to guess when `GITHUB_MCP_URL` points at a twin someone else started. The twins' `examples/` (`claude-github-agent.ts`, `claude-slack-agent.ts`, `rest-curl.sh`) still default to the literal; their documented flow sets `TWIN_AUTH_SECRET` on both sides, and they ship in no tarball and run in no CI step.
- The two per-package coverage configs (`packages/twin-{github,slack}/vitest.coverage.config.ts`) carry the admin opt-in too, because `test:coverage` bypasses the root vitest config — CI's coverage step caught that on the first push.
- Verified by hand on the built CLI: `twin start github --port 3399` → `.pome` is `700`, `twin-status.json` is `600`, MCP `tools/list` and `tools/call create_issue` work with the printed token, the tape records `state_mutation: true`.
- `npx vitest run` for sdk + the five twins: 2193 tests; the two stripe failures the first run showed (`api-keys.test.ts`, unknown/revoked key → 500) were the provider-token path surfacing the missing-secret error as a 500 and are what the `bearerAuth` catch fixes. CLI twin-start unit + e2e: 31 passed. `npm run typecheck -w packages/sdk -w cli` and `npm run lint` green.

## Review required

Yes — auth (the twin runtime's bearer and admin gates) and the twin runtime contract (`CONTRACT.md` boot-secret and administration sections).


AFFFPupu:
Closes F-1804

AFFFPupu:
Closes F-1801

AFFFPupu:
Closes F-1800

